PR #7, opened 2016-04-29
Fix memory leaks in SshCommand.cs

ghost:
Just some small reorganizing of deregistering event handlers and channel cleanup to prevent memory leaks

### DIFF
--- a/src/Renci.SshNet/SshCommand.cs
+++ b/src/Renci.SshNet/SshCommand.cs
@@ -563,7 +563,7 @@ namespace Renci.SshNet
                 this._sessionErrorOccuredWaitHandle.Dispose();
                 this._sessionErrorOccuredWaitHandle = null;
             }
-            
+
             // Dispose managed ResourceMessages.
             if (this._channel != null)
             {
--- a/src/Renci.SshNet/SshCommand.cs
+++ b/src/Renci.SshNet/SshCommand.cs
@@ -338,6 +338,9 @@ namespace Renci.SshNet
                 // TODO: check with Oleg if we shouldn't dispose the channel and uninitialize it ?
                 this._channel.Close();
             }
+            // deregister event handlers from session when the command is cancelled as well.
+            this._session.Disconnected -= Session_Disconnected;
+            this._session.ErrorOccured -= Session_ErrorOccured;
         }
 
         /// <summary>
--- a/src/Renci.SshNet/SshCommand.cs
+++ b/src/Renci.SshNet/SshCommand.cs
@@ -294,6 +294,9 @@ namespace Renci.SshNet
                             _channel.Close();
                         }
 
+                        this._session.Disconnected -= Session_Disconnected;
+                        this._session.ErrorOccured -= Session_ErrorOccured;
+
                         UnsubscribeFromEventsAndDisposeChannel();
                         _channel = null;
 
@@ -503,6 +506,7 @@ namespace Renci.SshNet
             _channel.ExtendedDataReceived -= Channel_ExtendedDataReceived;
             _channel.RequestReceived -= Channel_RequestReceived;
             _channel.Closed -= Channel_Closed;
+            _channel.Close();
 
             // actually dispose the channel
             _channel.Dispose();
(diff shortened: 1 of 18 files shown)


src/Renci.SshNet/SshCommand.cs
```diff
@@ -294,6 +294,9 @@ public string EndExecute(IAsyncResult asyncResult)
                             _channel.Close();
                         }
 
+                        this._session.Disconnected -= Session_Disconnected;
+                        this._session.ErrorOccured -= Session_ErrorOccured;
```
drieseng, line 300: I don't think we should unregister from these events here, as the same SshCommand can be re-executed after its finished.

src/Renci.SshNet/SshCommand.cs
```diff
@@ -503,6 +506,7 @@ private void UnsubscribeFromEventsAndDisposeChannel()
             _channel.ExtendedDataReceived -= Channel_ExtendedDataReceived;
             _channel.RequestReceived -= Channel_RequestReceived;
             _channel.Closed -= Channel_Closed;
+            _channel.Close();
```
drieseng: The dispose of the channel actually closes it as well, so I don't think this is necessary.

src/Renci.SshNet/SshCommand.cs
```diff
@@ -528,47 +532,38 @@ public void Dispose()
...
         {
             // Check to see if Dispose has already been called.
-            if (!this._isDisposed)
-            {
-                // If disposing equals true, dispose all managed
-                // and unmanaged ResourceMessages.
-                if (disposing)
-                {
-                    this._session.Disconnected -= Session_Disconnected;
-                    this._session.ErrorOccured -= Session_ErrorOccured;
```
drieseng, line 538: These should not be removed. 
See my previous comment.

src/Renci.SshNet/SshCommand.cs
ghost, line 300: Good point. I'll find a better location. My use case did not have re-exeuction where the leak was produced, nor was I aware you could re-execute SshCommands.

drieseng, line 300: Can you share a code snippet that allows me to reproduce the leak ?

src/Renci.SshNet/SshCommand.cs
```diff
@@ -528,47 +532,38 @@ public void Dispose()
...
-
-                    // Dispose managed ResourceMessages.
-                    if (this.OutputStream != null)
-                    {
-                        this.OutputStream.Dispose();
-                        this.OutputStream = null;
-                    }
+            if (this._isDisposed)
```
drieseng, line 539: definitely agree on this!

src/Renci.SshNet/SshCommand.cs
ghost, line 300: Something as simple as..

```
class Program
{
    static void Main(string[] args)
    {
        SshClient connection = new SshClient("127.0.0.1", "user", "password");
        connection.Connect();
        if (connection.IsConnected)
        {
            while (true)
            {
                connection.RunCommand("pwd"); //leaks here
            }
        }
    }
}
```

will accumulate EventHandlers that are from "this._session" that keep SshCommands alive. Note that it's hard to see the memory usage from task manager from this program. My use case uses 10+ threads polling a lot and over the course of a few hours the memory usage climbs noticeably. If you use a program called ".Net Memory Profiler" it will tell you this information very neatly. It has a free trial for a few days, using this is how I discovered this particular issue.

ghost:
Made some changes; 

mainly I moved registration of the event handler to the BeginExecute, to pair with EndExecute so that SshCommand could be re-executed. This still fixes the leak, as simply deregistering in Dispose doesn't seem to do the trick.

src/Renci.SshNet/SshCommand.cs
drieseng, line 300: Problem is that you’re not ever disposing the SshCommand.

Try modifying the code to this:

using Renci.SshNet;

namespace ConsoleApplication1

{

```
class Program

{

    static void Main(string[] args)

    {

        SshClient connection = new SshClient("127.0.0.1", "user", "password");

        connection.Connect();

        if (connection.IsConnected)

        {

            while (true)

            {

                var cmd = connection.RunCommand("pwd");

                cmd.Dispose();

            }

        }

    }

}
```

}

We could definitely move event subscribe and unsubscribe from ctor / dispose to BeginExecute / EndExecute.

We’d also need to add this cleanup to CancelAsync().

If however, we ever want to support multiple async executes in parallel using the same SshCommand, then we have to keep the subscribe / unsubscribe as it is right now.

What we could do in that case, is to not only unsubscribe when Dispose(bool) is invoked from the Dispose() method – as part of IDisposable - but also when it’s invoked from the finalizer (with Dispose(false)). That would also fix the leak.

What do you think ?

Regards,

Gert

drieseng:
The leak that you found only produces if you’re not ever disposing the SshCommand.

We could move event subscribe and unsubscribe from ctor / dispose to BeginExecute / EndExecute.
But then we’d also need to add this cleanup to CancelAsync().

If however, we ever want to support multiple async executes in parallel using the same SshCommand, then we have to keep the subscribe / unsubscribe as it is right now.

What we could do in that case, is to not only unsubscribe when Dispose(bool) is invoked from the Dispose() method – as part of IDisposable - but also when it’s invoked from the finalizer (with Dispose(false)). That would also fix the leak. I just commit this change.

Can you confirm that this fixes problem for you ?

drieseng:
Before I forget: thanks for the PR, and for bringing this issue to my attention!

ghost:
Disposing of the SshCommand implicitly with the `Using(SshCommand cmd = ...)` block previously did produce the leak as well. After running the new develop branch with the new fix for a few minutes, it seems I still have the issue. My live usage of SshCommand takes heavy use of the Using blocks, and no SshCommand is executed without the blocks. I just provided the simplest-code-to-bug example I could think of. 

I'll check out the code changes real quick. I can't really work on it too much this weekend.

ghost:
Okay, so I think I know why my changes prevent the leak for my use case. Under the assumption that the Using block will not explicitly call Dispose unless SshCommand has no references left, this explains why deregistering in CancelAsync and EndExecute fixes the problem. Indeed, calling dispose directly will always work.

Edit: Though, according to the docs, a Using block is _supposed_ to call Dispose when it's done.

drieseng:
Can you try to create a more complete (but still small) example that shows the leak ?
Unregistering the event subscriptions in **CancelAsync** should not be necessary if you always dispose the **SshCommand**.

I'm a little puzzled, but that's what we're in IT for :-)

ghost:
Good news. It looks like my error was that I missed a refactor -- a single line! I no longer have the leak with the latest commit here. 

That said, SshCommand _does_ need to get explicitly disposed with either Dispose or within a Using block like so:

> using (SshCommand cmd = connection.RunCommand("pwd"))
>                     {
>                     }

Simply using a command like this : 

> connection.RunCommand("pwd");

will produce the issue with the current code. However, my pull request will allow you to run this and have C# automatically garbage collect it. It does not matter to me if it can or not -- just that with either explicitly disposing or using a Using block fixes the problem.

I think it's safe to call this closed unless the functionality for automagical garbage collection is desired.

drieseng:
If you were in fact already disposing the **SshCommand** instances, then I'm even more puzzled.
In fact, I'm very much considering reverting part of my commit as it only covers a very small fraction of the leaks (if you don't dispose correctly).

I'll let you know when I've committed my last batch of changes, so you can verify the behavior at your end.

drieseng:
I did some refactoring, and reverted the unsubscribe in the finalizer.
Let me know if you now see the leaks again.

ghost:
Looks good as long as dispose is called.

drieseng:
I've now committed a fix that should also work if Dispose is not called (explicitly, or if no using statement is used).

It would be great if you could take time to perform this final test.
Thx!

ghost:
Seems to be the same as before, according to the memory profiler, without directly calling dispose, the two session eventhandlers hold a reference to SshCommand.

drieseng:
The **Session** is preventing a **SshCommand** instance from getting GC'ed, because it references the **SshCommand** through the **Disconnected** and **ErrorOccured** events.

If we do not want to rely on the client invoking **Dispose()**, then I can only see two options to resolve this issue:
1. use [weak references](https://msdn.microsoft.com/en-us/library/aa970850%28v=vs.100%29.aspx)
2. unsubscribe from the **Session** events as soon as possible (not waiting for the client to invoke **Dispose**)

The second approach is rather error prone.
For the **SshCommand** case, we'd need to dispose the channel (as that one also subscribes to **Session** events) and unsubscribe from the **Session** events in the following methods:
- EndExecute(IAsyncResult)
- Session_Disconnected(object, EventArgs)
- Dispose(bool)
- CancelAsync()
- Channel_Closed(object, ChannelEventArgs)

I'll need to take a closer look at the first approach, to see if that doesn't bring along too much overhead and complexity.

drieseng:
For reference, [this](https://gist.github.com/jnm2/e9703624ef5ca7cef74f) seems like a valid option.

WojciechNagorski:
Really old PR, it's hard to even compare it to the current develop branch.

PR #17, opened 2016-06-14
Use Infinite constants where appropriate

matthewvukomanovic:
All the places in the code where an infinite call was being made has been changed to use the two constants:
- Session.Infinite; and
- SshNet.Session.InfiniteTimeSpan.

The change also includes changes to the .gitignore file to ignore the build/target outputs
The build.cmd; build.proj; and build/sandcastle/SSH.NET.shfbproj
files has been fixed.

### DIFF
--- a/src/Renci.SshNet/NetConfClient.cs
+++ b/src/Renci.SshNet/NetConfClient.cs
@@ -21,8 +21,8 @@ namespace Renci.SshNet
         /// Gets or sets the operation timeout.
         /// </summary>
         /// <value>
-        /// The timeout to wait until an operation completes. The default value is <see cref="SshNet.Session.InfiniteTimeSpan"/>
-        /// , which indicates an infinite time-out period.
+        /// The timeout to wait until an operation completes. The default value is negative
+        /// one (-1) milliseconds, which indicates an infinite time-out period.
         /// </value>
         public TimeSpan OperationTimeout { get; set; }
 
--- a/src/Renci.SshNet/NetConfClient.cs
+++ b/src/Renci.SshNet/NetConfClient.cs
@@ -21,8 +21,8 @@ namespace Renci.SshNet
         /// Gets or sets the operation timeout.
         /// </summary>
         /// <value>
-        /// The timeout to wait until an operation completes. The default value is negative
-        /// one (-1) milliseconds, which indicates an infinite time-out period.
+        /// The timeout to wait until an operation completes. The default value is <see cref="SshNet.Session.InfiniteTimeSpan"/>
+        /// , which indicates an infinite time-out period.
         /// </value>
         public TimeSpan OperationTimeout { get; set; }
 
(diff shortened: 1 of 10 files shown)


src/Renci.SshNet/NetConfClient.cs
```diff
@@ -21,8 +21,8 @@ public class NetConfClient : BaseClient
         /// Gets or sets the operation timeout.
         /// </summary>
         /// <value>
-        /// The timeout to wait until an operation completes. The default value is negative
-        /// one (-1) milliseconds, which indicates an infinite time-out period.
+        /// The timeout to wait until an operation completes. The default value is <see cref="SshNet.Session.InfiniteTimeSpan"/>
```
drieseng: Please revert this. I do not plan on exposing Session.InfiniteTimeSpan publicly, and as such we should not mention it in our docs.

src/Renci.SshNet/ScpClient.cs
```diff
@@ -21,8 +21,8 @@ public partial class ScpClient : BaseClient
         /// Gets or sets the operation timeout.
         /// </summary>
         /// <value>
-        /// The timeout to wait until an operation completes. The default value is negative
-        /// one (-1) milliseconds, which indicates an infinite time-out period.
+        /// The timeout to wait until an operation completes. The default value is <see cref="SshNet.Session.InfiniteTimeSpan"/>
```
drieseng: Please revert this. I do not plan on exposing Session.InfiniteTimeSpan publicly, and as such we should not mention it in our docs.

src/Renci.SshNet/SftpClient.cs
```diff
@@ -37,8 +37,8 @@ public class SftpClient : BaseClient
         /// Gets or sets the operation timeout.
         /// </summary>
         /// <value>
-        /// The timeout to wait until an operation completes. The default value is negative
-        /// one (-1) milliseconds, which indicates an infinite timeout period.
+        /// The timeout to wait until an operation completes. The default value is <see cref="SshNet.Session.InfiniteTimeSpan"/>
```
drieseng: Please revert this. I do not plan on exposing Session.InfiniteTimeSpan publicly, and as such we should not mention it in our docs.

src/Renci.SshNet/NetConfClient.cs
matthewvukomanovic: Don't you think it would be better to expose those constants since they are static readonly structs anyway, so can't be changed, so that consumers of the class don't need to define their own infinite constants?

I've got the reverted code ready to go, just haven't added it to the request yet, until I hear what you think.

And yeah I hadn't noticed that they were internal, just assumed they were public.

drieseng: I do not want to expose them.
The BCL already defines a **Timeout** class with these constants.
Do not that this class is (may not be, to verify) not available on all our supported target frameworks.

matthewvukomanovic:
Ok, that's fair enough, the comments have been reverted.

drieseng:
Thanks!

PR #26, opened 2016-06-20
Update PseudoTerminalInfo.cs

trebor-salim:
Correct terminal width (pixels) in SSH_MSG_CHANNEL_REQUEST (pty-req).


drieseng:
Thanks!

Issue #28, opened 2016-06-22
Multiple IP addresses for a host leads to inconsistent behaviour


matthewvukomanovic:
The existing host resolve will always take the first IPAddress which is returned.

We have the situation where the ip address needs to be a IPv4 address since the SFTP server is only bound to the 4 address, however the server also has a IPv6 address.

In some cases there may be times when there will be two addresses of the same type that also need to be decided against.

olegkap:
Added ability to specify client ip. Its same as #130.

IgorMilavec:
Issue #132 seems to provide more information, closing this one

PR #29, opened 2016-06-22
Include a way to filter the ipaddresses resolved for the host with a custom func

matthewvukomanovic:
This relates the the #28 issue.

This includes a way to filter the addresses returned from DnsAbstraction.GetHostAddresses(host) to the actual address that you want.

It also provides some default classes to provide the filter functions.

**NOTE:** Until you have merged in #27 I wouldn't check this branch out.
This is because I used the github instructions (instead of the new git instructions) on how to reset end of lines for existing files, and so the files are still getting changed by the .gitattribute settings.

I just wanted your feedback on this for now, if you want changes done I'll do them, but then wait until #27 has been merged before I push up again, so that it's an easier merge.

### DIFF
--- a/src/Renci.SshNet.NET35/Renci.SshNet.NET35.csproj
+++ b/src/Renci.SshNet.NET35/Renci.SshNet.NET35.csproj
@@ -264,6 +264,12 @@
     <Compile Include="..\Renci.SshNet\Common\TerminalModes.cs">
       <Link>Common\TerminalModes.cs</Link>
     </Compile>
+    <Compile Include="..\Renci.SshNet\Common\AddressTypeFilterClass.cs">
+      <Link>Common\AddressTypeFilterClass.cs</Link>
+    </Compile>
+    <Compile Include="..\Renci.SshNet\Common\AddressTypesFilterClass.cs">
+      <Link>Common\AddressTypesFilterClass.cs</Link>
+    </Compile>
     <Compile Include="..\Renci.SshNet\Compression\CompressionMode.cs">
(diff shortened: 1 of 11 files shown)


drieseng:
I would prefer to add ctors that take (IP)EndPoint that way the client can take charge of this.

matthewvukomanovic:
While that could be a good solution, it doesn't handle the situation where there is a load balanced server which might change.
Especially if the connection is dropped at some point, and the client reconnects automatically under the covers it doesn't give any opportunity to change the address that might be having issues (since the reconnect is most of the time transparent to the user)

wolfsoul13:
Hi Mathew,

I've been working on some adjustments to allow binding of socket to a specific local IPEndPoint (I need this for downloading files using SFTP)

Please let me know if your changes cover this.

Many thanks,

drieseng:
@matthewvukomanovic At this moment I'm inclined against such changes.

wolfsoul13:
@drieseng Would you mind expanding on your reasons for this?
It seems to us to be useful functionality :)

drieseng:
I have no problem at all with introducing overloads that accept an **(IP)EndPoint**. However, I don't think **SSH.NET** should provide an address filtering mechanism. That's not to say we can't have a discussion on this to see if we can come up with a extension point to allow the consuming application to take control of this.

For now, I want to focus on hardening **SSH.NET** and preparing a new stable release.

Rob-Hague:
I'll close this as old and unloved

Issue #33, opened 2016-06-28
Sftp SynchronizeDirectories should not just use the size of the file to sync/update


manu-st:
Currently the implementation of SynchronizeDirectories from SftpClient uses the size of the file to know if it needs to override it. Unfortunately this is not good enough in most cases. There is a comment about using MD5, which would be nice indeed. In the meantime, could we use a some file stamp comparison?

manu-st:
I'm proposing changing this line of code:
`isDifferent = localFile.Length != dest.Length;`
to read
`isDifferent = (localFile.Length != dest.Length) || (dest.LastWriteTimeUtc.AddSeconds(-60.0)  <= localFile.LastWriteTimeUtc);`

The 60s is to avoid any potential clock differences between machines.

raffaeler:
I would keep the difference in no more than 10 seconds. If you are actively working on some files, 1 minute is definitely too much

PR #34, opened 2016-06-30
Use LastWriteTimeUtc when files have the exact same length

manu-st:
This is to better see if files are different or not while waiting
for the MD5 implementation. This is not bullet proof but better than
what exists today.

This is to address issue #33

### DIFF
--- a/src/Renci.SshNet/SftpClient.cs
+++ b/src/Renci.SshNet/SftpClient.cs
@@ -1681,10 +1681,11 @@ namespace Renci.SshNet
 
                 if (!isDifferent)
                 {
-                    var temp = destDict[localFile.Name];
+                    var dest = destDict[localFile.Name];
                     //  TODO:   Use md5 to detect a difference
-                    //ltang: File exists at the destination => Using filesize to detect the difference
-                    isDifferent = localFile.Length != temp.Length;
+                    // ltang: File exists at the destination => Using filesize to detect the difference
+                    // and the LastWriteTime (with a window of 60s to avoid clock differences between machines)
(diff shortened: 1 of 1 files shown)


x2764tech:
@manu-silicon [`FileInfo.LastWriteTimeUtc`](https://msdn.microsoft.com/en-us/library/system.io.filesysteminfo.lastwritetimeutc.aspx) doesn't seem to be supported on Windows Phone or Silverlight

manu-st:
There is LastWriteTime which would work for those platforms. Since this is just trying to copy files that have been changed while keeping the same size, it might still work better than not copying them at all. In any case the proposed PR is not a full solution. I would actually rather always copy the files rather than trying to be smart as this is very hard to fix from the client side.

x2764tech:
Maybe a better way would be allow passing a `Func<SftpFile, FileInfo, bool>` (or create a `delegate` with a similar signature) to let you override comparison? This way, people can implement their own strategies

jenshenneberg:
Some sftp servers, for instance DoveTail Co:Z SFTP (On z/OS mainframe), does not support MD5, gives only an approximate file length and gives a timestamp that is NOT equal to LastWriteTime(It is the last access time instead). The reasons for this are, as far as I know, technical limitations on z/OS.

So in that scenario the only sensible file comparison during synchronization is on filename, unfortunately.

For that reason I would suggest either the delegate of func as x2764tech suggested, or a SyncOption class/enum to specify behavior.

x2764tech:
@jenshenneberg maybe we could build in some default delegates:

    public static class SyncOptions
    {
        public static bool Default(SftpFile remoteFile, FileInfo localFile) 
        {  
            return localFile.Length != temp.Length;
        }
    }
These could then be used like an enum:

    client.SynchronizeDirectories("/", @"C:\temp\", "*.*", SyncOptions.Default);

PR #77, opened 2016-08-31
Add support for host resolution by the proxy itself

pirrmann:
I have a business case where the proxy I need to use has a set of permissions based on the targetted host name, and rejects connections targetting an IP address.
SOCKS5 allows sending the host name instead of the locally-resolved IP addess, so I've implemented support for that.

### DIFF
--- a/src/Renci.SshNet.NET35/Renci.SshNet.NET35.csproj
+++ b/src/Renci.SshNet.NET35/Renci.SshNet.NET35.csproj
@@ -289,7 +289,7 @@
       <Link>HashInfo.cs</Link>
     </Compile>
     <Compile Include="..\Renci.SshNet\HostResolutionMode.cs">
-      <Link>HashInfo.cs</Link>
+      <Link>HostResolutionMode.cs</Link>
     </Compile>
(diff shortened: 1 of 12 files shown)


pirrmann:
Hello,
I'm aware that this is an open-source project and that everyone involved is doing it on his/her spare time. However, could someone have a look at my pull request? I'm ready to improve it if it does not comply to the project's standards, but I'd like to know if merging it will be possible of if I have to find another way of solving my issue.
Thanks, Pierre.

ThomasWFan:
The repository on Github is only a mirror. You can commit to git.kernel.org to contribute.

pirrmann:
@ThomasWFan What are you talking about?

pirrmann:
Thanks for all the kind answers on this pull request, is it worth my time to try to fix the conflicts in this branch?

drieseng:
I'm not sure if we should go for this approach, or use an **EndPoint** argument (which could be **IPEndPoint** or **DnsEndPoint**). Any thoughts?

glenncarr:
Any more thoughts about this?

pirrmann:
Closing this one after 6 years. If someone thinks that this is still relevant, please consider #919 or open another PR.

PR #86, opened 2016-09-14
Suppressing a not-connected exception on socket shutdown

billycc:
I'm trying out SshNet on an ARM device with a very old linux kernel (2.6.29) and old version of mono (3.0.11). I'm not sure if this is a function of my ancient environment.

While closing a socket associated with ChannelDirectTcpip, the _socket seems to respond true to `_socket.Connected`; but throws `SocketException: The socket is not connected` when it hits `_socket.Shutdown()`. Since it's a shutdown procedure anyway, I assume this exception is not important so I just send it to the log and escape out.

I ran some limited tests and by suppressing this exception, SshNet no longer bombs out and instead continues to be available and seems to continue humming along without any other apparent issues. I have not checked if there's other sockets which also have an untrustworthy Connected property.

### DIFF
--- a/src/Renci.SshNet/Channels/ChannelDirectTcpip.cs
+++ b/src/Renci.SshNet/Channels/ChannelDirectTcpip.cs
@@ -132,7 +132,21 @@ namespace Renci.SshNet.Channels
                 if (_socket == null || !_socket.Connected)
                     return;
 
-                _socket.Shutdown(how);
+                try
+                {
+                    _socket.Shutdown(how);
+                }
+                catch (SocketException ex)
+                {
+                    if (ex != null && ex.Message != null &&  ex.Message.Contains("The socket is not connected"))
+                    {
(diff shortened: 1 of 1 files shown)


drieseng:
@billycc I probably won't accept the PR as is, as we don't want to (and cant't) base exception handling on the message of an exception. I case of a **Shutdown**, we could probably just ignore - meaning log - any **SocketException**. That what we do now in `Session.SocketDisconnectAndDispose()`.

I'd prefer to have test coverage for this change too.

billycc:
Yeah, I felt dirty writing it. I did filter it to SocketException, but I'm not sure if there's anything else to look for so I wanted to try to be a little more specific than that. It's hell to get anything over onto my arm box so I didn't spend too much time spelunking. 

I'll see if I can get a more reasonable/repro-able mono environment together and see if I can make a coverage case. Maybe later in the week.

drieseng:
This is now fixed in master.

Issue #100, opened 2016-10-11
SFTP performance


lochnar187:
I am seeing some performance issues and would like to ask others to take my test code and see if they have the same problem.  Mostly, this is because I am running this against an SFTP server that I don't control or have any knowledge about, it's "black box" to me.  So I want to make sure it's not that.  

After noticing I wasn't getting the speeds I expected (sometimes as little as half) I wrote some code that gives a head to head comparison of speeds between SSH.NET and WinSCP.  I used NuGet to get the latest versions of both SSH.NET and WinSCP.  Testing file is an 834KB file.

Using **VS 2015**
Code build is targeted on **.NET Framework 4.5.2**
OS is **Windows 10** (dev machine) and **Windows Server 2012 R2** (test machine)

SSH.NET version, **2016.0.0**
WinSCP version, 5.9.2

SFTP server reports it is: **SSH-2.0-WS_FTP-SSH_7.6.3**

**Test Code:**

```
using Renci.SshNet;
using System;
using System.Diagnostics;
using System.IO;
using WinSCP;

namespace SFTP_Compare {
    class Program {
        static string strServer     = "xxxx";
        static string strUser       = "xxxx";
        static string strPass       = "xxxx";
        static string strHostKey    = "ssh-rsa 1024 xxxxxx";
        static string strRmPath     = "/home/";
        static string strLcPath     = "C:\\Data\\";
        static string strFile       = "test_file.xyz";
        static int    intPort       = 22;
        static void TestWinSCP(int pintLoops) {
            long    lngByteKnt  = 0;
            double  dblCreated, dblConnected, dblOpenStream, dblDownloaded, dblEndTime;
            double  dblKBKntTotal = 0.0, dblCreatedTotal = 0.0, dblConnectedTotal = 0.0, dblOpenStreamTotal = 0.0, dblDownloadedTotal = 0.0, dblEndTimeTotal = 0.0;
            Stopwatch swTimer = new Stopwatch();

            Console.WriteLine();
            Console.WriteLine("WinSCP download test, {0} iteration(s).", pintLoops);
            Console.WriteLine("___________________________________________________________________");
            Console.WriteLine("|  Run  |  Client Cr |  Connection  |    Download    |   Delete   |");

            for (int i = 0; i < pintLoops; i++) {
                try {
                    swTimer.Start();
                    SessionOptions soptSettings = new SessionOptions();
                    soptSettings.Protocol = Protocol.Sftp;
[...]

PR #115, opened 2016-11-13
SftpFileStream.BeginRead/BeginWrite implementation

spkane86:
This is an important first step in addressing SFTP performance issues #100 --being able to asynchronously overlap/pipeline IO in the SFTP stream.

My purposes for this are a little unique in that I will be using these methods directly because my SFTP reads/writes will be directly to/from memory instead of disk, so I didn't go so far yet as to make a SftpClient.DownloadFile overload, for instance, which takes advantage of this. Accordingly, I've tested both of these thoroughly and they seem to work. Going further with this shouldn't be terribly difficult, but thought I should share what I have so far.


WojciechNagorski:
Old PR currently we have `OpenAsync` implemented

Issue #113, opened 2016-11-11
SSH exception after client connect using .NET 3.5 version


haimiko:
I'm able to successfully work with the .NET 4.0 version of SSH.NET but am experiencing issues with a when both the ssh.net and main project are .NET 3.5 version right after the client connect.  I get a "value cannot be null" exception.    This does not happen when the main project is .net 4.0 and SSH.NET is .NET 3.5.

I've tried both password and keyfile authentication and both exhibit the same issue.  See attached code for the way I'm using Renci.SSH. [sshclient.cs](https://github.com/sshnet/SSH.NET/files/586498/sshclient.txt)

System.ArgumentNullException was caught
  Message=Value cannot be null.
Parameter name: All lists are either null or empty.
  Source=Renci.SshNet
  ParamName=All lists are either null or empty.
  StackTrace:
       at Renci.SshNet.Session.WaitOnHandle(WaitHandle waitHandle, TimeSpan timeout) in Session.cs:line 812
       at Renci.SshNet.Session.WaitOnHandle(WaitHandle waitHandle) in Session.cs:line 783
       at Renci.SshNet.Session.Connect() in Session.cs:line 643
       at Renci.SshNet.BaseClient.Connect() BaseClient.cs:line 214
       at SSH.clientSSH(ConnectionInfo connectionInfo, String Computer, String command) in SSH.cs:line 225
  InnerException:  null

More specifically, line 643 in Session.cs (WaitOnHandle(_keyExcahangeCompletedWaitHandle) is what's failing.

sleeveroller:
Just to confirm I'm getting exactly the same issue with the 3.5 library.

petarkekez:
Same problem. Any solution?

sdahan-qualitrol:
Has there been any work done on this? Do we know why is the session hanging on that WaitOnHandle()? Where is that event set in 4.0?

drieseng:
I cannot seem to reproduce this problem, but I'm pretty sure the exception originally comes from ```Session.MessageListener()```.  It appears as if the ```readSockets``` gets cleared, even though the socket is still considered connected. When we subsequently go through the loop again, the ```Socket.Select(...)``` call will fail because all lists are either empty or null.

When an exception occurs in ```Session.MessageListener()```, it is "recorded" and rethrown on the next call to ```Session.WaitHandle(...)```. By rethrowing the exception later, the original stacktrace is lost.
Try subscribing to the ```ErrorOccurred``` event and log the exception.

[...]

PR #155, opened 2017-01-19
Change to Session.cs to address .NET 3.5 compiled version of SSH exce…

Beast09:
…ption after client connect #113

Addressing the outstanding issue .NET 3.5 compiled version of SSH exception after client connect #113 opened by haimiko. 

Replaced Socket.Select(readSockets, null, null, -1) with _socket.Poll(-1, SelectMode.SelectRead);

When executing the .Net 3.5 code in a Visual Studio 2008 C++ CLR session Socket.Select would return an ArgumentNullException. Upon investigation while reading in data the readSockets list would go from a count of 1 to a count of 0 upon execution of Socket.Select. I'm thinking that Socket.Select was editing the contents of the array.

Looking at the comments it appears that the use of Socket.Select over .Poll is preferred so please feel free to disregard these changes.

### DIFF
--- a/src/Renci.SshNet/Session.cs
+++ b/src/Renci.SshNet/Session.cs
@@ -1817,8 +1817,6 @@ namespace Renci.SshNet
         {
             try
             {
-                var readSockets = new List<Socket> {_socket};
-
                 // remain in message loop until socket is shut down or until we're disconnecting
                 while (_socket.IsConnected())
                 {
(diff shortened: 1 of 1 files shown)


drieseng:
Just committed a very similar fix as part of commit 710aa37bd583ced661b02d008514c07c505e6f3d.
Thanks!

Issue #99, opened 2016-10-10
do what 's ssh client can Connect to ssh.net proxy server ( forward port to 127.0.0.1:1080 )


cong88nv:
hi dev. i try connect my app write in c# with HttpWebRequest , proxy = "127.0.0.1:1080"
but it return proxy verssion 67 not support
can u help me

drieseng:
We only support dynamic port forwarding as SOCKS4 or SOCKS5 proxy, while you're expecting us to server as a HTTP proxy.

cong88nv:
@drieseng : how to connect to your " socks5 dynamic port forward - 127.0.0.1:1080" from my c# code
I want get html from website via socks
Thank

reang:
@cong88nv I have same issue with you. Try with set proxy "socks5://127.0.0.1:1080" and its solves in my side

drieseng:
@cong88nv Still need help with this?

cong88nv:
Yes i need

DucTai2308:
Anyone have example for connect ssh throught proxy using socks5 ??

PR #174, opened 2017-02-17
Fixed SOCKS4 and SOCKS5 proxies sending the incorrect port.

RobertCorell:
When establishing a connection with a SOCKS4 or SOCKS5 proxy, the port was being sent incorrectly. The masking logic was incorrect and it ignored the endianness of the host computer. Both issues are resolved,

Tangentially related to issue #99 as the given advice wouldn't work due to this bug.

### DIFF
--- a/src/Renci.SshNet/Session.cs
+++ b/src/Renci.SshNet/Session.cs
@@ -1908,8 +1908,9 @@ namespace Renci.SshNet
             SocketWriteByte(0x01);
 
             //  Send port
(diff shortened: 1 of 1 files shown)


ddalic:
Before i used this fix, SOCKS only worked on port 22 for me, now it's all fine. Thanx!

drieseng:
I'll be fixing this using our **Pack** class instead of using **BitConverter**.
Thanks for bringing this to my attention.
I'm very sorry that it took this long.

Issue #177, opened 2017-02-21
System.NotSupportedException in InternalDownload of ScpClient


DaveKP:
I'm attempting to download a directory from a UNIX system onto Windows. Some of the files contain ":" in timestamps in the filename.

This appears to be causing problems with the FileInfo creation in InternalDownload - Windows doesn't allow colons in filenames. 

Unless I'm missing something there is no option to translate invalid Windows characters from Unix?

Perhaps `ScpClient.Download(string directoryName, DirectoryInfo directoryInfo)`  could have an overload for UnixToWindows conversion where colons are replaced by hyphens?

PR #178, opened 2017-02-21
Added special character handling when copying directories from UNIX to Windows

DaveKP:
#177 

Changes I'm using locally. Not sure about your preferred style, so I haven't added it to constructors or anything for now.

### DIFF
--- a/src/Renci.SshNet/ScpClient.NET.cs
+++ b/src/Renci.SshNet/ScpClient.NET.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using Renci.SshNet.Channels;
 using System.IO;
+using System.Linq;
 using Renci.SshNet.Common;
 using System.Text.RegularExpressions;
 
@@ -13,6 +14,14 @@ namespace Renci.SshNet
     {
         private static readonly Regex DirectoryInfoRe = new Regex(@"D(?<mode>\d{4}) (?<length>\d+) (?<filename>.+)");
         private static readonly Regex TimestampRe = new Regex(@"T(?<mtime>\d+) 0 (?<atime>\d+) 0");
+
+        /// <summary>
+        /// Transferring UNIX files to Windows may require special treatment.
+        /// Enabling this will remove the following special characters from filenames when copying whole directories:
+        /// " &lt; &gt; | : * ? \ /
+        /// </summary>
+        public bool UnixToWindows { get; set; }
+
 
         /// <summary>
         /// Uploads the specified file to the remote host.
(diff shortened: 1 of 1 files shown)


src/Renci.SshNet/ScpClient.NET.cs
```diff
@@ -15,6 +16,14 @@ public partial class ScpClient
         private static readonly Regex TimestampRe = new Regex(@"T(?<mtime>\d+) 0 (?<atime>\d+) 0");
 
         /// <summary>
+        /// Transferring UNIX files to Windows may require special treatment.
+        /// Enabling this will remove the following special characters from filenames when copying whole directories:
+        /// " &lt; &gt; | : * ? \ /
+        /// </summary>
+        public bool UnixToWindows { get; set; }
```
drieseng, line 23: Do we need this extra property, or can we always enable this (or only when running on Windows)?
In that case, we should add a remark for this to ```ScpClient.Download(string, DirectoryInfo)```.

src/Renci.SshNet/ScpClient.NET.cs
```diff
@@ -268,5 +283,11 @@ private void InternalDownload(IChannelSession channel, Stream input, FileSystemI
                 SendConfirmation(channel, 1, string.Format("\"{0}\" is not valid protocol message.", message));
             }
         }
+
+        private static string SanitizeUnixFilenameForWindows(string fileName)
+        {
+            fileName = Path.GetInvalidFileNameChars().Aggregate(fileName, (current, c) => current.Replace(c.ToString(), string.Empty));
```
drieseng, line 289: I'm not sure If I should seek treatment for this, but I'm beginning to get aversion to Linq.
Also please log a message using ```DiagnosticAbstraction.Log(string)``` when at least one character needs to be "sanitized".

drieseng (commented):
I added some remarks.
It would also be great if you could add a unit test for ```ScpClient.Download(string, DirectoryInfo)```, but - too be honest - that method is not easy to test.  You may get inspiration from the ```ScpClientTest_Upload_FileInfoAndPath_Success``` test (or it may scare you away forever).

src/Renci.SshNet/ScpClient.NET.cs
DaveKP: No, not particularly. It was just my way of making sure any users knew that the filenames would be changed because they had to specifically request it, with a log and summary note it shouldn't be necessary.

I'd be torn on whether to always apply the change. Personally I don't see why there'd ever be a good reason to keep special characters that can break things in cross-system setups. But then maintaining original file names where possible might be preferred for some people.

Should be possible to just apply for Windows environments.

Also that test file looks...interesting, I'll take a look in the next few days when I get a chance.

src/Renci.SshNet/ScpClient.NET.cs
DaveKP: All hail the mighty Linq!

I think that was a resharper replacement, I'm just used to seeing Linq everywhere. Imagine that will go when I add the logging.

src/Renci.SshNet/ScpClient.NET.cs
drieseng: If we'd be downloading a single file, I'd advise against "fixing" the name and would instruct the user to specify a destination file without these special characters.

In this case,we're recursively downloading a bunch of files, so either we allow an extensible file name mapping mechanism or ensure it "just works".

Why should we only apply it for Windows? I know it's very unlikely to have a character that is valid on some OS but not on unix,, but for consistency we may just as well always apply it.

Note that there's small risk that - by replacing or removing invalid characters - two distinct files on the remote server may end up having the same file name locally.

PR #182, opened 2017-02-24
Fix a hangup when connecting to Windows Phone SFTP

aurabirb:
This fix would allow WinSSHFS to handle request failures, which are now silently dropped since IChannel does not expose all types of failure events. The problem occurs after issuing a shell command (exec request) that returns a negative error code with a subsequent channel failure response from the server. WinSSHFS waits for either success and closed/exception events which never come, since the channel is still open but no exception has been raised.
Technically, SSH.NET treats channel failures as request failures, so this might be a wrong way to fix things (e.g. break everything)

### DIFF
--- a/src/Renci.SshNet/Channels/Channel.cs
+++ b/src/Renci.SshNet/Channels/Channel.cs
@@ -737,6 +737,11 @@ namespace Renci.SshNet.Channels
                 try
                 {
                     OnFailure();
+                    // Close the channel since it failed anyway;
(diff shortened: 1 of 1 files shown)


aurabirb:
Ehm, I didn't touch anything that might have triggered the linter warning... All tests do pass, however. I am supposed to fix this anyway, I guess.

drieseng:
I'll fix the CI build.

WojciechNagorski:
Old PR, Currently SSH.NET do not support Windows Phone.

PR #186, opened 2017-03-03
Fix for data corruption

agayardo:
Fix for data corruption when reading to a buffer that is smaller than the internal buffer. This was introduced with the last commit 5 days ago.


drieseng:
Before I forget: thanks!!

agayardo:
no problem, thank you for this great library!

Issue #194, opened 2017-03-09
Consider revising behavior of SftpFileStream


drieseng:
Our **SftpFileStream** currently performs separate householding for read and write mode.
This means - for example - that if you:
1. open an existing file for read/write
1. read two bytes
1. write a single byte

The byte that you wrote in step **3** is actually written at position **0** instead of **2**.

The following code fragment exposes this:
```cs
using (var client = new SftpClient(...))
{
    client.Connect();

    using (var s = client.Open(path, FileMode.CreateNew, FileAccess.Write))
    {
        s.Write(new byte[] { 5, 4, 3, 2, 1 }, 1, 3);
    }

    using (var s = client.Open(path, FileMode.Open, FileAccess.ReadWrite))
    {
        Console.WriteLine("#1: " + s.ReadByte());
        Console.WriteLine("#2: " + s.ReadByte());
        Console.WriteLine("#3: " + s.Position);

        s.WriteByte(7);
        s.Write(new byte[] {8, 9, 10, 11, 12}, 1, 3);

        Console.WriteLine("#4: " + s.Position);
    }
}
```

drieseng:
As I mentioned offline, I think you're moving a little too fast here by rewriting **SftpFileStream** from scratch. Rewriting for performance without affecting behavior is of course ok, but you're definitely changing the way **SftpFileStream** works for a client.

From a single(-threaded) consumer point-of-view, the current behavior does not make any sense.

From a multi-threaded point-of-view, it makes sense that advancing the position in thread A by reading should not affect the position of thread B that is writing.  The original implementation was clearly ment to be thread-safe.

Before we rewrite **SftpFileStream** we should first discuss how we expect it to work (eg. do we want it to be thread-safe), and perhaps even involve our user base.

olegkap:
Well, the idea behind SftFileStream was to have same behaviour as FileStream, and for that I used  Microsoft implementation of it [here](https://referencesource.microsoft.com/#mscorlib/system/io/filestream.cs,804b80fe667d89ca) as a starting point (in the previous implementation I was using Mono as reference I think), and replace native functionality to access files with SSH equivalent functions.

I did leave some parts out of it, the one that has to do with async, at least for now we can restore it later.

So I guess the question is how SftpFileStream behavior now is different from FileStream and if should it be.

[...]

PR #199, opened 2017-03-19
SftpFileStream: No longer track write position separately

drieseng:
Reading from the SftpFileStream will now also affect the position at which a subsequent write will be performed. The position at which a write is performed in the server file is now always the current position minus the position in the buffer.

Also see #194.


drieseng:
If and when we decide to accept this - or a similar - PR, we should also discuss the thread safety of this class.

Issue #238, opened 2017-06-21
Support forward to unix socket


tmds:
OpenSSH supports forwarding to unix sockets: https://lwn.net/Articles/609321/.

For example, a mysql server on `remotehost` using socket `/var/lib/mysql/mysql.sock` can be made accessible on the local machine on TCP port `5000`:
```
ssh -L 127.0.0.1:5000:/var/lib/mysql/mysql.sock remotehost
```

LightCC:
Are you just talking about a normal forwarded port using SSH?  This functionality is already built-in.  See an example here: https://stackoverflow.com/questions/41152737/c-sharp-ssh-net-tunnel-mysql-client-ioexception.

Check out ForwardedPort (base class), ForwardedPortLocal, ForwardedPortRemote, and ForwardedPortDynamic classes.

There are a few quirks though...

drieseng:
@LightCC No, he's talking about [unix domain sockets](https://en.wikipedia.org/wiki/Unix_domain_socket).

Don't hesitate to file issues for those "quirks".

PR #239, opened 2017-06-21
Support forward to unix socket

tmds:
This is a minimal implementation to support forwarding to a unix socket (https://github.com/sshnet/SSH.NET/issues/238).
```
new ForwardedPortLocal("localhost", 5000, "/var/lib/mysql/mysql.sock");
```
makes the unix socket `/var/lib/mysql/mysql.sock` on the remote host available at `localhost:5000`.

Please provide feedback on the implementation and what tests should be implemented.

### DIFF
--- a/src/Renci.SshNet.NET35/Renci.SshNet.NET35.csproj
+++ b/src/Renci.SshNet.NET35/Renci.SshNet.NET35.csproj
@@ -410,6 +410,9 @@
     <Compile Include="..\Renci.SshNet\Messages\Connection\ChannelOpen\DirectTcpipChannelInfo.cs">
       <Link>Messages\Connection\ChannelOpen\DirectTcpipChannelInfo.cs</Link>
     </Compile>
+    <Compile Include="..\Renci.SshNet\Messages\Connection\ChannelOpen\DirectStreamLocalChannelInfo.cs">
+      <Link>Messages\Connection\ChannelOpen\DirectStreamLocalChannelInfo.cs</Link>
+    </Compile>
     <Compile Include="..\Renci.SshNet\Messages\Connection\ChannelOpen\ForwardedTcpipChannelInfo.cs">
       <Link>Messages\Connection\ChannelOpen\ForwardedTcpipChannelInfo.cs</Link>
     </Compile>
--- a/src/Renci.SshNet.NET35/Renci.SshNet.NET35.csproj
+++ b/src/Renci.SshNet.NET35/Renci.SshNet.NET35.csproj
@@ -86,6 +86,18 @@
     <Compile Include="..\Renci.SshNet\Channels\ChannelDirectTcpip.cs">
       <Link>Channels\ChannelDirectTcpip.cs</Link>
     </Compile>
+    <Compile Include="..\Renci.SshNet\Channels\ChannelDirectBase.cs">
+      <Link>Channels\ChannelDirectBase.cs</Link>
+    </Compile>
+    <Compile Include="..\Renci.SshNet\Channels\ChannelDirectStreamLocal.cs">
+      <Link>Channels\ChannelDirectStreamLocal.cs</Link>
+    </Compile>
+    <Compile Include="..\Renci.SshNet\Channels\IChannelDirectBase.cs">
+      <Link>Channels\IChannelDirectBase.cs</Link>
+    </Compile>
+    <Compile Include="..\Renci.SshNet\Channels\IChannelDirectStreamLocal.cs">
+      <Link>Channels\IChannelDirectStreamLocal.cs</Link>
+    </Compile>
     <Compile Include="..\Renci.SshNet\Channels\ChannelForwardedTcpip.cs">
       <Link>Channels\ChannelForwardedTcpip.cs</Link>
     </Compile>
--- a/src/Renci.SshNet.Silverlight/Renci.SshNet.Silverlight.csproj
+++ b/src/Renci.SshNet.Silverlight/Renci.SshNet.Silverlight.csproj
@@ -413,6 +413,9 @@
     <Compile Include="..\Renci.SshNet\Messages\Connection\ChannelOpen\DirectTcpipChannelInfo.cs">
       <Link>Messages\Connection\ChannelOpen\DirectTcpipChannelInfo.cs</Link>
     </Compile>
+    <Compile Include="..\Renci.SshNet\Messages\Connection\ChannelOpen\DirectStreamLocalChannelInfo.cs">
+      <Link>Messages\Connection\ChannelOpen\DirectStreamLocalChannelInfo.cs</Link>
+    </Compile>
     <Compile Include="..\Renci.SshNet\Messages\Connection\ChannelOpen\ForwardedTcpipChannelInfo.cs">
       <Link>Messages\Connection\ChannelOpen\ForwardedTcpipChannelInfo.cs</Link>
     </Compile>
--- a/src/Renci.SshNet.Silverlight/Renci.SshNet.Silverlight.csproj
+++ b/src/Renci.SshNet.Silverlight/Renci.SshNet.Silverlight.csproj
@@ -101,6 +101,18 @@
     <Compile Include="..\Renci.SshNet\Channels\ChannelDirectTcpip.cs">
       <Link>Channels\ChannelDirectTcpip.cs</Link>
     </Compile>
+    <Compile Include="..\Renci.SshNet\Channels\ChannelDirectBase.cs">
+      <Link>Channels\ChannelDirectBase.cs</Link>
+    </Compile>
+    <Compile Include="..\Renci.SshNet\Channels\ChannelDirectStreamLocal.cs">
+      <Link>Channels\ChannelDirectStreamLocal.cs</Link>
+    </Compile>
+    <Compile Include="..\Renci.SshNet\Channels\IChannelDirectBase.cs">
+      <Link>Channels\IChannelDirectBase.cs</Link>
+    </Compile>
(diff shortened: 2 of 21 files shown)


tmds:
> feedback on the implementation 

The implementation is using `IChannelDirectTcpip` and `ChannelDirectTcpip`. In `ChannelDirectTcpip` a `DirectStreamLocalChannelInfo` is used instead of `DirectTcpipChannelInfo`. Is this ok? Or should these classes be renamed/duplicated? Note: the port argument of IChannelDirectTcpip.Open is ignored and the socket path is passed as the remoteHost argument.

drieseng:
Thanks for the great contribution!

I only had very little time to look into it, but I would indeed prefer to introduce a **ChannelDirectStreamLocal** (or **ChannelDirectUnixDomainSocket** or something).
We can extract a base class from **ChannelDirectTcpip** and have both **ChannelDirectTcpip** and the new channel class derive from it.

I have the same remark/objection for the changes to **ForwardedPortLocal**. Adding just another ctor does not make this functionality discoverable.  Having a **Port** property that does not apply and a **Host** property that does not reflect its meaning do not give me a warm feeling (even though its 34°C here).

I'll try to give it some more thought.

_Hope I do not sound too negative, definitely not intentional!_

tmds:
Thanks for the feedback. I've added a `ChannelDirectStreamLocal` class and share code with `ChannelDirectTcpip` via the `ChannelDirectBase` class.

As a proposal for the `ForwardedPortLocal` I've changed the constructor to take two arguments which specify the local and remote address.
These are valid options:
```C#
new ForwardedPortLocal("localhost", "/var/lib/mysql/mysql.sock"); // unix socket forward
new ForwardedPortLocal("localhost:5000", "www.google.com:80"); // regular forward
```
The addresses are also exposed as `LocalAddress` and `RemoteAddress` property.
When the remote address is a unix socket address, the `Host` and `Port` now return string.Empty and 0.

tmds:
@drieseng I hope you'll find some time to look at this and provide feedback. This enables some interesting scenarios and it would be nice if it can be part of your next release.

drieseng:
I leave for a short holiday today.
I hope to have more time when I return.
Note that I first want to finalize the current release cycle (which has been ongoing for way too long).

tmds:
Thanks for letting me know. Enjoy your holiday.

tmds:
@drieseng hope you're having a good holiday. I'm looking forward to continue working on this. I can build and use this feature myself. I'm working on a new release of a dbus library I maintain. Since dbus is accessible via a Unix socket by default, this will be a great enabler to access it remotely in a secure way. Once the feature lands in a release, I can document how users can use ssh.net for remote dbus.

drieseng:
I plan to take a few days off in the next week or two, so that should give me more time to hack on SSH.NET.
My first priority is to get a new beta out.
This PR comes right after that.

drieseng:
@tmds
Can I ask you for your feedback and opinion on other issues as well?

tmds:
@drieseng let's try and see how it goes

tmds:
The PR has a
```
+            if (localAddress.IsUnixSocketAddress())
+            {
+                throw new NotSupportedException("localAddress cannot be a unix socket path");
+            }
```
This is when the user wants to use a unix socket locally instead of a tcp socket. This has some perf&security advantages but only works on Unix.
I can take a look at implementing that once we finish this PR.

tmds:
CI failed with
```
Failed   WaitOnHandleOnSessionShouldWaitForChannelCloseMessageToBeReceived
Error Message:
   Assert.IsTrue failed. 
Stack Trace:
   at 
```

drieseng:
@tmds That one has bitten me a few times already. I tried to harden it a few days ago, but apparently that didn't help much. I'll see if I can find time this weekend to have a closer look.

drieseng:
I hope my nightly labour fixed the test.

tmds:
@drieseng hope we can make some progress on this PR soon :)

drieseng:
I promise to make time for this PR soon.

tmds:
hey @drieseng I'd love to get this in a released version of ssh.net :)

drieseng:
Can you rebase, or should I add a branch in **SSHNET**?

tmds:
@drieseng I have rebased. CI failed with
```
Failed   ReceiveOnServerSocketShouldReturnZero
Error Message:
   Test method Renci.SshNet.Tests.Classes.SessionTest_Connected_ServerSendsBadPacket.ReceiveOnServerSocketShouldReturnZero threw exception: 
System.Net.Sockets.SocketException: An established connection was aborted by the software in your host machine
Stack Trace:
    at System.Net.Sockets.Socket.Receive(Byte[] buffer, Int32 offset, Int32 size, SocketFlags socketFlags)
   at Renci.SshNet.Tests.Classes.SessionTest_Connected_ServerSendsBadPacket.ReceiveOnServerSocketShouldReturnZero() in C:\projects\ssh-net\src\Renci.SshNet.Tests\Classes\SessionTest_Connected_ServerSendsBadPacket.cs:line 88
```

drieseng:
@tmds First of all, sorry that this is taking so long. I'm really swamped at work (and commuting 3 to 4 hours a day doesn't really help), which leaves me little to no time for anything else. Nonetheless, I'll try to push myself a little harder to make progress on this PR.

I started looking into the proposed changes, and I'm really in doubt between introducing a new **ForwardedPort** class (eg. **ForwardedPortLocalUnixDomainSocket**, but this is quite a mouthful) and updating the existing **ForwardPortLocal** to only take a (local and remote) **EndPoint**. 

Adding yet another ctor and special casing unix domain addresses all over (I exaggerate, I know) is something I'd like to avoid.

If we add choose to update **ForwardPortLocal** to only accept a local and remote endpoint, then we need to either roll our own **UnixDomainSocketEndPoint** class (and over time add support for **[System.Net.Sockets.UnixDomainSocketEndPoint](https://github.com/dotnet/corefx/blob/master/src/System.Net.Sockets/src/System/Net/Sockets/UnixDomainSocketEndPoint.Unix.cs)**.  

Removing these ctors is a breaking change, but one I've already consider before:
* ForwardedPortLocal(UInt32 boundPort, String host, UInt32 port)
* ForwardedPortLocal(String boundHost, String host, UInt32 port)
* ForwardedPortLocal(String boundHost, UInt32 boundPort, String host, UInt32 port)

tmds:
@drieseng don't worry about it taking some time. I often find myself in that spot too and it some occasions too it has taken me some months to be able to address a problem/feature.

A thought on using EndPoints: the EndPoints classes are meant as a user friendly way to represent various types of `SocketAddresses` (via the `Create`, `Serialize `methods). So for the local representation it is appropriate, since you can pass that `EndPoint` directly to a `Socket` method. For the remote side, the result of the _Serialize_ operation would be more something like `ChannelOpenInfo` (instead of SocketAddress) since those are the bits you want to pass over the wire. The `UnixDomainSocketEndPoint `currently doesn't expose a Path property, which is what you'd use to build the `ChannelOpenInfo`.

drieseng:
@tmds If I recall correctly, the path is exposed via the `ToString()` override. That said, we cannot use the **UnixDomainSocketEndPoint** from .NET Core as that would limit applications that can use this feature to those targeting **.NET Core 2.2** (or something), and even further restricted to those running on a **Unix** flavor.

Do you happen to know if there's a valid reason why the **UnixDomainSocketEndPoint** ctor throws a **PNSE** on **Windows**, instead of having those methods on *Socket* throw that take an **EndPoint** in case that endpoint is a **UnixDomainSocketEndPoint** (and the OS is not a **Unix** flavor).

tmds:
These classes are meant to represent socket address types to be used with the socket class. On Linux UnixDomainSocketEndpoint takes into account the max path length. It will throw if you try to create a path which is longer than what the platform supports, even if such a path would be valid on other systems.

drieseng:
Using it in **SSH.NET** is clearly not an option then. Having our own proper **UnixDomainSocketEndpoint** class would probably also only confuse users.

tmds:
hi @drieseng I hope this feature makes it into ssh.net at some point. The first two commits in this PR are the interesting ones. The others deal with API/abstraction. The initial commit added a `ForwardedPortLocal` constructor that accepts a string address for the remote unix path. I find this simplest but I don't have a strong opinion on the API. Please take a look when you find some time.

shynome:
the project is died ? any other project can replace ?

tmds:
Closing due to age.

drieseng:
Sorry dude. I commute almost 4 hours a day which has a huge impact on my OSS activity.

darkoperator:
Have you considered opening up the project to have more maintainers? Dont know if there would be more people bit it could be a option

tmds:
@drieseng don't worry, I know first hand how challenging it is to do OSS in spare time. I'm cleaning up my old PRs and issues, hence the close. Thank you for all the effort you put in this project!

darkoperator:
I second this, thanks for all the hard work :) My project would not be the success it is without this library. Thank You!

GamesVineyard:
I know this thread has been closed for a very long time, but in one of my current projects I needed exactly this functionality. So I used @tmds's dead branch, and it works nicely! 
I also know that there have been 269 commits since then and that it's not a good idea to use a 7 years old, deprecated branch. So my question is: is there any chance to revive this issue and merge the solution after all? 😳

GamesVineyard:
I know this thread has been closed for a very long time, but in one of my current projects I needed exactly this functionality. So I used @tmds's dead branch, and it works nicely! 
I also know that there have been 269 commits since then and that it's not a good idea to use a 7 years old, deprecated branch. So my question is: is there any chance to revive this issue and merge the solution after all? 😳

PR #245, opened 2017-06-27
add test setup

ilijamitkov:
In this PR I've added dockerized test servers

Added docker testing containers
- sshnettestserver
- socks5_proxy
- http_proxy
docker-compose.yml is located in /docker folder

Please note that Test project Resources has been updated to reflect the docker containers credential hence the failed appveyor test

This PR also contains bug fix for proxy destination address port in Session.cs


drieseng:
@ilijamitkov Sorry that I never got around to reviewing your PR. I'm now looking into using docker for our integration tests.

In these integration tests, I want to reconfigure OpenSSH a lot.
Stopping, rebuilding and starting the container for each test is not an option.
I was thinking about having a container with a dummy entrypoint (tail -f /dev/null), and then just stop/reconfigure/start sshd in each test. The container would not have to be restarted at all during the test run.

Do you have a better idea?

Issue #256, opened 2017-07-15
ScpClient does not properly quote paths


danvac:
scpClient.Download does not work on MikroTik 5.26
I am getting exception Renci.SshNet.Common.ScpException: scp error: /"file.backup": no such file or directory!

When tested scp from Fedora box, it worked without problems.
Please provide info what informations you will need for fixing (debugging).

drieseng:
Do you happen to have a device that I could test against?

danvac:
Device is prepared. Is there any private way to tell you credentials and connection info?
@drieseng

drieseng:
You should be able to contact me through my [profile](https://github.com/drieseng).

drieseng:
What (remote) path did you specify when you got the error you mentioned above?
Can you share the code fragment in question?

danvac:
```
downloads.Add("batchtelnet.rsc");
...
foreach (string download in downloads) {
...
                                    try
                                    {
                                        scp.Download(download, File.Create(inputfilepath));
                                        System.Threading.Thread.Sleep(scpDelay);
                                    }
                                    catch (Renci.SshNet.Common.ScpException ex)
                                    {
                                        _log.log(String.Format("{0}: {1}", ex.GetType().ToString(), ex.Message));
                                    }
                                    catch (Exception ex)
                                    {
                                        _log.log(String.Format("{0}: {1}", ex.GetType().ToString(), ex.Message));
                                    }

```

drieseng:
Was able to reproduce the issue.
There's a similar issue with `Upload(...)`.

drieseng:
I've fixed this issue (and a few related ones) locally.
I still want to add some tests and verify whether these fixes also resolve #108.

drieseng:
The problem here was that we were using a over-simplified quoting mechanism (meaning surround any path with double quotes). Apparently MikroTik does not handle (double) quoted paths.

I've now updated our implementation to apply the following scheme:

* embed a single-quotes in quotes (' => "'")
* escape exclamation points with a backslash, and do not embed this in single-quotes (! => \!)
* embed all other characters in single-quotes (a => 'a')

This should work with all shells.

drieseng:
Enclosing characters in single quotes should preserve the literal  value of each character within the single-quotes. However, I've just learned that MikroTik instead adds the single quotes to the actual file or directory name :(

This is not the first quirk I've seen in the MikroTik SCP implementation.
[...]

Issue #108, opened 2016-11-02
Cisco compatible SCP


bazauto:
I've been working with the Posh-SSH PowerShell module, which uses this library, to transfer files to/from Cisco devices via SCP. With the Cisco devices as the SCP server. Currently testing with a 2801 router running IOS 12.4(24)T1.

We also have newer routers running 15.4 but I have not had chance to test one of those yet.

I've found that the Cisco scp process interprets the quotes added around the filename as part of the filename. If I remove the quotes the file transfer works as expected.

I changed...
```
channel.SendExecRequest(string.Format("scp -t {0}", pathOnly));
```
to ...
```
channel.SendExecRequest(string.Format("scp -t \"{0}\"", pathOnly));
```

It is also worth noting, I'm using the stream based overrides as the Cisco scp process doesn't support the "-p, -r or -d" options.

Is this something you would consider adding compatibility for? Would you like me to propose a solution in a pull request?

cocowalla:
I've had the same problem with Cisco devices - would be nice to get a fix merged in.

drieseng:
@cocowalla @bazauto 
Do you guys have a device that I can test a fix against?

darkoperator:
Probably try with GNS3 if a device is not available, I will check my storage unit I may have an 800 there

bazauto:
I can setup a publicly accessible Cisco device, but it will be next week, if that is any use.

drieseng:
@bazauto Yes, that would definitely be useful. Thanks!

drieseng:
In PR #290 I've introduced a mechanism to disable path quoting, and at the same time we now allow for a custom quoting implementation to be plugged in.

That PR (and the develop branch) also features a new default quoting implementation.
More info on this implementation is available [here](https://github.com/sshnet/SSH.NET/blob/7dde512a9ca662d245b90ad87f28d5a981de637f/src/Renci.SshNet/RemotePathTransformation.cs#L12).

Can you build **SSH.NET** from the **[scp-quoting](https://github.com/sshnet/SSH.NET/tree/scp-quoting)** branch and:
* Try using the new default quoting implementation to see if that works for **CISCO** routers.
* If not, try using the **None** transformation to disable quoting altogether.

Please perform tests with both "regular" paths, and paths with spaces and special characters (#\<\>;%?'" ...).

bazauto:
[...]

PR #290, opened 2017-08-27
Introduce extensible remote path quoting mechanism

drieseng:
Add a **RemotePathTransformation** property to **ScpClient** that can be used to control if and how a remote path should be transformed before passed on to the scp command.

This allows for a custom transformation (escaping/quoting) - that implements **IRemotePathTransformation** - to be plugged into **ScpClient**.

Out of the box, we only provide two implementations that are exposed through the **RemotePathTransformation** (enum-like) class: **Quote** and **None**.

Usage:

```cs
using (var client = new ScpClient("HOST", "USER", "PWD"))
{
    // Quotes a path in a way to be suitable to be used with a shell.
    // This is the default.
    client.RemotePathTransformation = RemotePathTransformation.Quote;
    // Prforms no quoting or escaping.
    client.RemotePathTransformation = RemotePathTransformation.None;

    client.Connect();

    client.Upload(...);
}
```

Fixes issue #256.

**Note:**
I have not updated our unit tests, as I consider this to be a call for feedback.

### DIFF
--- a/src/Renci.SshNet/RemotePathQuoteTransformation.cs
+++ b/src/Renci.SshNet/RemotePathQuoteTransformation.cs
@@ -1,0 +1,161 @@
+﻿using System;
+using System.Text;
+
+namespace Renci.SshNet
+{
+    /// <summary>
+    /// Quotes a path in a way to be suitable to be used with a shell.
+    /// </summary>
+    internal class RemotePathQuoteTransformation : IRemotePathTransformation
+    {
+        /// <summary>
+        /// Quotes a path in a way to be suitable to be used with a shell.
+        /// </summary>
+        /// <param name="path">The path to transform.</param>
+        /// <returns>
+        /// A quoted path.
+        /// </returns>
+        /// <exception cref="ArgumentNullException"><paramref name="path"/> is <c>null</c>.</exception>
+        /// <remarks>
+        /// <para>
+        /// If <paramref name="path"/> contains a single-quote, that character is embedded
+        /// in quotation marks (eg. "'"). Sequences of single-quotes are grouped in a single
+        /// pair of quotation marks.
+        /// </para>
+        /// <para>
+        /// If a shell command contains an exclamation mark (!), the C-Shell interprets it as a
+        /// meta-character for history substitution. This even works inside single-quotes or
+        /// quotation marks, unless escaped with a backslash (\).
+        /// </para>
+        /// <para>
+        /// References:
+        /// <list type="bullet">
+        ///   <item>
+        ///     <description><a href="http://pubs.opengroup.org/onlinepubs/7908799/xcu/chap2.html">Shell Command Language</a></description>
+        ///   </item>
+        ///   <item>
+        ///     <description><a href="https://earthsci.stanford.edu/computing/unix/shell/specialchars.php">Unix C-Shell special characters and their uses</a></description>
+        ///   </item>
+        ///   <item>
+        ///     <description><a href="https://docstore.mik.ua/orelly/unix3/upt/ch27_13.htm">Differences Between Bourne and C Shell Quoting</a></description>
+        ///   </item>
+        /// </list>
+        /// </para>
+        /// </remarks>
+        /// <returns>
+        /// The transformed path.
+        /// </returns>
+        /// <exception cref="ArgumentNullException"><paramref name="path"/> is <c>null</c>.</exception>
+        public string Transform(string path)
+        {
+            if (path == null)
+            {
+                throw new ArgumentNullException("path");
+            }
+
+            // result is at least value and (likely) leading/trailing single-quotes
+            var sb = new StringBuilder(path.Length + 2);
+            var state = ShellQuoteState.Unquoted;
+
+            foreach (var c in path)
+            {
+                switch (c)
+                {
+                    case '\'':
+                        // embed a single-quote in quotes
+                        switch (state)
+                        {
+                            case ShellQuoteState.Unquoted:
+                                // Start quoted string
+                                sb.Append('"');
+                                break;
+                            case ShellQuoteState.Quoted:
+                                // Continue quoted string
+                                break;
+                            case ShellQuoteState.SingleQuoted:
+                                // Close single-quoted string
+                                sb.Append('\'');
+                                // Start quoted string
+                                sb.Append('"');
+                                break;
+                        }
+                        state = ShellQuoteState.Quoted;
+                        break;
+                    case '!':
+                        // In C-Shell, an exclamatation point can only be protected from shell interpretation
+                        // when escaped by a backslash
+                        // Source:
+                        // https://earthsci.stanford.edu/computing/unix/shell/specialchars.php
+
+                        switch (state)
+                        {
+                            case ShellQuoteState.Unquoted:
+                                sb.Append('\\');
+                                break;
+                            case ShellQuoteState.Quoted:
+                                // Close quoted string
+                                sb.Append('"');
+                                sb.Append('\\');
+                                break;
+                            case ShellQuoteState.SingleQuoted:
+                                // Close single quoted string
+                                sb.Append('\'');
+                                sb.Append('\\');
+                                break;
+                        }
+                        state = ShellQuoteState.Unquoted;
+                        break;
+                    default:
+                        switch (state)
+                        {
+                            case ShellQuoteState.Unquoted:
+                                // Start single-quoted string
+                                sb.Append('\'');
+                                break;
+                            case ShellQuoteState.Quoted:
+                                // Close quoted string
+                                sb.Append('"');
+                                // Start single-quoted string
+                                sb.Append('\'');
+                                break;
+                            case ShellQuoteState.SingleQuoted:
+                                // Continue single-quoted string
+                                break;
+                        }
+                        state = ShellQuoteState.SingleQuoted;
+                        break;
+                }
+
+                sb.Append(c);
+            }
+
+            switch (state)
+            {
+                case ShellQuoteState.Unquoted:
+                    break;
+                case ShellQuoteState.Quoted:
+                    // Close quoted string
+                    sb.Append('"');
+                    break;
+                case ShellQuoteState.SingleQuoted:
+                    // Close single-quoted string
+                    sb.Append('\'');
+                    break;
+            }
+
+            if (sb.Length == 0)
+            {
+                sb.Append("''");
+            }
+
+            return sb.ToString();
+        }
+
+        private enum ShellQuoteState
+        {
+            Unquoted = 1,
+            SingleQuoted = 2,
+            Quoted = 3
+        }
+    }
+}
--- a/src/Renci.SshNet/RemotePathTransformation.cs
+++ b/src/Renci.SshNet/RemotePathTransformation.cs
@@ -1,8 +1,25 @@
 ﻿namespace Renci.SshNet
 {
     /// <summary>
-    /// Allow access to built-in remote path transformations.
+    /// Provides access to built-in remote path transformations.
     /// </summary>
+    /// <remarks>
+    /// References:
+    /// <list type="bullet">
+    ///   <item>
+    ///     <description><a href="http://pubs.opengroup.org/onlinepubs/7908799/xcu/chap2.html">Shell Command Language</a></description>
+    ///   </item>
+    ///   <item>
+    ///     <description><a href="https://earthsci.stanford.edu/computing/unix/shell/specialchars.php">Unix C-Shell special characters and their uses</a></description>
+    ///   </item>
+    ///   <item>
+    ///     <description><a href="https://docstore.mik.ua/orelly/unix3/upt/ch27_13.htm">Differences Between Bourne and C Shell Quoting</a></description>
+    ///   </item>
+    ///   <item>
+    ///     <description><a href="https://blogs.msdn.microsoft.com/twistylittlepassagesallalike/2011/04/23/everyone-quotes-command-line-arguments-the-wrong-way/">Everyone quotes command line arguments the wrong way</a></description>
+    ///   </item>
+    /// </list>
+    /// </remarks>
     public static class RemotePathTransformation
     {
         private static readonly IRemotePathTransformation ShellQuoteTransformation = new RemotePathShellQuoteTransformation();
@@ -51,10 +68,6 @@
         ///   <item>
         ///     <term>/var/would be 'kewl'!, not?</term>
         ///     <term>'/var/would be '"'"'kewl'"'"\!', not?'</term>
-        ///   </item>
-        ///   <item>
-        ///     <term>!ignore!</term>
-        ///     <term>\!'ignore'\!</term>
         ///   </item>
         ///   <item>
         ///     <term></term>
@@ -109,12 +122,8 @@
         ///     <term>&quot;/var/garbage!/temp&quot;</term>
         ///   </item>
         ///   <item>
-        ///     <term>/var/would be 'kewl'!/not?</term>
+        ///     <term>/var/would be 'kewl'!, not?</term>
         ///     <term>&quot;/var/would be 'kewl'!, not?&quot;</term>
-        ///   </item>
-        ///   <item>
-        ///     <term>!ignore!</term>
-        ///     <term>&quot;!ignore!&quot;</term>
         ///   </item>
         ///   <item>
         ///     <term></term>
@@ -130,6 +139,5 @@
         {
             get { return DoubleQuoteTransformation; }
         }
-
     }
 }
--- a/src/Renci.SshNet/RemotePathTransformation.cs
+++ b/src/Renci.SshNet/RemotePathTransformation.cs
@@ -1,0 +1,83 @@
+﻿namespace Renci.SshNet
+{
+    /// <summary>
+    /// Allow access to built-in remote path transformations.
+    /// </summary>
+    public static class RemotePathTransformation
+    {
+        private static readonly IRemotePathTransformation QuoteTransformation = new RemotePathQuoteTransformation();
+        private static readonly IRemotePathTransformation NoneTransformation = new RemotePathNoneTransformation();
+
+        /// <summary>
+        /// Quotes a path in a way to be suitable to be used with a shell.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// If the path contains a single-quote, that character is embedded in quotation marks.
+        /// Sequences of single-quotes are grouped in a single pair of quotation marks.
+        /// </para>
+        /// <para>
+        /// An exclamation mark (!) is escaped with a backslash, because the C shell would otherwise
+        /// interprete it as a meta-character for history substitution. It does this even if it's
+        /// enclosed in single-quotes or quotation marks, unless escaped with a backslash (\).
+        /// </para>
+        /// <para>
+        /// All other character are enclosed in single-quotes, and grouped in a single pair of
+        /// single quotes where contiguous.
+        /// </para>
+        /// </remarks>
+        /// <example>
+        /// <list type="table">
+        ///   <listheader>
+        ///     <term>Original</term>
+        ///     <term>Quoted</term>
+        ///   </listheader>
+        ///   <item>
+        ///     <term>/var/log/auth.log</term>
+        ///     <term>'/var/log/auth.log'</term>
+        ///   </item>
+        ///   <item>
+        ///     <term>/var/mp3/Guns N' Roses</term>
+        ///     <term>'/var/mp3/Guns N'"'"' Roses'</term>
+        ///   </item>
+        ///   <item>
+        ///     <term>/var/garbage!/temp</term>
+        ///     <term>'/var/garbage\!/temp'</term>
+        ///   </item>
+        ///   <item>
+        ///     <term>/var/garbage!/temp</term>
+        ///     <term>'/var/garbage'\!'/temp'</term>
+        ///   </item>
+        ///   <item>
+        ///     <term>/var/would be 'kewl'!/not?</term>
+        ///     <term>'/var/would be '"'"'kewl'"'"\!'/not?'</term>
+        ///   </item>
+        ///   <item>
+        ///     <term>!ignore!</term>
+        ///     <term>\!'ignore'\!</term>
+        ///   </item>
+        ///   <item>
+        ///     <term></term>
+        ///     <term>''</term>
+        ///   </item>
+        /// </list>
+        /// </example>
+        public static IRemotePathTransformation Quote
+        {
+            get { return QuoteTransformation; }
+        }
+
+        /// <summary>
+        /// Performs no transformation.
+        /// </summary>
+        /// <remarks>
+        /// This transformation should be used for servers that do not support escape sequences in paths
+        /// or paths enclosed in quotes, or would preserve the escape characters or quotes in the path that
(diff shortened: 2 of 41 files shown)


drieseng:
cc @tmds

src/Renci.SshNet/RemotePathQuoteTransformation.cs
```diff
@@ -0,0 +1,161 @@
...
+        /// </list>
+        /// </para>
+        /// </remarks>
+        /// <returns>
+        /// The transformed path.
+        /// </returns>
+        /// <exception cref="ArgumentNullException"><paramref name="path"/> is <c>null</c>.</exception>
+        public string Transform(string path)
+        {
+            if (path == null)
```
tmds: Consider checking if the path doesn't consist of characters that don't need escaping (`[a-zA-Z0-9,._+:@%/-]`).
MicroTik and Cisco and others that don't do unescaping, will probably have paths that don't require escaping. So they will work even when using the Transform.

Next, you could scan the string for the single quote character (`'`). If it isn't present in the path, you can pre and postpend it to escape the path.

If the two cases above don't apply, scan through the string and prepend every char with a backslash except those that don't require escaping. There is no need to enclose the entire string.

examples:
```
file1.txt     -> file1.txt
my file.txt   -> 'my file.txt'
my 'file'.txt -> my\ \'file\'.txt
```
Also for `!` this is fine, e.g.:
```
/tmp$ echo text
text
/tmp$ echo !-1
echo echo text
echo text
/tmp$ echo \!-1
!-1
```

src/Renci.SshNet/RemotePathTransformation.cs
```diff
@@ -0,0 +1,83 @@
...
+        ///     <term>!ignore!</term>
+        ///     <term>\!'ignore'\!</term>
+        ///   </item>
+        ///   <item>
+        ///     <term></term>
+        ///     <term>''</term>
+        ///   </item>
+        /// </list>
+        /// </example>
+        public static IRemotePathTransformation Quote
```
tmds: I prefer the name `Escape`/`ShellEscape`. Quote is about enclusing with `'` or `"`. Escaping is doing more.

src/Renci.SshNet/RemotePathQuoteTransformation.cs
drieseng: I'll give it some more thought.

If we only enclose characters in single quotes when we encounter "special characters", then - on SCP servers that do not unescape - we risk having **SSH.NET** work just fine until a file that does require escaping is uploaded or downloaded.

I would prefer our implementation to either always work out-of-the-box, or to fail immediately (and require disabling of quoting through the **RemotePathTransformation** property).

src/Renci.SshNet/RemotePathTransformation.cs
drieseng: I initially had another transformation mechanism that I called **Escape**.
That one would only escape special characters with a backslash.
I removed that option as I wasn't sure I had an exhaustive list of "special characters".

In general, the mechanism is referred to as (Shell) quoting.
Also, our current implementation only escapes exclamation marks using a backslash.
For all other characters (except for a single quote), their literal value is preserved by enclosing them in single quotes. This is explaining in the xml doc.

src/Renci.SshNet/RemotePathQuoteTransformation.cs
tmds: > If we only enclose characters in single quotes when we encounter "special characters", then - on SCP servers that do not unescape - we risk having SSH.NET work just fine until a file that does require escaping is uploaded or downloaded.

It's a bit hard to know how to deal with a file that requires escaping on an scp server that doesn't unescape. So servers that don't unescape (on embedded devices) are not likely to have files that need escaping.

With the current implementation when downloading a file you observe the failure because it can't find the file. If you were to upload a file, you may end up creating a file called `'firmware'` instead of `firmware`.

drieseng: I agree that - with the current implementation - you'll only get failures when downloading a file, and you'll indeed risk uploading files enclosed in single quotes (and/or with double quotes, or a backslash). 

However, the advantage of the current implementation is that this behavior will immediately manifest itself and not only upon the first file that contains characters that deem to require quoting. Nothing bothers me more than something that works "most of the time".

PS. I only tested a few, but AFAIK none of the Windows SCP servers unescape even though you'll definitely have files or directories that contain "special characters" (mostly spaces).

drieseng:
# Problem statement

**SSH.NET** currently applies an over-simplified quoting mechanism for remote paths in **ScpClient**. More precisely, any remote path is simply enclosed in double quotes before it is passed to the **scp** command on the remote host.

This mechanism does not work on all types of SCP server.

At a high level, we can distinguish the following groups of SCP servers:
* Shell-based servers
* Non-shell based server

Shell-based servers require that certain metacharacters - that would for example otherwise result in history expansion - be quoted or escaped to preserve their literal value. The actual characters that require escaping or quoting depend on the shell.

Non-shell based servers typically do not require escaping or quoting of characters, and - equally important - do not unescape or unquote the path that is passed to the **scp** command.

To perform SCP operations for a remote host, the path on which to perform action must be supplied to the **scp** command. This path should take into account the quoting requirements (or lack thereof) of the remote host.

The SCP protocol does not provide a mechanism to "encode" a path using the quoting mechanism of the remote host, nor does it allow you to discover the preferred quoting mechanism of the remote host or the list of characters that require encoding/quoting for the remote host.

Even though not all shell-based servers share the same quoting requirements, a single quoting mechanism (such as the one implemented [here](https://github.com/sshnet/SSH.NET/blob/7dde512a9ca662d245b90ad87f28d5a981de637f/src/Renci.SshNet/RemotePathQuoteTransformation.cs)) can be applied that should meet the demand of all shell-based SCP servers.

Applying this same quoting mechanism to paths that are passed to non-shell based servers would actually result in IO requests that contain the literal value of the quote/escape characters.

**Example:**
> To pass an exclamation mark as literal value to an SCP server that is based on C shell, the exclamation mark must be escaped with a backslash ('!' -> '\\!').  To pass that same character to a non-shell based server escaping MUST NOT be applied. If - for a non-shell based server - the exclamation mark were to be escaped with a backslash, then the literal value of this backslash would be passed to the IO layer and you would end up having '\\!' in your file or directory path.

The path produced by the current double-quoting transformation does not preserve the literal value of all metacharacters for shell-based SCP servers, but - more importantly - for non-shell based SCP servers either the enclosing (double) quotes end up in the actual file or directory path, or the path is rejected altogether due to the enclosing quotes.

Up until now there's no way to disable this transformation or choose another quoting mechanism.

# Solution

Personally, I see the following options here:

1. Do nothing, and stick with the current double-quoting mechanism.
1. Introduce a new quoting mechanism, and enable it by default.
1. Do not performing any quoting by default.
1. Force users to choose the quoting mechanism by adding a (required) **remotePathTransformation** argument to all **ScpClient** ctors.
1. Introduce new ctors that add an additional **remotePathTransformation** argument, and mark the existing ctors obsolete.

In case of options 1 to 3, we need to allow the user/developer to change the quoting mechanism (through something like the **[RemotePathTransformation](https://github.com/sshnet/SSH.NET/blob/7dde512a9ca662d245b90ad87f28d5a981de637f/src/Renci.SshNet/ScpClient.cs#L66)** property).

## Option 1

Always enclose the path in double quotes.

#### Advantages
- Works out-of-the-box for most paths on shell based SCP servers.

#### Disadvantages
* Does not preserve the literal value of all metacharacters.
* Fails on non-shell based SCP servers.

## Option 2

Change the default transformation for remote path from double quoting to either a non-agressive or agressive quoting mechanism.

### Non-agressive quoting

Only applt quoting when there are characters that need to be quoted.

#### Advantages
* Works out-of-the-box for all paths on shell based SCP servers.
* Works out-of-the-box for paths that do not contain metacharacters on non-shell based SCP servers.

#### Disadvantages
* We need an exhaustif list of metacharacters (across all shell-based SCP servers).
* Fails on non-shell based servers for paths containing characters that are deemed to require quoting.
* Such failures may only surface in very specific conditions (surely only in production, and definitely not when you expect it).

### Agressive quoting

Always apply (adaptive) quoting. A proposal of this mechanism is available [here](https://github.com/sshnet/SSH.NET/blob/7dde512a9ca662d245b90ad87f28d5a981de637f/src/Renci.SshNet/RemotePathQuoteTransformation.cs).

#### Advantages
* Works out-of-the-box for shell based servers.
* Fails consistenly or consistently produces files containing literal value of quoting characters on non-shell based servers for both "simple" paths, and paths containing metacharacters.

#### Disadvantages
* Produces files containing the literal value of quoting characters on non-shell based servers when quoting mechanism is not changed to **None**.

### Option 3

Do not apply any quoting at all.

#### Advantages
* Works out-of-the-box for non-shell based servers.

#### Disadvantages
* Fails on shell-based servers for paths containing metacharacters.
* Huge breaking change (runtime).

### Option 4

Have users/developers specify the (remote) path transformation to apply when creating a **ScpClient** instance.

#### Advantages
* Forces users/developers to select the transformation that applies to their SCP server.

#### Disadvantages
* Breaking change (compile-time).
* Not all users/developers know if or which quoting mechanism should be used.
  A bad choice may only surface as such when a file path containing metacharacters is encountered.

### Option 5

Introduce new ctors that add an additional **remotePathTransformation** argument, and mark the existing ctors obsolete.

#### Advantages
* Encourages users/developers to select the transformation that applies to their SCP server.

#### Disadvantages
* Either leaves existing ctors in pseudo-broken state (if we continue applying option 1 to these ctors), or still requires us to choose option 1, 2 or 3 for these ctors.
* The new ctors share a disadvantage with option 4: do we trust users/developers to (have enough knowledge of the remote host to) make a well-informed choice?

darkoperator:
I prefer of the option the option 2

drieseng:
@darkoperator 
The non-agressive or agressive option?
(aggressive may have a negative connotation but I don’t consider it negative)

darkoperator:
Yep the none aggressive option. For my use case I can make it in to a flag on the cmdlet so the user can just enable the more aggressive method if needed.

drieseng:
As mentioned in the comment above, I see the following issues with that approach:
* For non-shell based servers, you'll risk having a great experience until one day you need to upload or download a file containing one of these metacharacters.
  As I mentioned above, my experience tells me that you won't run into this until you've gone live.
  The "aggressive" approach would work for all paths on shell-based servers, and fail fast for all paths on non-shell based servers. This makes the aggressive approach more predictable to me.
* We'd need an exhaustive list of metacharacters (characters that require escaping) across shellls; I prefer that over a list of safe characters or character classes.
  In the "aggressive" approach we enclose all characters (except for exclamation mark and single quote itself) in single quotes turning off their (potential) special meaning, hence there's no need to have an exhaustive list of safe or special characters.

I'm not trying to convince you here.
Neither approach is perfect.

drieseng:
@hifi 
Can you share your view on this?
Summary:
* Works fine on shell-based servers and fasts fail on non-shell based servers; works fine on non-shell base servers after changing value of **RemotePathTransformation** property to **None**.
-or-
* Works fine on shell-based servers for all paths, and appears to work fine on non-shell based servers until you encounter a file or directory path containing metacharacters (space, #, !, <, ...).

hifi:
Without reading too much into details I would look into `libssh2` and OpenSSH how they deal with paths or do they leave it to the user.

I don't see the proposed solution problematic as long as the default makes sense and would be in-line with other widely used implementations.

tmds:
Transforms will always cause certain failures. By transforming as little as possible, chances of failure are minimized.
Aggressive transform don't ensure the error is caught. If a transformed path returns file-not-found on download, the client can assume the file is not there. If a transformed path succeeds on upload, it may contain the escaped characters literally in its name.

The `scp` command doesn't transform (aka option 3). This is making the user responsible for escaping. That's not very user friendly.

+1 for option 2 non-agressive

> For my use case I can make it in to a flag on the cmdlet so the user can just enable the more aggressive method if needed.

It would be better to have a flag to switch to None. Then the cmdlet would behave like the `scp` command.

hifi:
> The scp command doesn't transform (aka option 3). This is making the user responsible for escaping. That's not very user friendly.

I would then do what the `scp` command does and explicitly require enabling some quoting mechanism. Likely `libssh2` does not try to do any quoting either.

It might not be the most user friendly way but at least it would be consistent with other implementations. The documentation should make the quoting options clearly visible and examples enable them when dealing with "normal" shell based servers.

The SFTP client is a different beast and does not have this issue, right? Many who deal with OpenSSH servers can opt to use the SFTP client of SSH.NET instead of SCP and embedded users who use SSH.NET probably know what they need to do with the target system.

Regardless what the default is, documenting the behaviour is more important and having the possibility to override it.

drieseng:
@hifi 
I wonder how many users know what quoting mechanism a given SCP server expects?
Forcing users to choose, may just as well backfire and cause them to pick the wrong mechanism (and still create issues for us to investigate, of course).
I prefer a good default.

@tmds 
None of the options are perfect; you can't fix a broken protocol :-)
From my experience, I still prefer to "fail" fast.
And you're right, to users it might not always be clear what is causing the "failures" or garbled paths, but at least they'll be confronted with it immediately.

darkoperator:
I wonder how does the OpenSSH scp binary does it?

drieseng:
@darkoperator It does not perform any quoting for you. On Windows (the MS port) it's even worse, you can't even use the backslah as escape character as it wants to be intelligent and translates it to a forward slash.

@darkoperator, @hifi, @tmds Do you guys recon that - by default - we should no longer perform any quoting? I'm just afraid that many of the **SSH.NET** users don't know that quoting may be necessary, and that the lack of quoting is causing their upload or download to fail.

tmds:
If an scp server doesn't support escaping and it has use-cases that require escaped paths, then I consider that server broken.
So to me, option 2 non agressive is the best default. It works on servers that I don't consider broken and it's user-friendly.

drieseng:
@tmds Are you referring to my response to @darkoperator ? That concerned the scp binary that's included in the (MS) Win64 port of OpenSSH. That one is indeed broken.

Option 2 non-aggressive works on shell servers for all paths, and works on non-shell servers for paths that do not contain metacharacters.  But as I mentioned, it will start failing the moment you encounter paths (file or directory) that contain such metacharacters.  The problem there is that it does not produce a consistent result. I prefer a consistent bad result as that is something you can deal with straight away (eg. by choosing the **None** transformation).  I prefer not having a default that all of a sudden may start producting "bad" paths.

I get the impression that neither of us succeeds in convincing the other. I'll have to bribe you with beer, "peins ik" (=Flemish for "I guess").

tmds:
I was referring to the scp server. My take is that, non-shell servers will be used where quoting isn't needed (e.g. uploading firmware to embedded devices or downloading their logs).
It's good to have a discussion to figure out the options. As maintainer, you get to choose and I'm fine with that.

Misschien komen we elkaar wel eens tegen en dan drinken we wel een biertje op SSH.NET en FOSS :wink:

drieseng:
@tmds Maintainer or not, I want to make a well-informed decision. I don't want to impose a bad solution on users.

I did some more tests, and learned that OpenSSH on Windows (MS port) requires paths containing metacharacters to be quoted (and quotes escaped with a backslash). The fun part is that they also appear to be replacing backslashes with forward slashes. I created an [issue](https://github.com/PowerShell/Win32-OpenSSH/issues/868) to learn more about their quoting requirements.

Enclosing a path in single quotes makes the single quotes literally end up in the file name.

drieseng:
I just did a test with a Cisco router (for issue #108).

Downloading or uploading a file only works if I do not apply any quoting mechanism all.
This matches what I saw with the Mikrotek router.
So far so good. We have the **None** path transformation in the proposal for this.

The bad news is that this only works if the file name does not contain a space (and/or certain metacharacters). The device I tested does not appear to support any quoting mechanism, but at the same time it fails when the path contains certain characters.

By now, I've spent so much time on this - and made this little progress - that I'm considering this:
* Keep the current default, which always encloses the path in double quotes, with one minor change which is to escape double quotes that are embedded in the path with a backslash.
* Introduce a **RemotePathTransformation** property on **ScpClient** that allows a user to specify the mechanism that is appropriate for its server(s).
* Provide the following remote path transformation out-of-the-box:
  * SingleQuote (or ShellQuote)
    This is the aggressive quoting mechanism.
  * DoubleQuote
    This is the mechanism described above.
  * Escape
     Escapes any metacharacter with a backslash.
  * None
     Does not perform any transformation on the path.

tmds:
@drieseng your thoughts on these alternative names:
```
SingleQuote -> UnixShell
DoubleQuote -> WindowsShell
None -> None
Escape -> who needs this?
```
This makes it more clear what value a user should pick. Also it matches with the cause of the problem: the shell on the remote server.

drieseng:
@tmds @hifi @darkoperator 
I've decided to keep the current default (see my last three comments above as to why I've given up), and prepared a page with more information.

Please review [this](https://github.com/sshnet/SSH.NET/wiki/ScpClient:-Remote-path-transformation) page.

Thanks!

drieseng:
@tmds 
> your thoughts on these alternative names:
> ```
> SingleQuote -> UnixShell

I had first thought of naming it **ShellQuote** (but I know, you'll prefer ShellEscape :p).

> DoubleQuote -> WindowsShell

Wouldn't this be a bit strange as it is - and has always been - the default? Associating that default with Windows hosts may raise a few eyebrows as very few of them have probably targeted any Windows hosts.

It should also work fine with Unix (shell) hosts (as long as there are no paths with any of the metacharacters).

However, I agree that we should encourage the use of **SingleQuote** for shell-based hosts (which typically run on a Unix flavor).

The only Windows-based SCP server that appears to require double quoting is the MS-port of OpenSSH, but I only tested a few. Can you even consider that MS port based on a Windows shell?

> None -> None
> Escape -> who needs this?

I agree we can do without.

> This makes it more clear what value a user should pick. Also it matches with the cause of the problem: the shell on the remote server.

After all I ran into the last few days, I'm not sure if it's possible at all to make something clear. 
I'd love to have clear naming and clear guidelines as to when to use a specific quoting mechanism, but I consider this a lost cause :-)

drieseng:
@tmds 
Is [this](https://repository.genmymodel.com/drieseng/SCP2?utm_source=dashboard) the decision matrix we expect users to apply?

tmds:
> The only Windows-based SCP server that appears to require double quoting is the MS-port of OpenSSH, but I only tested a few. Can you even consider that MS port based on a Windows shell?

For Windows, this is the OpenSSH code path:
Check if it is 'scp':

https://github.com/PowerShell/Win32-OpenSSH/blob/c1680168e4035be32fb06fb1e1853a2b0607378d/session.c#L448

Then copy the command as is:

https://github.com/PowerShell/Win32-OpenSSH/blob/c1680168e4035be32fb06fb1e1853a2b0607378d/session.c#L488

And give it to `CreateProcess`:

https://github.com/PowerShell/Win32-OpenSSH/blob/c1680168e4035be32fb06fb1e1853a2b0607378d/session.c#L553

So regular Windows command line parsing applies: https://msdn.microsoft.com/en-us/library/17w5ykft.aspx.

(**note** that the same function does have an alternate code path which does care about special characters: https://github.com/PowerShell/Win32-OpenSSH/blob/c1680168e4035be32fb06fb1e1853a2b0607378d/session.c#L491-L495. The `scp` call is not following that path.)

> Is this the decision matrix we expect users to apply?

I'm not sure if users can answer those questions.
Try-and-error order: ShellQuote, DoubleQuote, None(, custom).
(or: Unix~~Shell~~, Windows~~Shell~~, None :wink:).

drieseng:
@tmds 
I agree that there probably aren't many SSH.NET users that can answer these questions, which is exactly what I wanted to make clear by defining the decision process that we expect them to follow.

Have you had time to take a look at the wiki page?

Regarding the name of the transformations: don't you agree that it's a bit strange to have a transformation named "Windows" as default?  More because it's not really Windows specific.
And yes, I'm still convinced we should keep that double-quoting mechanism as default.
While I don't like the current names (and I'm convinced naming isn't my best quality), I don't think your proposal adds more clarity. I can live with the Unix name, but that leaves the "other" :-)

**Note:**
We could also detect the transformation that is best suited for a given server, but I'm not sure I want to go in that direction. I haven't checked if you can even distinguish between OpenSSH on *nix and Windows.

tmds:
> Have you had time to take a look at the wiki page?

Yes, I had a look.

> More because it's not really Windows specific.

I'd consider this the spec: https://msdn.microsoft.com/en-us/library/17w5ykft.aspx. This entire thing is 'Microsoft Specific'.

> I don't think your proposal adds more clarity.

I think it does since you can match with the OS on the server.
Naming things is hard...

> The default.

The most user-friendly default imo is to try for each path to do `None, fallback to DoubleQuote, fallback to ShellQuote`.
Selecting `DoubleQuote` as the default is the next best option. This is failing on servers that require `None` and making paths that require `ShellQuote` fail. So it will work in most cases.

drieseng:
> Yes, I had a look.

Any remarks (apart from the discussion concerning the double quoting transformation)?

> I think it does since you can match with the OS on the server.

Double quoting an argument is hardly Windows specific.
It works for pretty much all metacharacters on *nix too.

> The most user-friendly default imo is to try for each path to do None, fallback to DoubleQuote, fallback to ShellQuote.

Problem is that it's not straightforward (or 100% foolproof) to detect that a scp operation was rejected because the path was not accepted. I don't want to base logic on the error message returned by the scp server.

And I don't like too much magic anyway.

tmds:
> Any remarks (apart from the discussion concerning the double quoting transformation)?

Nothing specific.

> Double quoting an argument is hardly Windows specific. It works for pretty much all metacharacters on *nix too.

I know. For Windows, it's the spec. If you don't want to call it Windows because it works in most of cases on Unix, that's fine for me.

> Problem is that it's not straightforward (or 100% foolproof) to detect that a scp operation was rejected because the path was not accepted. I don't want to base logic on the error message returned by the scp server.

I wasn't proposing to try the scp operation. But another 'default' Transform that would be the 'least-aggressive'. DoubleQuote is fine as default too.

drieseng:
@hifi I'm in the final stage of completing the (long overdue) 2016.1.0 release. I plan to include a way to explicitly set the "quoting" mechanism in the beta3 release, but keep the current (double quote) mechanism as default.

For that release, the priority is to provide flexibility to users and describe the options available.
To that end, I'd love to have constructive feedback on the [wiki](https://github.com/sshnet/SSH.NET/wiki/ScpClient:-Remote-path-transformation) that I created.

I don't think many users can (or have enough information to) make a well-informed decision on the quoting mechanism to apply. This is what's keeping me from forcing the users to select one of the options available.

drieseng:
@tmds 
I've seen double and single quoting being referred to as weak and strong quoting.
Would that be better than DoubleQuote and ShellQuote?
If we use weak quoting instead of double quoting, then we should escape a few extra characters (eg. $, ` and backslash itself).

tmds:
> I've seen double and single quoting being referred to as weak and strong quoting.
Would that be better than DoubleQuote and ShellQuote?
If we use weak quoting instead of double quoting, then we should escape a few extra characters (eg. $, ` and backslash itself).

I wouldn't change DoubleQuote so it becomes more unix like (we have ShellQuote for unix). So keeping the current is better imo.

drieseng:
Yes, I came to that conclusion too since my comment.
Naming it "weak" would mean what we would have to change it to match the weak "spec", making it less usable on Windows.

PR #300, opened 2017-09-15
Fix #153 support of Task-based Asynchronous Pattern (TAP) over existi…

### LINKED ISSUE #153: Add support for a Task based async API

bartizan:
Fix #153 support of Task-based Asynchronous Pattern (TAP) over existing async API for SftpClient.  Added into a new project targeting .net 4.5.

What is done. Added a new SshNet project for support await/async and prepared an extensions over SftpClient. Nuget package is still waiting for an update it the change approved.

### DIFF
--- a/src/Renci.SshNet.NET45/Properties/AssemblyInfo.cs
+++ b/src/Renci.SshNet.NET45/Properties/AssemblyInfo.cs
@@ -1,0 +1,6 @@
+﻿using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+[assembly: AssemblyTitle("SSH.NET .NET 4.5")]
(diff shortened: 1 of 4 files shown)


mmoo9154:
@Bartizan it looks like a number of unneccesary changes to the sln files got checked in with your pull request.  You might have better luck if you just commit the changes to src/Renci.SshNet.NET45/SftpClientExtensions.cs which is the significant change.

bartizan:
@mmoo9154 I see a lot of changes since my PR. I have prepared a new one (#661) according to latest codebase. I had to avoid net35 targeting, maybe it was beyond general convention and it would be better to introduce a definition like FEATURE_SFTPCLIENT_TAP_EXT-something instead of NET35.

PR #323, opened 2017-10-17
Added new Method - EnsureDirectory

kstolte:
## EnsureDirectory("")
Added new method that extends current Existing Folder check to allow for a path to be passed and for SSH.Net to automagically create the missing file structure for a path. It is smart enough to build out the directory hierarchy to maintain the same structure and prevent errors in cases where the same directory hierarchy is required. Use case is when Syncronizing directories with Sub Folders.

### Use Case
You have a local file structure that you are trying to duplicate on another server. It is imperative that these files maintain the same relative paths.

#### Local File Structure
./Folder1
|- SubFolder1
|- SubFolder2
        |- SubSubFolder1

#### Pre-`EnsureDirectory` of Server File Structure
Folder1
|- SubFolder1

#### After executing: `EnsureDirectory("./Folder1/SubFolder2/SubSubFolder1")`
./Folder1
|- SubFolder1
|- SubFolder2
        |- SubSubFolder1

### DIFF
--- a/src/Renci.SshNet/SftpClient.cs
+++ b/src/Renci.SshNet/SftpClient.cs
@@ -605,6 +605,37 @@ namespace Renci.SshNet
             }
         }
 
+		/// <summary>
+		/// Checks to see if an entire directory path exists on the server, if it does not exist it will create it for you.
+		/// <para>This is best used when you are trying to syncronize directories or ensure that the file structure remains the same in the .</para>
+		/// </summary>
+		/// <param name="path">SFTP path that you want to check</param>
+		/// <exception cref="ArgumentException"><paramref name="path"/> is <b>null</b> or contains only whitespace characters.</exception>
+		/// <exception cref="SshConnectionException">Client is not connected.</exception>
+        /// <exception cref="SftpPermissionDeniedException">Permission to perform the operation was denied by the remote host. <para>-or-</para> A SSH command was denied by the server.</exception>
+        /// <exception cref="SshException">A SSH error where <see cref="Exception.Message"/> is the message from the remote host.</exception>
+        /// <exception cref="ObjectDisposedException">The method was called after the client was disposed.</exception>
+		public void EnsureDirectory(string path)
+		{
+			CheckDisposed();
+
+			if (path.IsNullOrWhiteSpace())
+				throw new ArgumentException("path");
+
+			if (_sftpSession == null)
+				throw new SshConnectionException("Client not connected.");
+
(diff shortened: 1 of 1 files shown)


src/Renci.SshNet/SftpClient.cs
```diff
@@ -605,6 +605,37 @@ public bool Exists(string path)
...
+		{
+			CheckDisposed();
+
+			if (path.IsNullOrWhiteSpace())
+				throw new ArgumentException("path");
+
+			if (_sftpSession == null)
+				throw new SshConnectionException("Client not connected.");
+
+			string[] splitDirNames = path.Replace("\\", "/").Split('/').Where(s => !s.IsNullOrWhiteSpace()).ToArray();
```
DanielVukelich, line 628: " " (and other all-whitespace combinations) is a valid directory name on Linux.  Changing your `Where()` lambda to `!string.IsNullOrEmpty(s)` would be better.

src/Renci.SshNet/SftpClient.cs
```diff
@@ -605,6 +605,37 @@ public bool Exists(string path)
...
+
+			List<string> rollupDirectories = new List<string>();
+			for (int i = 2; i <= splitDirNames.Length; i++)
+				rollupDirectories.Add(string.Join("/", splitDirNames.Where(sdir => !sdir.IsNullOrWhiteSpace()).Take(i).ToArray()));
```
DanielVukelich, line 632: Optional: Since splitDirNames is was already stripped of null entries when it was created, the .Where() invocation here is unnecessary

src/Renci.SshNet/SftpClient.cs
```diff
@@ -605,6 +605,37 @@ public bool Exists(string path)
...
+
+			foreach (string dir in rollupDirectories)
+				if (!this.Exists(dir))
```
DanielVukelich, line 635: Optional, but this work could all be done in the loop where you append elements to rollupDirectories, thus removing the need to have rollupDirectories in the first place

DanielVukelich (changes requested):

src/Renci.SshNet/SftpClient.cs
kstolte: Good point. I’ll adjust.

src/Renci.SshNet/SftpClient.cs
kstolte: Yep. I’ll adjust that here.

WojciechNagorski:
Old PR. Currently we have `public bool Exists(string path)`

PR #333, opened 2017-11-01
Add support for .NET Standard 2.0

drieseng:
Adds a .NET Standard 2.0 assembly to the nuget and zip distribution package.


ScottRFrost:
Looking forward to the .net standard 2.0 build.  I can't upgrade my app to .net core 2.0 until SSH.net updates and I want to play with some of the new stuff!

ponmudivn:
is it released already? can use in .net core 2.0 project?

Issue #351, opened 2017-12-06
Call to md5 has algorithm is made when not used


Sebastma:
In ConnectionInfo.cs contructor a call to prepare all supported hash algorithms is made. this violates a securtity setting on some servers that does not allow MD5 hash algorithms. This causes the program to exit

WojciechNagorski:
The issue is already fixed.

PR #352, opened 2017-12-06
Fixes issue #351

Sebastma:
Fixes issue #351 where a call to a hashalgorithm is made for each supported hash algorithm that can triggure security  breaches on som servers not permitting the use of certain has alogirthms. Each call is now made only when the actual hashalgorithm is used


gnab:
👍

WojciechNagorski:
The issue is already fixed.

PR #356, opened 2017-12-13
Fix SFTP file UTC time handling

menees:
I changed SftpFileAttributes to preserve the last access and modified times as UTC values rather than always forcing a local conversion.  This primarily affected the FromBytes(SshDataStream) method, but the changes cascade out from there to make sure that SftpFileAttributes and SftpFile can be used without ever forcing a (potentially lossy) conversion to local time.  In the SSH data stream the atime and mtime values are "Unix times" in seconds, so they're already in UTC.  We need to preserve that fact by returning and storing DateTime values with Kind == Utc.

Previously, the code called DateTime.FromFileTime.  Internally, FromFileTime calls DateTime.ToLocalTime(), which is a lossy function.  For example, any local time >= 1am and < 2am during the "fall back" hour of a daylight saving time shift in the fall could have come from two different UTC times.  Avoiding this ambiguity is why file systems store times in UTC, so the SFTP classes need to preserve UTC times too.  Here's a simple example to see this in VS's C# interactive window:

```
TimeZoneInfo central = TimeZoneInfo.FindSystemTimeZoneById("Central Standard Time");
DateTime first = new DateTime(2017, 11, 5, 6, 0, 0, DateTimeKind.Utc);
DateTime second = new DateTime(2017, 11, 5, 7, 0, 0, DateTimeKind.Utc);
Console.WriteLine(TimeZoneInfo.ConvertTimeFromUtc(first, central));
Console.WriteLine(TimeZoneInfo.ConvertTimeFromUtc(second, central));
// Both UTC times convert to the same local time: 11/5/2017 1:00:00 AM
```

.NET also has an issue where it caches the current local time zone offset, so it can incorrectly calculate local times after a DST shift if nothing has called CultureInfo.ClearCachedData or TimeZoneInfo.ClearCachedData.  This is discussed more at [https://stackoverflow.com/a/297189/1882616](https://stackoverflow.com/a/297189/1882616), and it can be a big problem in long-running services if not handled correctly.  By making SSH.NET internally preserve UTC times correctly, the library can be immune to problems with DST shifts for callers that only work with UTC times.

I also cleaned up some warnings in the unit tests and added an .editorconfig file so VS 2017 will know to use spaces for indentation in this solution (even though my global preference is set to tabs).


stefer:
This is not only "improve" - I would say it is "correct" UTC time handling, see https://github.com/sshnet/SSH.NET/issues/415

menees:
@drieseng Is there any way you could bump up the priority on merging in this pull request please? I originally said "improved", but I actually agree with @stefer that this is the **correct** UTC time handling. I'd really like to stop having to maintain and ship my own fork because correct UTC time support is extremely important in my app. Thanks!

src/Renci.SshNet/Sftp/SftpFileAttributes.cs
```diff
@@ -397,8 +435,8 @@ private SftpFileAttributes()
 
         internal SftpFileAttributes(DateTime lastAccessTime, DateTime lastWriteTime, long size, int userId, int groupId, uint permissions, IDictionary<string, string> extensions)
         {
-            LastAccessTime = _originalLastAccessTime = lastAccessTime;
-            LastWriteTime = _originalLastWriteTime = lastWriteTime;
+            LastAccessTimeUtc = _originalLastAccessTimeUtc = ToUniversalTime(lastAccessTime);
+            LastWriteTimeUtc = _originalLastWriteTimeUtc = ToUniversalTime(lastWriteTime);
```
drieseng: If both the *lastAccessTime* and *lastWriteTime* argument now hold UTC times (since you used `DateTime.FromFileTimeUtc` to construct them), then why convert to UTC again?

drieseng:
@bmenees I first want to have an integration test to validate these changes. I hope to have time for that this weekend.

src/Renci.SshNet/Sftp/SftpFileAttributes.cs
menees: It's [defensive programming](https://en.wikipedia.org/wiki/Defensive_programming) since I can't guarantee that I'll be passed DateTime values where Kind == UTC. The primary caller of the internal SftpFileAttributes constructor _does_ pass UTC times now, and in that case the ToUniversalTime method won't change the value.

However, several callers from the Renci.SshNet.Tests assembly (which is allowed to access that internal constructor via the InternalsVisibleToAttribute) pass in non-UTC DateTimes (e.g., DateTime.MinValue and DateTime.Now). Those cases have Kind equal to Unspecified and Local, respectively. Calling ToUniversalTime in the SftpFileAttributes constructor is required to handle those correctly. You can use Visual Studio's "Find All References" on the internal SftpFileAttributes constructor to see these cases.

So, the ToUniversalTime method is a safety check to ensure both the production cases and the unit test cases all convert times consistently to UTC.

menees:
Thanks! I really appreciate it and all of your hard work on this project!

menees:
@drieseng Have you had a chance to validate these changes? I'd really like to get this pull request merged in. Thanks!

drieseng:
@menees Sorry to inform you, but our policy is to not accept PRs from developers that prefer tabs over spaces.

drieseng:
We may reconsider :)

drieseng:
I'm **really** gonna try to find time tomorrow to create an integration test for this PR.
Sorry it has taken this much time. If you could swap for a week, you'd surely understand.

menees:
Thanks. Regarding tabs vs. spaces, all of my changes in this PR use spaces for indentation. That's _why_ I added the .editorconfig file. My "global" preference in VS is tabs, but the .editorconfig I added at the solution-level makes VS use and respect the solution-level setting of spaces (instead of my or anyone else's global preference).

So, that's another advantage to merging this PR. In the future, anyone editing the SSH.NET code in VS 2017 or later (or [any of the 20+ other editors that support .editorconfig](https://editorconfig.org/#download)) will automatically use spaces for indentation since that's what the .editorconfig declares as the solution's preference. Then you'll have fewer PRs using tabs (or none, hopefully).

drieseng:
@menees I should've been more explicit: the fact that I closed the PR due to the tabs vs. spaces battle was a joke!

Can you move the .editorconfig to a separate PR? I'd like to review/discuss the config (and again, I'm not talking about tabs vs. spaces) before we merge it.

menees:
I deleted .editorconfig from this PR because GitHub makes that easy. I really only need the fixes for UTC time handling merged in.

src/Renci.SshNet/Sftp/SftpFileAttributes.cs
```diff
@@ -397,8 +435,8 @@ private SftpFileAttributes()
 
         internal SftpFileAttributes(DateTime lastAccessTime, DateTime lastWriteTime, long size, int userId, int groupId, uint permissions, IDictionary<string, string> extensions)
         {
-            LastAccessTime = _originalLastAccessTime = lastAccessTime;
-            LastWriteTime = _originalLastWriteTime = lastWriteTime;
+            LastAccessTimeUtc = _originalLastAccessTimeUtc = ToUniversalTime(lastAccessTime);
```
drieseng: Please remove the `ToUniversal()`as there's no need to convert to UTC here.
I propose to rename the arguments from **last\*Time** to **last\*TimeUtc**.

If there are tests that depend on this, these tests should be updated.

src/Renci.SshNet/Sftp/SftpFileAttributes.cs
drieseng: Please remove the `ToUniversalTime()`, and rename the arguments from **last\*Time** to **last\*TimeUtc**.
If necessary, the tests should be adapted.

src/Renci.SshNet/Sftp/SftpFileAttributes.cs
```diff
@@ -544,10 +582,12 @@ internal static SftpFileAttributes FromBytes(SshDataStream stream)
...
             if ((flag & 0x00000008) == 0x00000008)   //  SSH_FILEXFER_ATTR_ACMODTIME
             {
+                // The incoming times are "Unix times", so they're already in UTC.  We need to preserve that
+                // to avoid losing information in a local time conversion during the "fall back" hour in DST.
                 var time = stream.ReadUInt32();
-                accessTime = DateTime.FromFileTime((time + 11644473600) * 10000000);
+                accessTime = DateTime.FromFileTimeUtc((time + 11644473600) * 10000000);
                 time = stream.ReadUInt32();
-                modifyTime = DateTime.FromFileTime((time + 11644473600) * 10000000);
+                modifyTime = DateTime.FromFileTimeUtc((time + 11644473600) * 10000000);
```
drieseng, line 591: Please remove the default value of **DateTime.MinValue**, and add an else block in which **accessTime** is initialized to `DateTime.SpecifyKind(DateTime.MinValue, DateTimeKind.Local)`.

src/Renci.SshNet/Sftp/SftpFileAttributes.cs
```diff
@@ -572,5 +612,37 @@ internal static SftpFileAttributes FromBytes(byte[] buffer)
...
+            }
+            else
+            {
+                result = value.ToLocalTime();
+            }
+
+            return result;
+        }
+
+        internal static DateTime ToUniversalTime(DateTime value)
```
drieseng: This method can be removed.

src/Renci.SshNet/Sftp/SftpFileAttributes.cs
```diff
@@ -572,5 +612,37 @@ internal static SftpFileAttributes FromBytes(byte[] buffer)
                 return FromBytes(stream);
             }
         }
+
+        internal static DateTime ToLocalTime(DateTime value)
```
drieseng: Please change to private.

drieseng (changes requested):

src/Renci.SshNet/Sftp/SftpFileAttributes.cs
menees: Done.

src/Renci.SshNet/Sftp/SftpFileAttributes.cs
menees: Done, but I specified DateTimeKind.Utc from the else block.

src/Renci.SshNet/Sftp/SftpFileAttributes.cs
menees: Done.

src/Renci.SshNet/Sftp/SftpFileAttributes.cs
menees: I kept ToUniversalTime and made it private. It's still needed by the setters for the LastAccessTime and LastWriteTime properties. Since those are public properties in a public class, it's probable that some code somewhere calls them and passes in DateTime values with Kind == Local or Unspecified.

drieseng (approved):

drieseng:
@menees, thanks!!

PR #360, opened 2017-12-22
Do not lock on this

UselessToucan:


drieseng (changes requested):
Please undo any code formatting changes, and include a motivation in the PR description and commit message.


Something in the line of "Use private lock object instead of 'this'. This avoids publicly exposing the lock object."

drieseng:
@UselessToucan Will you submit a new PR?

PR #361, opened 2017-12-22
Dispose IDisposable object before method return

UselessToucan:


src/Renci.SshNet/Sftp/SftpFileAttributes.cs
```diff
@@ -442,73 +442,75 @@ public void SetPermissions(short mode)
         /// </returns>
         public byte[] GetBytes()
         {
-            var stream = new SshDataStream(4);
+            using (var stream = new SshDataStream(4))
+            {
 
-            uint flag = 0;
```
drieseng, line 447: Please undo all code formatting changes.

drieseng:
@UselessToucan Will you submit a new PR?

PR #374, opened 2018-01-19
Replace Queue<byte> based buffer by array based buffer in PipeStream.

zwergenaufstand:
This replaces the rather slow Queue<byte> based PipeStream by an array based one. The new PipeStream uses a fixed size array as ring buffer instead of a size limited queue. The default buffer size is 1 MiB. Apart from that it's a drop-in replacement for the current PipeStream.
The array based PipeStream is quite a bit faster than the current one. On my machine (Core i7 4790) in a single-producer/single-consumer scenario I can read and write over 6 GB / second. The current PipeStream's speed is about 65 MB / second.

### DIFF
--- a/src/Renci.SshNet.Tests/Classes/Common/PipeStreamTest.cs
+++ b/src/Renci.SshNet.Tests/Classes/Common/PipeStreamTest.cs
@@ -175,18 +175,6 @@ namespace Renci.SshNet.Tests.Classes.Common
             Assert.AreEqual(0L, target.Length);
         }
 
-        /// <summary>
-        ///A test for MaxBufferLength
-        ///</summary>
-        [TestMethod]
-        public void MaxBufferLengthTest()
-        {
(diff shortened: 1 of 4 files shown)


scottstephens:
Have you seen this lead to improvements in overall download speed? I've been using SftpClient.DownloadFile and have been getting about 28 Mbps (3.5 MB/s), vs 100 Mbps using the scp command line. I'm wondering if this could help.

EDIT: Updated to 2016.1.0 and that got me up to 66 Mbps.

zwergenaufstand:
> Have you seen this lead to improvements in overall download speed? I've been using SftpClient.DownloadFile and have been getting about 28 Mbps (3.5 MB/s), vs 100 Mbps using the scp command line. I'm wondering if this could help.

No, the SftpClient does not use the PipeStream at all. Only SshClient and ScpClient use it.

PR #388, opened 2018-02-27
Allow extending authentication without forking

withinboredom:
Currently, the only way to handle custom authentication mechanisms is via forking the repository and maintaining a fork. I'd like to propose making some classes and functions public, so they can be inherited/extended for custom solutions instead of requiring a fork.


WojciechNagorski:
PR needs refreshing. If you feel this change is needed feel free to open PR.

PR #396, opened 2018-03-04
Fix #395 : SftpClient Enumerates Rather Than Accumulates Directory Items

### LINKED ISSUE #395: SftpClient Enumerates Rather Than Accumulates Directory Items
Hi,
_Thank you for really great library! Thank you once again!_

I've a proposal to enhance (fix) some behavior which leads to delayed processing and huge memory footprints. They are actually our current production issues caused by 65 thousand directory items in single parent directory. Our consumer service has to list all of them and then process one by one or in batches. The current SftpClient code basically reads all the items and make SftpFile objects of it. That generates a peak in memory usage which is still acceptable. However, the criteria or behavior which is not acceptable is the time required to make the list of the folder items. In our scenario, **it takes 1 minute and 45 seconds**. The distance is between London and New York for production environment. The service we maintain and which uses SSH.NET has to check the remote directory as frequent as possible at max by 1 minute. In my test environment between West Europe and West US the listing operation lasts 2 minutes and 6 seconds over 75 thousand folder items.

```javascript
            var files = _sftpSession.RequestReadDir(handle);

            while (files != null)
            {
                result.AddRange(from f in files
                                select new SftpFile(_sftpSession, string.Format(CultureInfo.InvariantCulture, "{0}{1}", basePath, f.Key), f.Value));

                //  Call callback to report number of files read
                if (listCallback != null)
                {
                    //  Execute callback on different thread
                    ThreadAbstraction.ExecuteThread(() => listCallback(result.Count));
                }

                files = _sftpSession.RequestReadDir(handle);
            }
```

There is nothing actually to do with the principle on listing the directory into items but how the library handles result data. Currently, it accumulates all items and then passes them to consumer as one big list. The enhancement should be about to introduce a new method which does real enumeration. This would merge/join the time spent on listing and consuming the items and pushes it directly to consumer. The consumer's logic may decide - ah, we've got enough items to process so we can stop enumerating (i.e. in the middle or at the first 150 items). This scenario actually shorten the processing time to acceptable minimum.

Obviously, I've looked into the source code, have done the change, tested it and thought about multithreading scenarios. I wish it to be merged in if you agree on.

Many thanks,
Pavel

Mine proposal is about to add dedicated new methods into SftpClient which does enumeration:
```javascript
        /// <summary>
        /// Enumerates files and directories in remote directory.
        /// </summary>
        /// <remarks>
        /// This method differs to <see cref="ListDirectory(string, Action{int})"/> in the way how the items are returned. 
        /// It yields the items to the last moment for the enumerator to decide if it needs to continue or stop enumerating the items. 
        /// It is handy in case of really huge directory contents at remote server - meaning really huge 65 thousand files and more.
        /// It also decrease the memory footprint and avoids LOH allocation as happen per call to <see cref="ListDirectory(string, Action{int})"/> method.
        /// There aren't asynchronous counterpart methods to this because enumerating should happen in your specific asynchronous block.
        /// </remarks>
        /// <param name="path">The path.</param>
        /// <param name="listCallback">The list callback.</param>
        /// <returns>
        /// An <see cref="System.Collections.Generic.IEnumerable{SftpFile}"/> of files and directories ready to be enumerated.
        /// </returns>
        /// <exception cref="ArgumentNullException"><paramref name="path" /> is <b>null</b>.</exception>
        /// <exception cref="SshConnectionException">Client is not connected.</exception>
        /// <exception cref="SftpPermissionDeniedException">Permission to list the contents of the directory was denied by the remote host. <para>-or-</para> A SSH command was denied by the server.</exception>
        /// <exception cref="SshException">A SSH error where <see cref="Exception.Message" /> is the message from the remote host.</exception>
        /// <exception cref="ObjectDisposedException">The method was called after the client was disposed.</exception>
        public IEnumerable<SftpFile> EnumerateDirectory(string path, Action<int> listCallback = null)`

        /// <summary>
        /// Internals the list directory.
        /// </summary>
        /// <param name="path">The path.</param>
        /// <param name="listCallback">The list callback.</param>
        /// <returns>
        /// A list of files in the specified directory.
        /// </returns>
        /// <exception cref="ArgumentNullException"><paramref name="path" /> is <b>null</b>.</exception>
        /// <exception cref="SshConnectionException">Client not connected.</exception>
        private IEnumerable<SftpFile> InternalEnumerateDirectory(string path, Action<int> listCallback)
```

znamenap:
Hi @drieseng ,

Please, could you check this pull request for #395 ?

The enumerating directory approach has better time results on very huge directorries (a lot of files meaning 80000). In many situations happened the call timed out while using the actual method ListDirectory.

I've also tested in my test environment and the same tests as like in ListDirectory.

Thanks.
Pavel

### DIFF
--- a/src/Renci.SshNet.IntegrationTests/OldIntegrationTests/SftpClientTest.EnumerateDirectory.cs
+++ b/src/Renci.SshNet.IntegrationTests/OldIntegrationTests/SftpClientTest.EnumerateDirectory.cs
@@ -1,0 +1,209 @@
+﻿using System.Diagnostics;
+
+using Renci.SshNet.Common;
+
+namespace Renci.SshNet.IntegrationTests.OldIntegrationTests
+{
+    /// <summary>
+    /// Implementation of the SSH File Transfer Protocol (SFTP) over SSH.
+    /// </summary>
+    public partial class SftpClientTest : IntegrationTestBase
+    {
+        [TestMethod]
+        [TestCategory("Sftp")]
+        [ExpectedException(typeof(SshConnectionException))]
+        public void Test_Sftp_EnumerateDirectory_Without_Connecting()
+        {
+            using (var sftp = new SftpClient(SshServerHostName, SshServerPort, User.UserName, User.Password))
+            {
+                var files = sftp.EnumerateDirectory(".");
+                foreach (var file in files)
+                {
+                    Debug.WriteLine(file.FullName);
+                }
+            }
+        }
+
+        [TestMethod]
+        [TestCategory("Sftp")]
+        [TestCategory("integration")]
+        [ExpectedException(typeof(SftpPermissionDeniedException))]
+        public void Test_Sftp_EnumerateDirectory_Permission_Denied()
+        {
+            using (var sftp = new SftpClient(SshServerHostName, SshServerPort, User.UserName, User.Password))
+            {
+                sftp.Connect();
+
+                var files = sftp.EnumerateDirectory("/root");
+                foreach (var file in files)
+                {
+                    Debug.WriteLine(file.FullName);
+                }
+
+                sftp.Disconnect();
+            }
+        }
+
+        [TestMethod]
+        [TestCategory("Sftp")]
+        [TestCategory("integration")]
+        [ExpectedException(typeof(SftpPathNotFoundException))]
+        public void Test_Sftp_EnumerateDirectory_Not_Exists()
+        {
+            using (var sftp = new SftpClient(SshServerHostName, SshServerPort, User.UserName, User.Password))
+            {
+                sftp.Connect();
+
+                var files = sftp.EnumerateDirectory("/asdfgh");
+                foreach (var file in files)
+                {
+                    Debug.WriteLine(file.FullName);
+                }
+
+                sftp.Disconnect();
+            }
+        }
+
+        [TestMethod]
+        [TestCategory("Sftp")]
+        [TestCategory("integration")]
+        public void Test_Sftp_EnumerateDirectory_Current()
+        {
+            using (var sftp = new SftpClient(SshServerHostName, SshServerPort, User.UserName, User.Password))
+            {
+                sftp.Connect();
+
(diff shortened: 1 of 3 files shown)


Alexinio97:
Hi,
I'm also interested in these changes, is there a chance for this to get merged soon?

znamenap:
This PR is ready to review and approve potentially.

Rob-Hague:
I like the change, but I don't think we should add a new method for it. Instead, we should just change `ListDirectory` to yield. We now have `ListDirectoryAsync` which yields.

I also think we should remove the `listCallback` parameter from `ListDirectory`: yielding will make it less useful, but also executing an arbitrary callback on the threadpool during enumeration is unnecessary complication and not very canonical of an `IEnumerable` method (it seems to be a leftover of the Begin/End APM methods)

What do you think @WojciechNagorski ?

znamenap:
Hi @WojciechNagorski , please put this PR into the queue to review and potentially merge it.

znamenap:
My two cents two the discussion regarding changing the internal implementation of the ListDirectory, changing it to use the yield return will certainly break the existing consumer's business logic with the unwritten behavior of returing a complete list in fact. Many of the consumers had already implemented logic such way they are in fact expecting the full list. Therefore, I'd rather suggest changing ListDirectory to return IList<SftpFile> interface as part of the major breaking changes and do not introduce the yield return at all. If consumers would like to change the behavior, they can simply change the call from ListDirectory to EnumerateDirectory.
The open-ended enumeration is a contract to the runtime of the consumer's application context created with the SftpConnection, people may be reporting issues because of reaching the out of the scope for the connection and request enumerating/listing the directory. We should consider this option also.

znamenap:
Regarding the callback, I second that idea, the thread pool may be in excess. I presume the callback was added in order to report the progress and hence, I'm suggesting this may be replaced with ```IProgress<T>```. Plus, I think, this change should be out-scoped to this PR and reported as another issue to solve later.

CurtisMonroe (approved):

PR #425, opened 2018-04-23
Update the verify host identify example in the README

Kim-SSi:
Update the verify host identify example to use Enumerable.SequenceEqual from System.Linq.


drieseng:
It would definitely make the example much more consice, but I don't want to encourage **System.Linq**.

PR #428, opened 2018-05-02
Make SemaphoreLight AuthenticationConnection in Session.cs not static

Didjeeh:
In doing so we avoid needless waiting on a semaphore to signal if you
have more than 3 simultaneous SSH clients in your application.

### DIFF
--- a/src/Renci.SshNet/Session.cs
+++ b/src/Renci.SshNet/Session.cs
@@ -94,7 +94,7 @@ namespace Renci.SshNet
         /// <remarks>
         /// Some server may restrict number to prevent authentication attacks
         /// </remarks>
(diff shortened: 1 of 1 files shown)


keithallenjackson:
Wow, this might explain my low CPU usage and longer than expected wait times when connecting to 20+ devices at a time. Is the semaphore still needed at all?

Didjeeh:
It is a security measurement. Personally, I would leave it there as a non-static field.

WojciechNagorski:
It's done on master

Didjeeh:
Nice. Thanks. Best wishes.

WojciechNagorski:

PR #432, opened 2018-05-10
add disconnect and close events

bryceg:
The close event is needed to know if the stream is disconnected on the target side for any reason


keithallenjackson:
@bryceg I like your pull request! Seems I might have overlapped some of your efforts unintentionally in my pull request #458. I added a Disconnect event with tests among other fixes. I like your (cleaner) implementations better. Would you be willing to take my tests and I remove the duplicated work from my PL?

bryceg:
@keithallenjackson yeah, sounds good to me.

PR #438, opened 2018-06-08
Add support for openssh known_hosts files

DanielVukelich:
We're moving our systems to .NET Core, and one of our dependencies is an incompatible sftp library.  SSH.NET would be a perfect drop-in replacement, except that the sftp library supports verifying hosts through an openssh known_hosts file.

This pull request implements parsing and export for known_host files, and is compatible with all behavior specified in [sshd's manpage entry](https://www.freebsd.org/cgi/man.cgi?sshd(8)#SSH_KNOWN_HOSTS%09FILE_FORMAT)  with the exception of certificate authority signing.

Example usage:
```
using Renci.SshNet;
using Renci.SshNet.Common;
using System;

namespace SftpTest
{
    class Program
    {
        private static KnownHostStore _knownHosts;

        static void Main(string[] args)
        {
            _knownHosts = new KnownHostStore("~/.ssh/known_hosts");
            using (SftpClient client = new SftpClient("test.rebex.net", 22, "demo", "password"))
            {
                client.HostKeyReceived += (sender, eventArgs) =>
                {
                    eventArgs.CanTrust = CanTrustHost(client.ConnectionInfo.Host, eventArgs);
                };

                try
                {
                    client.Connect();
                    Console.WriteLine("Connected!");
                }
                catch (Exception ex)
                {
                    Console.WriteLine("Failed to connect: " + ex.Message);
                }
            }

            _knownHosts.ExportKnownHostsToFile("known_hosts.txt");
            Console.WriteLine("Press any key to continue");
            Console.ReadKey();
        }

        private static bool CanTrustHost(string hostname, HostKeyEventArgs e)
        {
            if (_knownHosts.Knows(hostname, e.HostKeyName, e.HostKey, 22))
                return true;

            Console.WriteLine("Hostname:\t" + hostname);
            Console.WriteLine("KeyType:\t" + e.HostKeyName);
            Console.WriteLine("Fingerprint:\t" + BitConverter.ToString(e.FingerPrint));
            Console.WriteLine(Environment.NewLine + "Unknown host.  Trust it?  y/N: ");

            int choice = Console.Read();
            if (choice != 'y' && choice != 'Y')
                return false;

            _knownHosts.AddHost(hostname, 22, e.HostKeyName, e.HostKey, false);
            return true;
        }
    }
}
```

### DIFF
--- a/src/Renci.SshNet.Tests/Classes/KnownHostStoreTests.cs
+++ b/src/Renci.SshNet.Tests/Classes/KnownHostStoreTests.cs
@@ -1,0 +1,52 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Renci.SshNet.Common;
+
+namespace Renci.SshNet.Tests.Classes
+{
+    [TestClass]
+    public class KnownHostStoreTests
+    {
+        private static readonly byte[] ExampleKey = new byte[]
+        {
+            0x00, 0x11, 0x22, 0x33, 0x44,
+            0x55, 0x66, 0x77, 0x88, 0x99,
+            0xaa, 0xbb, 0xcc, 0xdd, 0xee,
+            0xff
+        };
+
+        private const string ExampleHost = "some.domain.com";
+        private const string ExampleSubDomain = "*.domain.com";
+        private const string ExampleKeyType = "rsa";
+
+        [TestMethod]
+        public void TestKnownHostStoreFindsHostIfMatch()
+        {
+            KnownHostStore hostStore = new KnownHostStore();
+            hostStore.AddHost(ExampleHost, 22, ExampleKeyType, ExampleKey, false);
+            Assert.IsTrue(hostStore.Knows(ExampleHost, ExampleKeyType, ExampleKey, 22));
+        }
+
+        [TestMethod]
+        public void TestKnownHostStoreFindsHashedHostIfMatch()
+        {
+            KnownHostStore hostStore = new KnownHostStore();
+            hostStore.AddHost(ExampleHost, 22, ExampleKeyType, ExampleKey, true);
+            Assert.IsTrue(hostStore.Knows(ExampleHost, ExampleKeyType, ExampleKey, 22));
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(RevokedKeyException), "Did not throw an exception upon finding a revoked key in the store")]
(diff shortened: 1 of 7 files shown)


BorisAmelyan (commented):
can we release this support for OPENSSH?
looks like someone already put an effort to add this feature

avr2222:
Open SSH Known hosts still not supported?

A9G-Data-Droid:
To avoid conflicts with efforts to make this library work on FIPS enabled systems we have to be careful to use approved hash algorithms. `HMACSHA1` was recently deprecated by both the government and Microsoft.

> Due to collision problems with SHA1, Microsoft recommends a security model based on SHA256 or better.

https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.hmacsha1.-ctor?view=net-6.0#system-security-cryptography-hmacsha1-ctor(system-byte()-system-boolean)

The new versions of HMACSHA256, HMACSHA384, and HMACSHA512 are all FIPS compliant now.

https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/w8h3skw9(v=vs.100)?redirectedfrom=MSDN

I would use `HMACSHA256` at a minimum to avoid deprecation warnings.

```C#
HMACSHA256 hmac = new HMACSHA256(_hashSalt);
```

DanielVukelich:
From a compatibility standpoint, sha1 is the only hashing algorithm that I can find is supported by tools like OpenSSH.  As far as I can tell, there is no way for a known hosts entry to specify which algorithm it uses for hashing, so everyone assumes sha1.  If the goal is to support known_hosts files from any source, then the hashing would have to be sha1.

I can understand the concern with FIPs compliance, but from a security standpoint the hash is only used to obscure the hostnames from a human attacker who tries to enumerate all the known hosts.  I'm not a cryptographer, but I think in this case that collision attacks would be meaningless, since any colliding hostname would necessarily be different from the actual hostname that the known_hosts entry is valid for, so there's no reason for an attacker to try and connect to the collided host.  Again though, I know that FIPS is more about compliance than actual security.

Maybe if there are some FIPS compile flags, this could be selectively supported?  Otherwise, support for hashed hosts should be removed.  It's been years since I wrote this commit.

A9G-Data-Droid:
Thanks for the thoughtful response. I think the solution is to have an option to turn off hashed known_hosts. At closer inspection of your code, it looks like it does support plaintext known_hosts. So I apologize for the detour.

Looks like people are moving away from hashing known_hosts because it doesn't have any security value:
https://security.stackexchange.com/questions/56268/ssh-benefits-of-using-hashed-known-hosts

PR #468, opened 2018-09-16
Add support for PuTTY ppk format private key files

rmini:


darkoperator:
Nice!!!! How this PR makes it in to the next beta

darinkes:
This PR is missing ecdsa and ed25519 keys. DSS keys support should be dropped.
Tests are also missing.

There is also this SSH.NET Extension: https://github.com/darinkes/SshNet.PuttyKeyFile

IgorMilavec:
@darinkes what are your plans with ```SshNet.PuttyKeyFile``` extension? Will you contribute it to SSH.NET? I kinda like the idea of separating out different file format support into different packages. If you decide to go down that route, could you release the extension on NuGet to make it easier for developers to find and use?

darinkes:
@IgorMilavec Thats the plan. SSH.NET offers interfaces so ppl can add extensions seperated from the core.
Especially Auth-Options can benefit from this, since there are sooo many of them.

Started some Extension, based on a custom IPrivateKeyFile-Extension, which I would like to discuss:

https://github.com/darinkes/SshNet.Agent
https://github.com/darinkes/SshNet.Keygen
https://github.com/darinkes/SshNet.PuttyKeyFile

Some of them are already on nuget.org: https://www.nuget.org/profiles/darinkes
But they need changes in ssh.net core, so nothing announced yet.

Maybe they should also be maintained on the sshnet-Github Account to show they are trustworthy?

IgorMilavec:
I propose to first implement support for extensibility, so external "key providers" can be built without forking. After that I think it would be great if we could bring the most relevant (= usable to most developers) extensions under sshnet maintenance.

IgorMilavec:
@mini #794 has been merged, can you please check how it affects this PR?

mini:
No idea! :P

Kim-SSi:
The is a project that adds ppk support this but it depends on the latest develop.
It is @darinkes project [SshNet.PuttyKeyFile](https://github.com/darinkes/SshNet.PuttyKeyFile)
I have opened a pull request on to add ppk v3 support https://github.com/darinkes/SshNet.PuttyKeyFile/pull/2
Although this does require net46 or .netstandard20 for encrypted ppk v3.

Rob-Hague:
I'll close this since as above, https://github.com/darinkes/SshNet.PuttyKeyFile is a more complete implementation.

PR #494, opened 2018-11-29
Add the ability to read/sample N lines of a file

finleysg:
We had the need to sample large files from a UI without pulling the entire file across the wire. Thought I'd submit this as a PR in case anyone else might find that useful.


Rob-Hague:
I'm closing this as old and undesired - and simple enough as an extension method.

Issue #394, opened 2018-03-02
Can't delete file right after creating them


benedicttobias:
I am wondering if this is an issue or my process flaw. I am not be able to delete a file that just created. But, on WinSCP, I am able to delete them manually.

There is a case where the file got locked and I can't delete them because it says that "Permission Denied" and I can't delete them as SFTP user or OS (windows) user. I did checked that the file is owned by Administrator.

Here is the code that I use to produce the problem.

```
private static void uploadZip(string fTPAddress, string ftpPath, string fTPUsername, string fTPPassword, string zipFilePath, string zipFileName) {
	// Build username/password
	var passwordAuth = new PasswordAuthenticationMethod(fTPUsername, fTPPassword);

	// Build connection
	var connectionInfo = new ConnectionInfo(fTPAddress, fTPUsername, passwordAuth);

	using (var client = new SftpClient(connectionInfo)) {
		Console.WriteLine("Connecting " + fTPAddress + "... ");
		client.Connect();
		Console.WriteLine(client.IsConnected ? "Connected!" : "Disconnected.");

		// Upload file
		try {
			Console.WriteLine("Creating new path + file...");
			client.Create(ftpPath + zipFileName);
			Console.WriteLine("Created!");

			if (client.Exists(ftpPath + zipFileName)) {
				Console.WriteLine(ftpPath + zipFileName + " Exists!");
				Console.WriteLine("Deleting " + ftpPath + zipFileName + "... ");
				client.Delete(ftpPath + zipFileName);
				Console.WriteLine("Deleted");
			} else {
				Console.WriteLine("Uploading file...");
				using (FileStream stream = File.Open(zipFilePath, FileMode.Open)) {
					client.UploadFile(stream, ftpPath + zipFileName);
				}
				Console.WriteLine("Uploaded!");
			}
		} catch (Exception e) {
			Console.WriteLine("-- Something wrong on execution: " + e.Message);
			Console.WriteLine(e.StackTrace);
		} finally {
			Console.WriteLine("Disconnecting... ");
			client.Disconnect();
			Console.WriteLine(client.IsConnected ? "Connected!" : "Disconnected.");

			Console.WriteLine("Deleting " + zipFilePath + "...");
			File.Delete(zipFilePath);
			Console.WriteLine(zipFilePath + " Deleted...");
		}
	}
}
```

**The output:**
[...]

Issue #327, opened 2017-10-20
'No Such File' during client.CreateDirectory(path)


AlanGreyjoy:
I am trying to create directories in our CentOS server and I get thrown an exception. 'No Such File'

Now, this doesn't happen all the time, only sometimes. But when it starts, it doesn't go away for a couple of days.

savePath = "/PhoneManager/Help/Pages/Phones/Adding_Extensions/";
if (!client.Exists(savePath))
                        {
                            client.CreateDirectory(savePath);
                        }

drieseng:
Are you sure it's only the last part of the path that does not exist?
This may not be the case for all SFTP servers, but in general `CreateDirectory(String)` (**SSH_FXP_MKDIR**) will only create the specified directory. It will not create any missing parent directories.

**For example:**
If you attempt to create **"/PhoneManager/Help/Pages/Phones/Adding_Extensions"**, it will only succeed if the **"/PhoneManager/Help/Pages/Phones"** path already exists.

jehrenzweig:
@AlanGreyjoy I suspect it's because `SftpClient.Exists()` is bugged and cannot reliably determine if a directory exists when the path begins with a forward-slash character (`/`).  Your `savePath` value (`/PhoneManager/Help/Pages/Phones/Adding_Extensions/`) does appear to fit that use-case. Could you do me a favor and try using the same value without the leading slash (i.e. `PhoneManager/Help/Pages/Phones/Adding_Extensions/`) and see if that workaround fixes things for you?

drieseng:
No feedback.

EngrKhizarIqbal:
Hi, is there a way to create nested directories?

PR #495, opened 2018-11-30
Fix SftpClient.Exists() to work with forward-slash prefix notation

ghost:
Yesterday, I discovered that SftpClient.Exists() would return `false` when checking certain directory paths that did actually exist.  It appears that the presence of a forward-slash character (`/`) in a directory path (e.g. `/MyDir`, `/ParentDir/ChildDir`) will trigger this behavior; removing this character seems to make the method work properly.

**Summary of Changes**
* Update `SftpClient.Exists()` method to ignore forward-slash character if it appears at the start of a directory path.

* Add unit tests to help confirm the existence of the bug & the effectiveness of the fix.

**Related Issues**
* #327
* #394


ghost:
@drieseng Could you take a look at this PR, whenever you get a sec?  I'm not sure why the leading slash was causing such a problem, and it looks like there are other unit tests that are failing the build.

Issue #505, opened 2019-01-03
Identification string should be sent before waiting for server identification string


martinprikryl:
According to [RFC 4253 Section 4.2. Protocol Version Exchange](https://tools.ietf.org/html/rfc4253#section-4.2):

> When the connection has been established, both sides **MUST send an identification string.**

SSH.NET sends its identification string only after waiting for server's identification string. Amazon Managed SFTP server has the same issue. So you cannot connect to that SFTP server with SSH.NET. Both sides first wait for the other side to send the identification string before sending its own.

Spotted thanks to: https://stackoverflow.com/q/54014749/850848

BrainBacon:
Wow, this is exactly the same problem we are having. Going to report a bug to Amazon myself too.

psaffrey-origami:
We are also experiencing this issue. @BrainBacon - did you file a bug with Amazon? Can you link it?

denispakizh:
The same issue is with KEY EXCHANGE INIT message, library waits for the server's KEY EXCHANGE INIT message while it should not:
>Key exchange (kex) begins by each side sending name-lists of supported algorithms.

https://tools.ietf.org/html/rfc4253#section-7

drieseng:
This is fixed in [2020.0.0-beta1](https://github.com/sshnet/SSH.NET/releases/tag/2020.0.0-beta1).

PR #506, opened 2019-01-03
Sending identification string before waiting for server's

martinprikryl:
According to RFC 4253 Section 4.2. Protocol Version Exchange:

> When the connection has been established, both sides MUST send an identification string.

SSH.NET sends its identification string only after waiting for server's identification string. Amazon Managed SFTP server has the same issue. So you cannot connect to that SFTP server with SSH.NET. Both sides first wait for the other side to send the identification string before sending its own.

https://github.com/sshnet/SSH.NET/issues/505

### DIFF
--- a/src/Renci.SshNet/Session.cs
+++ b/src/Renci.SshNet/Session.cs
@@ -601,6 +601,9 @@ namespace Renci.SshNet
                             break;
                     }
 
+                    // Send our identification string before waiting for server's, as mandated by RFC 4253.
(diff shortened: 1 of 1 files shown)


Strohbichler:
Hi Martin,
we implement this bugfix in our solution and it works. 
THX :-)
Best Regards,
Strohbichler

orudge:
I came up with the same fix while debugging an issue while trying to connect to an SSH server via an HTTP proxy (direct connection and SOCKS5 connections worked, HTTP did not). If this could be merged it would be very helpful!

drieseng:
This is fixed in [2020.0.0-beta1](https://github.com/sshnet/SSH.NET/releases/tag/2020.0.0-beta1).
Thanks!

PR #508, opened 2019-01-11
Even with SeekOrigin.End offset should be added

martinprikryl:
... not subtracted from the end position


bronumski:
We are facing this bug as well. Would love to see this addressed. Until then we will try using the work around as suggested here:
https://stackoverflow.com/questions/63991495/list-files-inside-zip-file-located-on-sftp-server-in-c-sharp

WojciechNagorski:
Old outdated PR.

PR #536, opened 2019-04-21
[Do not merge] Pageant patch port to current HEAD

ghost:
```
'Session' does not contain a definition for 'MessageReceived' and no accessible extension method 'MessageReceived' accepting a first argument of type 'Session' could be found
```

Mainly opening this PR for anyone else who'd like to work on this.

[patches-ssh-net.zip](https://github.com/sshnet/SSH.NET/files/3100848/patches-ssh-net.zip) - original patches. Had to manually patch the csproj.


ghost:
https://github.com/sshnet/ssh.net/commit/f9773a67861c89c7024870c5638118204e7a99af#diff-2f129846bc0bc158cc4105e5ee3dbaddR144 gave me enough of a hint to fix the build for .NET Framework, but someone else will need to test.

https://docs.microsoft.com/en-us/dotnet/api/system.io.memorymappedfiles.memorymappedfile.getaccesscontrol does not exist for Core.

ghost:
Next, I've put (part) of the code behind a constant that's only enabled for netstandard2.0. Adds a dependency on `System.IO.MemoryMappedFiles` but I don't know if you can toggle dependencies with a constant like that. Getting out of my depth here.

ghost:
I think I'm hitting https://www.chiark.greenend.org.uk/~sgtatham/putty/wishlist/pageant-backwards-compatibility.html without the ability to use SetAccessControl / SetOwner in Core. Someone much smarter than me will need to take a look at this.

PR #564, opened 2019-07-12
Added SCPClient.Download overload that returns a stream.

tim-mills:
SCP is significantly quicker when dealing with a straight forward read as opposed to Sftp. The Download method on the SCPClient didn't return a stream.

For this scenario, I'm reading a large file, I then need to pass the stream to downward processes to handle, decryption, filtering and write to our end data store. For this to be possible I need to be obtain the read stream first and then I can pass through to each onward process.

This allows for a 17Gb file to be read, decrypted, filtered and saved to a data store all without the file touching the disk at any time.

To ensure notifications are still issued and the channel is cleaned up I've created an ScpDownloadStream, which handles reading and giving notifications back to the client. 

In my testing I found the SCPClient to be twice the speed of the SftpClient when comparing my new method with the SFTPClient.OpenRead.

I'm happy to take on any feedback to get this in. Whilst I can work off a fork I'd prefer to have this in the codebase as I think beneficial and it would align the SCPClient and SFTPClient, both being able to obtain a stream rather than provide one.

I've added an integration test to ensure everything works correctly.

### DIFF
--- a/src/Renci.SshNet.NET35/Renci.SshNet.NET35.csproj
+++ b/src/Renci.SshNet.NET35/Renci.SshNet.NET35.csproj
@@ -206,6 +206,9 @@
     <Compile Include="..\Renci.SshNet\Common\ScpDownloadEventArgs.cs">
       <Link>Common\ScpDownloadEventArgs.cs</Link>
     </Compile>
+    <Compile Include="..\Renci.SshNet\Common\ScpDownloadStream.cs">
+      <Link>Common\ScpDownloadStream.cs</Link>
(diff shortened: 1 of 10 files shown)


tim-mills:
@drieseng hopefully ready to review - anything you're not happy with let me know and hopefully I can rectify to get it in.

WojciechNagorski:
@tim-mills PR needs refreshing.

asos-timmills:
I did this 4 years ago... feel free to just close.

PR #573, opened 2019-08-04
Remove MD5 and SHA1 support

LoadingPleaseWait:
MD5 and SHA1 are no longer considered secure and therefore should not be used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,17 +30,10 @@ This project was inspired by **Sharp.SSH** library which was ported from java an
 
 **SSH.NET** supports the following key exchange methods:
 * diffie-hellman-group-exchange-sha256
-* diffie-hellman-group-exchange-sha1
-* diffie-hellman-group14-sha1
-* diffie-hellman-group1-sha1
 
 ## Message Authentication Code
 
 **SSH.NET** supports the following MAC algorithms:
-* hmac-md5
-* hmac-md5-96
-* hmac-sha1
-* hmac-sha1-96
 * hmac-sha2-256
 * hmac-sha2-256-96
 * hmac-sha2-512
--- a/src/Renci.SshNet/ConnectionInfo.cs
+++ b/src/Renci.SshNet/ConnectionInfo.cs
@@ -323,9 +323,9 @@ namespace Renci.SshNet
             KeyExchangeAlgorithms = new Dictionary<string, Type>
                 {
                     {"diffie-hellman-group-exchange-sha256", typeof (KeyExchangeDiffieHellmanGroupExchangeSha256)},
-                    {"diffie-hellman-group-exchange-sha1", typeof (KeyExchangeDiffieHellmanGroupExchangeSha1)},
-                    {"diffie-hellman-group14-sha1", typeof (KeyExchangeDiffieHellmanGroup14Sha1)},
-                    {"diffie-hellman-group1-sha1", typeof (KeyExchangeDiffieHellmanGroup1Sha1)},
+                    //{"diffie-hellman-group-exchange-sha1", typeof (KeyExchangeDiffieHellmanGroupExchangeSha1)},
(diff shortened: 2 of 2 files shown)


drieseng:
I agree these are not/less secure, but isn't it enough if SSH servers quit supporting these (in environments where security is considered important)?

darinkes:
Please check https://man.openbsd.org/sshd_config#KexAlgorithms and https://man.openbsd.org/sshd_config#MACs

Defaults include SHA1, so I disagree on removing it.
MD5 I agree with @drieseng.

darinkes:
Maybe adding a way to set the KEX, MAC Algorithms and other Options in ConnectionInfo might be a better approach.

Like the -o KexAlgorithms="" and -o MACs="" ssh-client Options.

LoadingPleaseWait:
The client should not support weak algorithms either to prevent downgrade attacks. Under no circumstance would I want MD5 or SHA1 to be used for one of my connections.

darinkes:
_diffie-hellman-group14-sha1_ and _hmac-sha1_ need to stay, since both are still used by OpenSSH too.
Also the list should be reorded to reflect the default order of OpenSSH IMHO.
Reference: https://github.com/openbsd/src/blob/master/usr.bin/ssh/myproposal.h

LoadingPleaseWait:
I disagree. OpenSSH supports SHA2 and we should try to use SHA2 instead of MD5 or SHA1.

darinkes:
> I disagree. OpenSSH supports SHA2 and we should try to use SHA2 instead of MD5 or SHA1.

I don't see a conflict here. Order is of course sha2 then sha1. _diffie-hellman-group14-sha1_ and _hmac-sha1_ are the last in their lists. And MD5 should be removed.
As stated in https://github.com/openbsd/src/blob/master/usr.bin/ssh/myproposal.h

LoadingPleaseWait:
SHA1 should be removed entirely. SHA1's presence on the list makes a downgrade attack possible.

darinkes:
Can you provide papers/links so I can understand your argument.
OpenSSH still ships with this config and I can't see any reason not to follow them.

darinkes:
Talked with an OpenSSH Developer. Reason its still in the default list is compat with implementations which can't do other KEX or MAC. SSH.NET should do the same I think.

LoadingPleaseWait:
https://en.wikipedia.org/wiki/Downgrade_attack

darinkes:
I understand, you dont want to discuss. I would never break stuff for all users just because someone posts a Wikipedia-Link 😉
@drieseng your turn

LoadingPleaseWait:
That's not the case. You asked for a link that would help explain my argument. That link helps explain my argument. We need to remove support for weak algorithms in order to minimize the risk of a man-in-the-middle attack. Also, this will not "break stuff for all users."

darinkes:
Sorry, looked like a lmgtfy for me.
There is no such things for SSH. No MITM (you have to verify the hostkey for sure) and no Downgrade Attack.
SSH-Protocol prevents this by letting server and client exchange their list included in a hash.
See: https://github.com/sshnet/SSH.NET/blob/master/src/Renci.SshNet/Session.cs#L311
and: https://github.com/sshnet/SSH.NET/blob/master/src/Renci.SshNet/Security/KeyExchangeDiffieHellmanGroupExchangeShaBase.cs#L20

Of course this change will break stuff for all users. There is no opt-out, it justs removes them.
Thats why I propose to make the list changeable via ConnectionInfo and as default keep the list in sync with OpenSSH.

drieseng:
I don't think we should be removing anything at all.
What I would like is to have a fluent (but still extensible) API that allows one to explicitly specify the algorithms to support (and the order in which they're specified would control the order).
Such an approach offers flexibility for power users, but risks to be obstacle for less experienced users. Or am I mistaken?

LoadingPleaseWait:
Alright, that approach makes sense.

Issue #576, opened 2019-08-10
A proposal to add TPM support


ghost:
Hello,

I am not sure if this is the correct place to ask, so please point out if I am doing this wrong.

I have been working on a small PoC codebase forked from SSH.NET that supports private keys stored in TPM instead of the filesystem.
The additional code talks to the TPM via a newer subset of CNG API, so it works only on some later versions of Windows 10, in which the API was first introduced.

The quality of the code is really low right now, and I don't think it could be suddenly made into a PR.
Before I start working on this, I wanted to ask if anyone is interested in implementing TPM support.
I don't want to waste time on a feature that nobody asked for...

Thank you in advance.

keithallenjackson:
I would love that feature! I've been looking at smart card support for my work.

ghost:
@keithallenjackson Thank you for your comment! I'll start working. Unfortunately I cannot test SmartCard support because I do not have the device, but it should definitely be supported by just changing a few lines, designating SmartCard instead of TPM as the Key Storage Provider.

ghost:
@keithallenjackson 
I set up a repo [here](https://github.com/sadlame/Renci.SshNet.Cng) which contains all the extra code required to support TPM/Smartcard authentication as a separate library.

It includes a sample program which demonstrates how to use the additional functionalities, e.g. creating a persistent key in TPM/Smartcard, deleting them, exporting them to OpenSSH format, and most importantly, logging in to a remote host with the key via SSH.

The quality of the code still needs improvement and the library lacks some important functionalities like importing keys from external files, but I believe it is already self-contained as a proof of concept.

The change required in SSH.NET itself (#577) is not available from NuGet of course, so the repo includes the modified version of SSH.NET as a submodule.
So before you build, you have to initialize the submodule so its content is recognized by Visual Studio.

The sample program is quite self-explanatory, and you would be able to discover how to deal with Smartcard as well as TPM I hope. Please note that it might require admin privilege to read from/write to these devices.

Please feel free to ask me any questions. Thank you.

darinkes:
How about adding PKCS#11 to work with smartcards, too? It works with .NET Core on Linux and also on older Windows Versions.

Tested and used it with SSH.NET in a project which source I currently cannot publish.

ghost:
[...]

PR #577, opened 2019-08-11
Make PrivateKeyFile an implementation of IPrivateKeySource to make room for other key sources

ghost:
Hello,

I am planning to add support for TPM/SmartCard-based private key authentication to SSH.NET (please see #576).
TPM/SmartCard-based private key authentication is a new authentication method in which the private key is stored in a special hardware (outside of the filesystem), reinforcing client-side security by preventing anyone (even intruders w/o certain privileges) from stealing the private key.

This PR does not involve the functionality itself, but it includes the only breaking change to the API, which is why I opened this PR.
Specifically, I want to create a new interface IPrivateKeySource which is implemented by PrivateKeyFile so that in the future, PrivateKeyCng (which represents a private key stored somewhere else) could be handled transparently.
The rest of the changes required to implement TPM/SmartCard authentication is not going to impact the public API at all, just adding new classes and new files.

If this change is deemed unacceptable, I will try to come up with another way to pass the new kind of private keys to the API.

Thank you in advance.

### DIFF
--- a/src/Renci.SshNet.NET35/Renci.SshNet.NET35.csproj
+++ b/src/Renci.SshNet.NET35/Renci.SshNet.NET35.csproj
@@ -308,6 +308,9 @@
     <Compile Include="..\Renci.SshNet\IForwardedPort.cs">
       <Link>IForwardedPort.cs</Link>
     </Compile>
+    <Compile Include="..\Renci.SshNet\IPrivateKeySource.cs">
+      <Link>IPrivateKeySource.cs</Link>
+    </Compile>
     <Compile Include="..\Renci.SshNet\IRemotePathTransformation.cs">
       <Link>IRemotePathTransformation.cs</Link>
     </Compile>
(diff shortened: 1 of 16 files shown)


darkoperator:
Since the library is portable via .Net Standard and Core and used in mobile platforms and other platforms like Mac OS and Linux would this provide problems in those?

ghost:
@darkoperator I'm sorry I forgot to clarify that. This feature is going to be exclusively available to .NET 3.5 due to the use of P/Invoke. I must think of a way to perform tests without problems, but apart from the changes in this PR, users on the other platforms will not notice any changes in the API.

ghost:
If anyone is interested, I posted a link to a repo in #576 which contains a demo in which a private key from TPM/Smartcard is used to log in to a remote host.

drieseng:
Please provide a small text fragment that describes this change, so that we can include it in the release notes.

Thx!

ghost:
@drieseng I'm new and I am not sure if I'm doing this right, does this suffice? Thank you.

> Added a public interface IPrivateKeySource to enable authentication by plugins.

poizan42:
What about alternatively just replace all usages of PrivateKeyFile with HostAlgorithm? The only accessible field in PrivateKeyFile is the HostAlgorithm anyways, and HostAlgorithm is already extensible.

IgorMilavec:
@sadlame #794 has been merged, can you please check if this PR can be closed.

WojciechNagorski:
@sadlame PR needs refreshing.

PR #588, opened 2019-10-02
Fixed Deadlock when disposing in .NET Core 3.0

iodes:
When trying to Dispose SSHClient in .NET Core 3.0 (macOS Mojave), It will not end forever.
I solved the problem with a few modifications:

- Added support for .NET Standard 2.1
- Use SocketShutdown.Both for Session Shutdown
- Use Socket.Poll instead of Socket.Select (otherwise randomly deadlock)

Thanks!


drieseng:
@iodes Do you happen to have a small self-contained repro for this issue?

drieseng:
This should already be fixed in [2020.0.0-beta1](https://github.com/sshnet/SSH.NET/releases/tag/2020.0.0-beta1).

PR #590, opened 2019-10-02
Support on sftp local changedir mimmicking

lirunki:
There are some implementations of sftp servers where "ChangeDir" will cause the server to go to a different physical path (due to soft links on the server) where the permissions are different from those permissions if the client-perceived paths are used.
The current sftp client implementation makes use of GetCanonicalPath which causes for those servers to potentially cause a permissions error where the path provided by the client has the right permissions.

This change allows the sftpclient to be configured so the client will always use the paths provided by the application and will not try to changedirs on the server, thus avoiding this issue.

For reference, this problem is present in some Sftp implementation for certain Cisco routers, for example.


WojciechNagorski:
@lirunki PR needs refreshing.

Issue #479, opened 2018-10-16
Support for ssh-rsa-cert-v01 certificates


robertkentish:
I'd like to implement support for ssh-rsa-cert-v01@openssh.com (and others) type certificate authentication and am looking for some guidance on the best way to implements this.  

Looking at the code I propose to create a new PrivateKeyCertAuthenticationMethod class and associated CertificateHostAlgorithm to be used instead of the current PrivateKey/KeyHostAlgorithm classes.  This would make the high level connection code something like

```csharp
var keyFile = new PrivateKeyFile(@"C:\temp\ssh_keys\id_rsa", "xxxx");
var certFile = new PublicKeyCertFile(@"C:\temp\ssh_keys\id_rsa-cert.pub");
var authMethod = new PrivateKeyCertAuthenticationMethod("user", keyFile, certFile);
var connectionInfo = new ConnectionInfo("192.168.1.1", "user", authMethod);

using (var client = new SshClient(connectionInfo))
{
    client.Connect();
}

```

What I'm wondering though, is this the best way to structure things or should I create only the CertificateHostAlgorithm and place some switching logic in the PrivateKeyFile class to swap out the HostAlgorithm? 

Just looking for a bit of direction before I put through a huge PR that gets rejected... :-)

nokinger:
Hi there, 

has anybody some stuff regarding the certificate authentication? 
Or @robertkentish  did you implement something?

robertkentish:
Hi @nokinger ,
I did end up getting it working but was waiting on guidance on the PR and so haven't pushed up any of my mods yet. I'll clean up the local branch tomorrow and push up the changes to my fork [here](https://github.com/robertkentish/SSH.NET) under the allow_signed_keys branch

nokinger:
thanks @robertkentish that will be nice, I found also a [fork](https://github.com/msvprogs/SSH.NET) what contains a pretty simple implementation. But yesterday there was to many interruptions, so today i will have a try for that fork.

robertkentish:
@nokinger I've just pushed up my changes so if you still need, feel free to try them.

nokinger:
i didn't have any knowledge about the ssh-rsa-cert-v01@openssh.com  in deep and so on. But when i use your code as it is, it will fail. 
I just changed your code base like this: 

`                var message = new RequestMessagePublicKey(ServiceName.Connection,
                                                            Username,
                                                            CertificateFile.HostCertificate.Name,
[...]

PR #595, opened 2019-10-17
Allow signed keys

robertkentish:
PR as per issue #479. PRoposal to support public certs for RSA signatures as in the following manner.

var keyFile = new PrivateKeyFile(@"C:\temp\ssh_keys\id_rsa", "xxxx");
var certFile = new PublicKeyCertFile(@"C:\temp\ssh_keys\id_rsa-cert.pub");
var authMethod = new PrivateKeyCertAuthenticationMethod("user", keyFile, certFile);
var connectionInfo = new ConnectionInfo("192.168.1.1", "user", authMethod);

using (var client = new SshClient(connectionInfo))
{
    client.Connect();
}

### DIFF
--- a/src/Renci.SshNet/Security/CertificateHostAlgorithm.cs
+++ b/src/Renci.SshNet/Security/CertificateHostAlgorithm.cs
@@ -1,4 +1,7 @@
-﻿using System;
+﻿using Renci.SshNet.Common;
+using System;
+using System.Collections.Generic;
+using System.Text;
 
 namespace Renci.SshNet.Security
 {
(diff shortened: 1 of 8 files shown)


src/Renci.SshNet/Security/CertificateHostAlgorithm.cs
```diff
@@ -7,12 +10,23 @@ namespace Renci.SshNet.Security
     /// </summary>
     public class CertificateHostAlgorithm : HostAlgorithm
```
drieseng, line 11: Does this class have to be public? We do not expose an instance publicly right now.
If not, **RsaCertificate** should also not be public.

src/Renci.SshNet/Security/CertificateHostAlgorithm.cs
```diff
@@ -46,5 +84,280 @@ public override bool VerifySignature(byte[] data, byte[] signature)
...
+            /// Gets or sets the name of the algorithm as UTF-8 encoded byte array.
+            /// </summary>
+            /// <value>
+            /// The name of the algorithm.
+            /// </value>
+            private byte[] _name { get; set; }
+            public string Name
+            {
+                get { return Utf8.GetString(_name, 0, _name.Length); }
+                set { _name = Utf8.GetBytes(value); }
```
drieseng, line 101: Do any of these properties actually need a public setter?

src/Renci.SshNet/Security/Cryptography/RsaCertificate.cs
```diff
@@ -0,0 +1,157 @@
+﻿using System;
+using System.Collections.Generic;
+using Renci.SshNet.Common;
+
+namespace Renci.SshNet.Security
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public class RsaCertificate : IDisposable
```
drieseng, line 10: If there's no use in **CertificateHostAlgorithm** being public, we should also make this class internal.

drieseng:
@robertkentish I did an initial sweep of your changes. Still need to dive a little deeper though.
I wonder if - when complete - we should consider shipping this as a separate package (under the **SSH.NET** unbrella, or not).

WojciechNagorski:
@robertkentish PR needs refreshing.

PR #615, opened 2020-01-14
WIP: Remote Dynamic Forwarding

darinkes:
This is still WIP and in RFC state.

It implements reverse dynamic forwarding added in OpenSSH 7.5.

https://www.openssh.com/releasenotes.html

>  * ssh(1): add support for reverse dynamic forwarding. In this mode,
>    ssh will act as a SOCKS4/5 proxy and forward connections
>    to destinations requested by the remote SOCKS client. This mode
>    is requested using extended syntax for the -R and RemoteForward
>    options and, because it is implemented solely at the client,
>    does not require the server be updated to be supported.

Usage is the same like in the ssh(1) commandline tool (https://man.openbsd.org/ssh.1)

```
using (var client = new SshClient("localhost", 22, "username", "password"))
{
    client.Connect();
    var forwarding = new ForwardedPortRemote("127.0.0.1", 5555);
    client.AddForwardedPort(forwarding);

    forwarding.Start();
}
```

### DIFF
--- a/src/Renci.SshNet/Channels/ChannelForwardedTcpip.cs
+++ b/src/Renci.SshNet/Channels/ChannelForwardedTcpip.cs
@@ -1,6 +1,9 @@
 ﻿using System;
+using System.IO;
+using System.Linq;
 using System.Net;
 using System.Net.Sockets;
+using System.Threading;
 using Renci.SshNet.Abstractions;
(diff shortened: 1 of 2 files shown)


WojciechNagorski:
@darinkes PR needs refreshing.

PR #634, opened 2020-02-29
Remove all net35 build targets

chrgraefe:
These small commit removes all "net35" build targets


drieseng:
@chrgraefe Why would you want to do this?

chrgraefe:
Sorry, I was wrong. I will close it

PR #639, opened 2020-03-10
Updated NETCONF framing protocol detection to check both client & server capabilities

### LINKED ISSUE #342: NETCONF Incorrect Client Hello Response
Hello,
In src\Renci.SshNet\Netconf\NetConfSession.cs, lines 41 - 42 is missing the client request for the 1.1 protocol.  The code should read:

```
            ClientCapabilities.LoadXml("<?xml version=\"1.0\" encoding=\"UTF-8\"?>" +
                                                "<hello xmlns=\"urn:ietf:params:xml:ns:netconf:base:1.0\">" +
                                                    "<capabilities>" +
                                                        "<capability>" +
                                                            "urn:ietf:params:netconf:base:1.0" +
                                                        "</capability>" +
                                                        "<capability>" +
                                                            "urn:ietf:params:netconf:base:1.1" +
                                                        "</capability>" +
                                                    "</capabilities>" +
                                                "</hello>");
```
Without adding the second capability, line 134 is invalid:
```
            _usingFramingProtocol = (ServerCapabilities.SelectSingleNode("/nc:hello/nc:capabilities/nc:capability[text()='urn:ietf:params:netconf:base:1.1']", nsMgr) != null);
```

The reason is that the server will not be using framing if the client does not state that it is supported.  Line 138 uses this flag to handle chunking with the framing protocol, though the server correctly responds without framing.  

Thanks,

-- Joe

### LINKED ISSUE #405: netconf capability wrong
The client advise the base 1.0, however the flag to decide whether data framing mechanism to use is depended on the server response. When base 1.1 is advise by the server, the netconf gets wrong, since the client only advises 1.0 but the flag sets to use data framing mechanism.

Suggest that: in  "NetConfSession.cs" add in line 44

`"<capability>" +
                                                            "urn:ietf:params:netconf:base:1.1" +
                                                        "</capability>" +`

### LINKED ISSUE #1197: what is difference ssh.net vs ssh.net-netconf
1. My program is doing netconf communication.  (me is netconf client)
1-1. When I use 'ssh.net-netconf', all of them(connect/sendreceieve ) success.
1-2. When I use 'ssh.net', only connect success. 
what difference?

2. I need host key algorithm "ecdsa-sha2-nistp256". But netconf package has just rsa, dss.
How can I add "ecdsa-sha2-nistp256" algorithm without ssh.net package?

declspec:
This fixes an issue where the NetConfSession would expect the framing protocol to be used if ServerCapabilities contained 1.1, however the server would actually be using the legacy protocol as the client only advertises support for 1.0.

While SSH.NET claims to support the framing protocol, the implementation appears to be buggy / broken; advertising 1.1 support on the client-side is not advisable until that is investigated / fixed.

Fixes #405 
Fixes #342 
Fixes #1197


schaveyt:
Please merge this

Rob-Hague (approved):

PR #661, opened 2020-04-22
Fix #153 support of Task-based Asynchronous Pattern (TAP) over existing async API for SftpClient. (Not supported by .net 3.5)

### LINKED ISSUE #153: Add support for a Task based async API

bartizan:


bartizan:
@mmoo9154 Some tests were down for a while, it seems it is not related to the PR.

WojciechNagorski:
Old PR adding already existing functionality.

PR #675, opened 2020-05-06
Send keys proactively

MK0151:
Added flag to send keys proactively. Some devices wait for the keys (like SSH.Net does) so it will timeout because both parties are waiting for a message from the other.


WojciechNagorski:

WojciechNagorski:

PR #680, opened 2020-05-21
🐛 AsyncResult contains invalid value (Fix #329)

### LINKED ISSUE #329: UploadedBytes wrong value
Hello,
I have a small problem and do not know if this is my mistake or a bug.
I use the SftpClient to upload a file. For this I use BeginUploadFile with Filestream.
In the end I check with uploadhandler.UploadedBytes if the file was uploaded correctly.
But there always comes a different result out and only occasionally the right number.
Here is the simplified code:
```
SftpClient client = new SftpClient(host, port, user, password);
client.Connect();
FileStream fileStream =  new FileStream(source, FileMode.Open);
SftpUploadAsyncResult uploadhandler = client.BeginUploadFile(fileStream, target, null, null) as SftpUploadAsyncResult;

                        while (!uploadhandler.IsCompleted)
                        {
                                Thread.Sleep(500);

                        }
                        client.EndUploadFile(uploadhandler);
                        FileInfo fi = new FileInfo(source);
                        if (uploadhandler.UploadedBytes == Convert.ToUInt64(fi.Length))
                        {
                            return true;
                        } else {
                            return false
                        }

miroslavpokorny:
- AsyncResult should contain invalid value immediately after the async operation is marked as completed
- there were race condition problems with a callback method which is invoked on a different thread so updating of value is done without any synchronization. So in some cases, the async operation is marked as completed but async result value is not yet updated and contains an invalid value

#329

### DIFF
--- a/src/Renci.SshNet/SftpClient.cs
+++ b/src/Renci.SshNet/SftpClient.cs
@@ -473,7 +473,7 @@ namespace Renci.SshNet
         {
             CheckDisposed();
 
-            return InternalListDirectory(path, listCallback);
+            return InternalListDirectory(path, null, listCallback);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


roewer:
Looks like this fixed has not been merged into develop yet.  The problem still exists in the current 2020.0.1 release.  Can this fix get released?

WojciechNagorski:
@miroslavpokorny PR needs refreshing.

Rob-Hague:
Thanks @miroslavpokorny, this was rebased and merged in #1184 (you are a co-author on that PR).

WojciechNagorski:

PR #685, opened 2020-05-26
PrivateKeyFile constructor xml documentation update

savstars:
While I was making use of the PrivateKeyFile constructors, I noticed the xml documentation state that a NullReferenceExeception would be thrown if the passPhrase was null.  This made my code just a little bit more complex.
Upon investigation, I found that documentation did not reflect the implementation.  This pull request is to update that piece of documentation.


daviburg (approved):

WojciechNagorski:
PR needs refreshing. If you feel this change is needed feel free to open the current PR.

PR #698, opened 2020-06-23
Preserve exception stack trace in Session.WaitOnHandle()

bretthysuik:
Throwing `_exception` loses the original stack trace, making it difficult to debug problems.
Original stack trace is now saved in the exception data for .NET 3.5 and 4.0 clients, or
correctly rethrown with the full stack for .NET Standard 1.3 and 2.0 clients.

### DIFF
--- a/src/Renci.SshNet/Session.cs
+++ b/src/Renci.SshNet/Session.cs
@@ -2,6 +2,9 @@
 using System.Collections.Generic;
 using System.Net;
 using System.Net.Sockets;
+#if NETSTANDARD
+using System.Runtime.ExceptionServices;
+#endif
 using System.Security.Cryptography;
 using System.Text;
 using System.Text.RegularExpressions;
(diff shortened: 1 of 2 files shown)


src/Renci.SshNet/Session.cs
```diff
@@ -905,7 +905,13 @@ internal void WaitOnHandle(WaitHandle waitHandle, TimeSpan timeout)
             switch (WaitHandle.WaitAny(waitHandles, timeout))
             {
                 case 0:
+#if NETSTANDARD1_3 || NETSTANDARD2_0
+                    System.Runtime.ExceptionServices.ExceptionDispatchInfo.Capture(_exception).Throw();
```
daviburg: Capture and throw shouldn't be on the same line. Capture should happen when the exception is created.

daviburg (changes requested):

src/Renci.SshNet/Session.cs
bretthysuik: You're right, my implementation is totally out to lunch. I'll get it fixed up shortly.

daviburg: I haven't heard that one before. You just made my day.

bretthysuik: Is it OK if I update the language version of the project to `<LangVersion>latest</LangVersion>`? It's currently at C#5 but I'd like to use null propagating operators in C#6 to simplify some of my changes.

daviburg: Oh boy, I'd love to see the code modernized to never practices including null propagation and string interpolation. I am not a repo owner however, and don't know where they stand on that.

bretthysuik: Fixed

daviburg (approved):

WojciechNagorski:
PR needs refreshing. If you feel this change is needed, feel free to open the current PR.

PR #705, opened 2020-07-14
Adding Pagant Support

kins-dev:
Adds pagent support.

That said, the dependencies **requires** .Net 4.0.  Pagent support in non-windows desktop environment does not make sense, so this requirement seems reasonable.

### DIFF
--- a/src/Renci.SshNet/PageantProtocol.cs
+++ b/src/Renci.SshNet/PageantProtocol.cs
@@ -1,0 +1,200 @@
+﻿#if NETFRAMEWORK && !NET20 && !NET35
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.IO.MemoryMappedFiles;
+using System.Net;
+using System.Runtime.InteropServices;
+using System.Text;
+using Renci.SshNet.Common;
+
+namespace Renci.SshNet.Pageant {
+    /// <summary>
+    /// 
+    /// </summary>
+    public class PageantProtocol : IAgentProtocol {
+
+        #region  Constants
+
+        private const int WM_COPYDATA = 0x004A;
+
+        private const long AGENT_COPYDATA_ID = 0x804e50ba;
+        private const int AGENT_COPYDATA_ID_INT = unchecked ((int) AGENT_COPYDATA_ID);
+
+        private const int AGENT_MAX_MSGLEN = 8192;
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public const byte SSH2_AGENTC_REQUEST_IDENTITIES = 11;
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public const byte SSH2_AGENT_IDENTITIES_ANSWER = 12;
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public const byte SSH2_AGENTC_SIGN_REQUEST = 13;
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public const byte SSH2_AGENT_SIGN_RESPONSE = 14;
+
+        #endregion
+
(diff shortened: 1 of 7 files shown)


darkoperator:
Would this support not make more sense on a Appa built with the library than on the library that provides protocol core support?

kins-dev:
Honestly I'm not sure what an Appa is in this context.

I looked at it this way,  The library handles the entire key exchange back and forth.  Since a pagent isn't getting a key but instead using side channel as a challenge response, it would require changes on the core to widen interfaces to allow that new mechanism.  This would then lead to a shim everyone who might want pagent support would need to use (add an extra layer of indirection and a possible security vector) or each application would need to do it themselves.

Rather than sending the people down the road of re-inventing the wheel, (especially since the documentation on how to do this, would basically be the same code I wrote, only in outside programs) I figured it was better to do it once, in the core.

Maybe I'm missing something here, so I'm open to other suggestions.  It would also be trivial to make this a compile time option instead, if there's a concern about performance/security.

darkoperator:
Sorry was typing a n my phone, I meant App. A better one would be OpenSSH ssh-agent since it would mean compatibility with Windows, Mac and Linux. Specially since OpenSSH is now part of Windows. Only providing my opinion as a user. Im not a project Dev :)

kins-dev:
Ah I see where you're coming from.  

Implementing the same thing in every app probably is a bad idea.  You get a bunch of copies of the example code, with stylistic or other arbitrary changes, a shim layer that gets created later that a number of people will rely on, and then a fork of the original code base.  This creates a mess and makes it very hard to push out security updates.

As for the platforms:
 - Linux: this is pretty straightforward with example code of how to read and write to a socket file.
 - Mac: this is a socket file as well, but I believe Apple's key chain is involved.  May require something special, but should be pretty easy
 - Windows:
   - Pagent: this is the putty SSH agent.  It is far and away the oldest and most widely used SSH agent.  Uses Window handles  and messages to pass challenge/response data so it works with no configuration from the user.
   - Cygwin: Uses a socket file like Linux
   - Msys: Also uses a socket file, but the format is different than cygwin
   - Native OpenSSH Windows: Uses a Windows service, the documentation is not clear if there is a socket file or if this is using a different IPC mechanism. 

Pagent solved my immediate problem and I was able to find good examples to work from.  That said I would expect in the future more ssh agent protocols to be implemented based on need.  They might even be split out to a plugin mechanism.

src/Renci.SshNet/PageantProtocol.cs
```diff
@@ -0,0 +1,200 @@
...
+
+            string mmFileName = Path.GetRandomFileName ();
+
+            using (var mmFile = MemoryMappedFile.CreateNew (mmFileName, AGENT_MAX_MSGLEN)) {
+                var security = mmFile.GetAccessControl ();
+                security.SetOwner (System.Security.Principal.WindowsIdentity.GetCurrent ().User);
+                mmFile.SetAccessControl (security);
+                using (var accessor = mmFile.CreateViewAccessor ()) {
+
+                    accessor.Write (0, IPAddress.NetworkToHostOrder (AGENT_MAX_MSGLEN - 4));
```
daviburg, line 88: Here would a be good place to have code comments to explain what we are doing, point to documentation and/or specification that apply to this implementation.

daviburg (changes requested):

WojciechNagorski:
I'm closing PR. This is an old issue with no votes.
@kins-dev If you still feel a change is needed, you can refresh this PR and reopen it. SSH.NET support right now .NET Standard 2.0 and .NET 7 that are cross-platform. This feature should only be usable on the relevant systems/target-framework.

PR #727, opened 2020-09-10
Fix exception when reading duplicate sftp extensions

### LINKED ISSUE #726: Exception when reading sftp extensions
Hello,

I encounter the following exception reading the SFTP extensions => An item with the same key has already been added

The server is returning a duplicate extension : posix-rename@openssh.com which make failed the dictionary Add  in ReadExtensionPair().

I know the server shouldn't return twice the extension but I think it would be good to not fail and maybe to keep the last key/value pair received.

I can submit a fix. What do you think ?

Regards

---------- EXCEPTION -------------
System.ArgumentException: An item with the same key has already been added.
   at Renci.SshNet.SubsystemSession.WaitOnHandle(WaitHandle waitHandle, Int32 millisecondsTimeout)
   at Renci.SshNet.Sftp.SftpSession.OnChannelOpen()
   at Renci.SshNet.SubsystemSession.Connect()
   at Renci.SshNet.SftpClient.OnConnected()
   at Renci.SshNet.BaseClient.Connect()

alb3ric:
If the server return duplicated sftp extensions then we keep the last value instead of throwing an exception

Fix #726

### DIFF
--- a/src/Renci.SshNet/Common/SshData.cs
+++ b/src/Renci.SshNet/Common/SshData.cs
@@ -251,7 +251,14 @@ namespace Renci.SshNet.Common
             {
                 var extensionName = ReadString(Ascii);
                 var extensionData = ReadString(Ascii);
-                result.Add(extensionName, extensionData);
+                if (result.ContainsKey(extensionName)) //if sftp extension is present multiple-times then keep the last one
+                {
+                    result[extensionName] = extensionData;
+                }
+                else
+                {
+                    result.Add(extensionName, extensionData);
+                }
             }
             return result;
         }


src/Renci.SshNet/Common/SshData.cs
```diff
@@ -251,7 +251,14 @@ protected string[] ReadNamesList()
             {
                 var extensionName = ReadString(Ascii);
                 var extensionData = ReadString(Ascii);
-                result.Add(extensionName, extensionData);
+                if (result.ContainsKey(extensionName)) //if sftp extension is present multiple-times then keep the last one
```
daviburg, line 254: It's difficult to judge what the right behavior is for duplicate extension value.
Do you have a specification reference for SFTP extension to support a decision one way or another?
Can there be an approach similar to HTTP's multiple value for same header, which combines the values to a string with a known separator?
Or update the dictionary to be <string, IEnumerable<string>>

Issue #45, opened 2016-07-21
Tracing is only available in Debug


JustinMellor:
The "SshNet.Logging" Trace Source is only written to in in debug builds. There is a [Conditional("DEBUG")] on the DiagnosticAbstraction.Log() method.

I know that the trace is limited currently but would assume you should be able to turn it on anyway in the distributed NuGet packages.

It worked in the last stable release (2013.4.7).

lchengNetBall:
Please advise how to add trace log?

daviburg:
Oh my, ouch. Tracing is essential for consumers of the library. Furthermore the logging should be updated to modern EventSource to emit standard ETW events.

Rob-Hague:
Will close this and #738 since the library supports Microsoft.Extensions.Logging since 2025.0.0

PR #738, opened 2020-10-25
Tracing levels and actual trace event Ids

daviburg:
Fix event ids, they are supposed to be unique identifiers per trace event message, not a thread identifier.
Implemented trace event levels ("types")
Git ignore hidden VS 2019 folder.
Re-ordered out of order using declarations.


drieseng:
Most of these changes should be reverted:
* You cannot use string interpolation as we still support target framework that require us to use an old(er) C# compiler.
* I don't see why our diagnostics class should be public. You can register listeners without this.

I also don't like these "magic" event ids. Perhaps consider using an internal enum for this?
I'll try to have a closer look later this week.

daviburg:
> * You cannot use string interpolation as we still support target framework that require us to use an old(er) C# compiler.

SSH .NET targets Framework 3.5 as the oldest, and all framework version are compatible with C# version 7.3 per https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#defaults

daviburg:
> * I don't see why our diagnostics class should be public. You can register listeners without this.

Because the trace source Listeners needs to be accessible to enable programmatically adding listeners. See examples in public doc: https://docs.microsoft.com/en-us/dotnet/framework/debug-trace-profile/how-to-create-and-initialize-trace-sources#to-initialize-trace-sources-listeners-and-filters-without-a-configuration-file

```
            mySource.Listeners.Add(console);
            mySource.Listeners.Add(textListener);
```

daviburg:
> I also don't like these "magic" event ids. Perhaps consider using an internal enum for this?

Yes "magic" event ids are not ideal. I agree that an enum is the best way to address that. I will add an enum for trace event ids. It will also become handy if we latter either upgrade or supplement trace source with a modern ETW event source (they also use unique id per event).

daviburg:
@drieseng could you take a new look at this?

daviburg:
@drieseng ping for review / next step

daviburg:
@drieseng I noticed you are active today. Could you review the answers here?

drieseng:
> > * You cannot use string interpolation as we still support target framework that require us to use an old(er) C# compiler.
> 
> SSH .NET targets Framework 3.5 as the oldest, and all framework version are compatible with C# version 7.3 per https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#defaults

I have to use VS 2012 to build some of our older target frameworks (Silverlight, WP7).
Please remove usage of string interpolation.

src/Renci.SshNet/Channels/Channel.cs
```diff
@@ -551,7 +552,7 @@ protected virtual void Close()
                         var closeWaitResult = _session.TryWait(_channelClosedWaitHandle, ConnectionInfo.ChannelCloseTimeout);
                         if (closeWaitResult != WaitResult.Success)
                         {
-                            DiagnosticAbstraction.Log(string.Format("Wait for channel close not successful: {0:G}.", closeWaitResult));
+                            DiagnosticAbstraction.Log($"Wait for channel close not successful: {closeWaitResult:G}.", TraceEventType.Warning, TraceEventId.UnsuccessfulChannelCloseWait);
```
drieseng: Revert all use of string interpolation.

src/Renci.SshNet/Renci.SshNet.csproj
```diff
@@ -5,7 +5,7 @@
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <AssemblyName>Renci.SshNet</AssemblyName>
     <AssemblyOriginatorKeyFile>../Renci.SshNet.snk</AssemblyOriginatorKeyFile>
-    <LangVersion>5</LangVersion>
+    <LangVersion>6</LangVersion>
```
drieseng: Revert back to version 5.

src/Renci.SshNet/Abstractions/DiagnosticAbstraction.cs
```diff
@@ -1,34 +1,47 @@
...
+        public static readonly TraceSource Logging =
 #if DEBUG
-            new TraceSource("SshNet.Logging", SourceLevels.All);
+            new TraceSource(name: "SshNet.Logging", defaultLevel: SourceLevels.All);
 #else
             new TraceSource("SshNet.Logging");
 #endif // DEBUG
 #endif // FEATURE_DIAGNOSTICS_TRACESOURCE
 
-        [Conditional("DEBUG")]
```
drieseng, line 27: Please make it conditional for now.
We'll need to review the impact on performance.

drieseng:
I left some remarks in my latest review.
The most important remarks are:
1. Revert usage of string interpolation in order to keep support for older target frameworks.
1. Do not make DiagnosticAbstraction public.
1. Keep the `Log(...)` method conditional.

I know these last two remarks are problematic, but I'd like to discuss and review these in more detail.
Note that I definitely agree that a user should not have to rebuild **SSH.NET** to enable logging.

src/Renci.SshNet/Abstractions/DiagnosticAbstraction.cs
```diff
@@ -1,34 +1,47 @@
...
-using System.Threading;
-#endif // FEATURE_DIAGNOSTICS_TRACESOURCE
 
 namespace Renci.SshNet.Abstractions
 {
-    internal static class DiagnosticAbstraction
+    /// <summary>
+    /// Diagnostics for Renci library
+    /// </summary>
+    public static class DiagnosticAbstraction
```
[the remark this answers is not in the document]
daviburg: What is your opinion on ETW events? There are very broadly adopted across the industry, and well known for their minimal impact on performance.

src/Renci.SshNet/Abstractions/DiagnosticAbstraction.cs
```diff
@@ -1,34 +1,47 @@
...
         }
 
-        private static readonly TraceSource Loggging =
+        /// <summary>
+        /// The trace source for Renci
+        /// </summary>
+        public static readonly TraceSource Logging =
 #if DEBUG
-            new TraceSource("SshNet.Logging", SourceLevels.All);
+            new TraceSource(name: "SshNet.Logging", defaultLevel: SourceLevels.All);
```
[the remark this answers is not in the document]
daviburg: Good to know. Our team has broadly adopted named arguments to remove ambiguity when either constants or same typed parameters are used. But out of respect for this repo's standard, I will revert it here.

src/Renci.SshNet/Abstractions/DiagnosticAbstraction.cs
daviburg: That was the whole point of the PR, and the bug filed against this product years ago, to enable some tracing for retail build. There isn't much substance without that. I guess this problem then need solving through modernization of the tracing approach for SSH .NET as a whole. For that, well-known and broadly adopted ETW tracing should likely be preferred.

src/Renci.SshNet/Abstractions/DiagnosticAbstraction.cs
```diff
@@ -1,34 +1,47 @@
...
 
-        [Conditional("DEBUG")]
-        public static void Log(string text)
+        /// <summary>
+        /// Log the provided text
+        /// </summary>
+        /// <param name="text">The text string to log</param>
+        /// <param name="eventType">The trace event type</param>
+        /// <param name="id">A numeric identifier for the event.</param>
+        public static void Log(string text, TraceEventType eventType = TraceEventType.Verbose, TraceEventId id = TraceEventId._default)
```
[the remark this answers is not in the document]
daviburg: That would be a breaking change then, but I can go modify all callers.

daviburg:
> I left some remarks in my latest review.
> The most important remarks are:
> 
> 1. Revert usage of string interpolation in order to keep support for older target frameworks.
> 2. Do not make DiagnosticAbstraction public.
> 3. Keep the `Log(...)` method conditional.
> 
> I know these last two remarks are problematic, but I'd like to discuss and review these in more detail.
> Note that I definitely agree that a user should not have to rebuild **SSH.NET** to enable logging.

Ok, applied all 3 of these asks.

For modernization, I recommend replacing this legacy trace with `System.Diagnostics.Tracing.EventSource`. We would need a `SshNetEventSource `class derived from `EventSource `with an event source attribute (e.g. `[EventSource(Name = "SSH.NET")]`) and a method per event type. The new event ids introduced in this PR should be possible to re-use as-is.

The per event type methods would look something like:

```
        [Event(eventId: (int)AdapterEventIds.ListenerAcceptChannelClosedEarly, Level = EventLevel.Warning)]
        public void ListenerAcceptChannelClosedEarly(string listenerUri, string message)
        {
            this.WriteEvent(eventId: (int)AdapterEventIds.ListenerAcceptChannelClosedEarly, arg1: listenerUri, arg2: message);
        }
```

For convenience I typically add a lazy-initialized singleton like that:

```
        private static readonly Lazy<SshNetEventSource> EventSourceInstance = new Lazy<SshNetEventSource>(() => new SshNetEventSource());

        public static SshNetEventSource Log => EventSourceInstance.Value;

        /// <summary>
        /// Prevents construction of instances by others, enforcing singleton pattern as intended by .NET team authors of EventSource
        /// </summary>
        private SshNetEventSource()
        {
        }
```

There is an EventSource User's Guide by Cosmin Radu and Vance Morrison, architects from the .NET team, on this topic.

WojciechNagorski:
@daviburg Can you refresh this PR? We've already removed old .NET Frameworks.

Rob-Hague:
https://github.com/sshnet/SSH.NET/issues/45#issuecomment-3453444194

PR #746, opened 2020-12-02
Move IDisposable implementation declaration from inheritees to parent

bad-samaritan:
Hello,
I was using SSH.NET library in my project and I noticed that by following inheritance rules I fell into a trap.

My SFTP wrapper class contains private field of a general _AuthenticationMethod_ type.
It is is assigned in different constructors depending on selected authentication method.
```
	public class SftpFileHandler : IDisposable
	{
(...)
		private readonly AuthenticationMethod _authMethod;
(...)

		public SftpFileHandler(string handlerName, string address, int port, string? relativeWorkingDir, string? username, string? password, bool? keepConnectionOpen)
		{
(...)
			_authMethod = new PasswordAuthenticationMethod(username, password ?? "");
(...)
		}

		public SftpFileHandler(string handlerName, string address, int port, string? relativeWorkingDir, string? username, string privateKeyFilePath, string? privateKeyPass, bool? keepConnectionOpen)
		{
(...)
			_authMethod = new PrivateKeyAuthenticationMethod(Username, new PrivateKeyFile(privateKeyFilePath, privateKeyPass ?? ""));
(...)
		}
```
When I was implementing _Dispose()_ for _SftpFileHandler_ I noticed that _AuthenticationMethod_ does not require disposing which seemed a bit suspicious to me since it may contain sensitive data that may require proper disposing.
So I browsed descendants of _AuthenticationMethod_ to discover that all 4 of them implement _IDisposable_ interface.
That being said in other to properly dispose the whole _SftpFileHandler_ I need to do something like:
```
	public class SftpFileHandler : IDisposable
	{
(...)
		protected virtual void Dispose(bool disposing)
		{
			if (!_disposed)
			{
				if (_authMethod != null && _authMethod is IDisposable disposableAuthMethod)
				{
					disposableAuthMethod.Dispose();
				}
(...)
```
Which is always true...
Instead of just simply:
```
	public class SftpFileHandler : IDisposable
	{
(...)
		protected virtual void Dispose(bool disposing)
		{
			if (!_disposed)
			{
				if (_authMethod != null)
				{
					_authMethod .Dispose();
				}
(...)
```

I would like to propose a small improvement that would make disposing requirement move visible thus making it easier to maintain a little bit more secure code.
My suggestion is to move _IDisposable_ a level above -> to the _AuthenticationMethod_:
```
namespace Renci.SshNet
{
    /// <summary>
    /// Base class for all supported authentication methods
    /// </summary>
    public abstract class AuthenticationMethod : IAuthenticationMethod, IDisposable
    {
```

This way code analyzing tools would warn programmers before making a mistake of not disposing inheritee of _AuthenticationMethod_ class.

I tested changes mentioned above in my code and they work as expected.

What do you think about this improvement?


IgorMilavec:
Dispose pattern should be implemented in accordance with [Implement a Dispose method](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose) to avoid code duplication.

bad-samaritan:
Hey @IgorMilavec, thanks for the feedback.
I moved common Dispose stuff along with some fields to _AuthenticationMethod_ to avoid code duplication - see 8ca32fad.
Looks better?

IgorMilavec (changes requested):
Looks much cleaner now. Why not instantiate _authenticationCompleted in AuthenticationMethod's constructor? Also, do not declare finalizers, if they do not actually perform cleanup.

bad-samaritan:
> Why not instantiate _authenticationCompleted in AuthenticationMethod's constructor?

Because derived classes use different types for the _EventWaitHandle_ : _PrivateKeyAuthenticationMethod_ is using _ManualResetEvent_ and the rest three are using _AutoResetEvent_.
We can also instantiate it with _AutoResetEvent_ in _AuthenticationMethod_ ctor and override in _PrivateKeyAuthenticationMethod_ to be _ManualResetEvent_. I assumed the first one is clearer because it indicates that this field is used differently in derived classes but I don't have a strong opinion on that.

>Also, do not declare finalizers, if they do not actually perform cleanup.

Totally agree! Especially that classes with finalizers can take more time to dispose, so removing them is actaully small performance optimization.
However, this was another part of the code that I did not touch because it was not primary objective of this pull request. But sure, let's also correct that.

Updated code without finalizers on c34ad239.

IgorMilavec:
Huh, I totally missed different EventWaitHandles... I understand now. :)

IgorMilavec (approved):
Looking good to me.

bad-samaritan:
Pleasure doing business with you.

Thanks

Rob-Hague (approved):

PR #766, opened 2021-01-20
Implement rsync

ghost:
It would be nice to implement rsync in the same way as scp.


darkoperator:
This is interesting!

A9G-Data-Droid:
It looks like all of this code got merged in to main through other commits. Now this orphaned PR contains minimal differences with main.

This should be closed as complete.

WojciechNagorski:
I'm closing, nothing left here.

PR #791, opened 2021-03-05
Added filepart (.tmp) extension

Ledrunning:
Added another extension (.tmp) if the file has not yet been uploaded to the server, when it was uploaded, then the extension is changed to the original


WojciechNagorski:
I'm closing PR. This is an old entry with no votes.

PR #781, opened 2021-02-13
Minimal Changes Required for Pageant Support

shellster:
Minimum set of changes required for Pageant Support.

DO NOT MERGE WITHOUT FURTHER DISCUSSION.

See https://github.com/sshnet/SSH.NET/issues/31

PR #794, opened 2021-03-09
Agent auth and Keygen

darinkes:
Its based on the openssh_format_rsa branch, so ignore the first three commits.

"Allow to set PrivateKeyFile Key directly" and "Make ECDSA Key Bits accessible" Commits are needed for the WIP SSH.NET Extensions https://github.com/darinkes/SshNet.Keygen and https://github.com/darinkes/SshNet.Agent.

"Enable netstandard2.1 build" is a build fix, which respect netstandard versions newer than 2.0 for ECDSA Keys

### DIFF
--- a/src/Renci.SshNet/PrivateKeyFile.cs
+++ b/src/Renci.SshNet/PrivateKeyFile.cs
@@ -63,7 +63,7 @@ namespace Renci.SshNet
     /// </list>
     /// </para>
     /// </remarks>
-    public class PrivateKeyFile : IDisposable
+    public class PrivateKeyFile : IPrivateKeySource, IDisposable
     {
         private static readonly Regex PrivateKeyRegex = new Regex(@"^-+ *BEGIN (?<keyName>\w+( \w+)*) PRIVATE KEY *-+\r?\n((Proc-Type: 4,ENCRYPTED\r?\nDEK-Info: (?<cipherName>[A-Z0-9-]+),(?<salt>[A-F0-9]+)\r?\n\r?\n)|(Comment: ""?[^\r\n]*""?\r?\n))?(?<data>([a-zA-Z0-9/+=]{1,80}\r?\n)+)-+ *END \k<keyName> PRIVATE KEY *-+",
 #if FEATURE_REGEX_COMPILE
@@ -271,7 +271,7 @@ namespace Renci.SshNet
 
                     if (decryptedLength > blobSize - 4)
                         throw new SshException("Invalid passphrase.");
-                    
+
                     if (keyType == "if-modn{sign{rsa-pkcs1-sha1},encrypt{rsa-pkcs1v2-oaep}}")
                     {
                         var exponent = reader.ReadBigIntWithBits();//e
--- a/src/Renci.SshNet/PrivateKeyFile.cs
+++ b/src/Renci.SshNet/PrivateKeyFile.cs
@@ -524,8 +524,7 @@ namespace Renci.SshNet
                     throw new SshException("OpenSSH key type '" + keyType + "' is not supported.");
             }
 
-            //comment, we don't need this but we could log it, not sure if necessary
-            var comment = privateKeyReader.ReadString(Encoding.UTF8);
+            parsedKey.Comment = privateKeyReader.ReadString(Encoding.UTF8);
 
             //The list of privatekey/comment pairs is padded with the bytes 1, 2, 3, ...
             //until the total length is a multiple of the cipher block size.
--- a/src/Renci.SshNet/PrivateKeyFile.cs
+++ b/src/Renci.SshNet/PrivateKeyFile.cs
@@ -82,6 +82,15 @@ namespace Renci.SshNet
         /// <summary>
         /// Initializes a new instance of the <see cref="PrivateKeyFile"/> class.
         /// </summary>
+        /// <param name="key">The key.</param>
+        public PrivateKeyFile(Key key)
+        {
+            HostKey = new KeyHostAlgorithm(key.ToString(), key);
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PrivateKeyFile"/> class.
+        /// </summary>
         /// <param name="privateKey">The private key.</param>
         public PrivateKeyFile(Stream privateKey)
         {
--- a/src/Renci.SshNet/Security/Cryptography/EcdsaDigitalSignature.cs
+++ b/src/Renci.SshNet/Security/Cryptography/EcdsaDigitalSignature.cs
@@ -39,12 +39,12 @@ namespace Renci.SshNet.Security.Cryptography
             // for 521 sig_size is 132
             var sig_size = _key.KeyLength == 521 ? 132 : _key.KeyLength / 4;
             var ssh_data = new SshDataSignature(signature, sig_size);
-#if NETSTANDARD2_0
-            return _key.Ecdsa.VerifyData(input, ssh_data.Signature, _key.HashAlgorithm);
-#else
+#if NETFRAMEWORK
             var ecdsa = (ECDsaCng)_key.Ecdsa;
             ecdsa.HashAlgorithm = _key.HashAlgorithm;
             return ecdsa.VerifyData(input, ssh_data.Signature);
+#else
+            return _key.Ecdsa.VerifyData(input, ssh_data.Signature, _key.HashAlgorithm);
 #endif
         }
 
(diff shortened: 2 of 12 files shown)


darinkes:
Hey @drieseng,

Would be awesome if you could take a look here. A merge and release of SSH.NET would allow me to finalize my planned SSH.NET Extensions.

- https://github.com/darinkes/SshNet.Agent
- https://github.com/darinkes/SshNet.PuttyKeyFile
- https://github.com/darinkes/SshNet.Keygen

Thanks in advance

src/Renci.SshNet/Security/Cryptography/EcdsaDigitalSignature.cs
```diff
@@ -39,7 +39,7 @@ public override bool Verify(byte[] input, byte[] signature)
             // for 521 sig_size is 132
             var sig_size = _key.KeyLength == 521 ? 132 : _key.KeyLength / 4;
             var ssh_data = new SshDataSignature(signature, sig_size);
-#if NETSTANDARD2_0
+#if NETSTANDARD
```
IgorMilavec: Could you change this to ```#if NETSTANDARD2_1_OR_GREATER || NET5_0_OR_GREATER```, so it won't be included for NETSTANDARD1_3 (OK, not really an issue, but as long as we are here...) but it will be included for net5.0 targets (futureproofing)?

IgorMilavec:
Apart from minor conditionals change (see above) this code looks great to me.
I have found some other private key related PRs: #468, #595, #577, #468 and #179. Could you have a look if they affect ```IPrivateKeyFile``` design in any way?

Edit:
Found #705 and #781 which tackle Pageant support. Do you think we could have an extensibility interface that would support not only files (with direct access to key bytes) but also agents?

darinkes:
> Apart from minor conditionals change (see above) this code looks great to me. I have found some other private key related PRs: #468, #595, #577, #468 and #179. Could you have a look if they affect `IPrivateKeyFile` design in any way?

Can't see any conflict.

> Edit: Found #705 and #781 which tackle Pageant support. Do you think we could have an extensibility interface that would support not only files (with direct access to key bytes) but also agents?

This extension uses the same interface: https://github.com/darinkes/SshNet.Agent (Includes SSH-Agent und Pageant)
For SSH there is no difference between a Keyfile or Agent.

IgorMilavec:
Great! Can you please change the ```#if``` so we can ask Gert if this could be merged before the final release with legacy targets?

darinkes:
> Great! Can you please change the `#if` so we can ask Gert if this could be merged before the final release with legacy targets?

I actually think it should be an extra issue and PR. Or is there a reason we need it now?

IgorMilavec:
With the current condition, the implementation of ```EcdsaDigitalSignature``` will revert to pre-netstandard implementation when/if we target net5.0 or greater. Of course when we raise the target, we should check all the conditional compiles if they are affected. So I guess in the end it's a matter of personal preference. :)

IgorMilavec:
@drieseng can you please have a look at this PR? It provides a great extensibility point and would either close or support closing many issues.

drieseng:
@IgorMilavec I'll try to find time for this sometime this week. Sorry for the lag.

src/Renci.SshNet/Security/Cryptography/EcdsaDigitalSignature.cs
drieseng: I propose to reverse the checks (here and elsewhere):

**Now**
```cs
#if NETSTANDARD2_0
            return _key.Ecdsa.VerifyData(input, ssh_data.Signature, _key.HashAlgorithm);
#else
            var ecdsa = (ECDsaCng)_key.Ecdsa;
            ecdsa.HashAlgorithm = _key.HashAlgorithm;
            return ecdsa.VerifyData(input, ssh_data.Signature);
#endif
```

**Proposal**
```cs
#if NETFRAMEWORK
            var ecdsa = (ECDsaCng)_key.Ecdsa;
            ecdsa.HashAlgorithm = _key.HashAlgorithm;
            return ecdsa.VerifyData(input, ssh_data.Signature);
#else
            return _key.Ecdsa.VerifyData(input, ssh_data.Signature, _key.HashAlgorithm);
#endif
```

src/Renci.SshNet/Security/Cryptography/EcdsaKey.cs
```diff
@@ -18,7 +18,7 @@ public class EcdsaKey : Key, IDisposable
         internal const string ECDSA_P384_OID_VALUE = "1.3.132.0.34"; // Also called nistP384 or secP384r1
         internal const string ECDSA_P521_OID_VALUE = "1.3.132.0.35"; // Also called nistP521or secP521r1
 
-#if !NETSTANDARD2_0
+#if !NETSTANDARD
```
drieseng:
```suggestion
#if NETFRAMEWORK
```

src/Renci.SshNet/PrivateKeyFile.cs
```diff
@@ -74,10 +74,27 @@ public class PrivateKeyFile : IDisposable
 
         private Key _key;
 
+        private HostAlgorithm _hostKey;
+
         /// <summary>
         /// Gets the host key.
         /// </summary>
-        public HostAlgorithm HostKey { get; private set; }
+        public HostAlgorithm HostKey
```
drieseng: What was wrong with the automatic property?

src/Renci.SshNet/PrivateKeyFile.cs
```diff
@@ -63,7 +63,7 @@ namespace Renci.SshNet
     /// </list>
     /// </para>
     /// </remarks>
-    public class PrivateKeyFile : IDisposable
+    public class PrivateKeyFile : IPrivateKeyFile, IDisposable
```
drieseng: Do we want to make this one sealed to allow the JIT to devirtualize calls?
Not a big deal.

src/Renci.SshNet/IPrivateKeyFile.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Renci.SshNet.Security;
+
+namespace Renci.SshNet
+{
+    /// <summary>
+    /// Represents private key file interface.
+    /// </summary>
+    public interface IPrivateKeyFile
```
drieseng: There's nothing file-related in this interface. Do we want to extend it with file related properties or methods, or change the name of the interface?

drieseng (changes requested):

src/Renci.SshNet/IPrivateKeyFile.cs
darinkes: maybe IPrivateKeySource?

src/Renci.SshNet/PrivateKeyFile.cs
darinkes: I guess nothing. At least cant remember why I changed it and dont see any reason why it should be 🤷

drieseng (approved):

drieseng:
@darinkes Thanks!

darinkes:
@drieseng also thanks! I will update my extensions ASAP :)

WojciechNagorski:

stevesobol:
@darinkes am I missing something? I installed 2023.0 into my project, and I am trying to figure out how to use Pageant - the instructions at https://github.com/darinkes/SshNet.Agent don't work on 2023.0.0 as the Pageant class apparently does not exist... Thanks

zybexXL:
@stevesobol This PR adds the IPrivateKey hooks into SSH.Net, but the Pageant class is added by the SshNet.Agent code, not this one. You just need to compile SshNet.Agent with the latest SSH.Net as a reference.

stevesobol:
> @stevesobol This PR adds the IPrivateKey hooks into SSH.Net, but the Pageant class is added by the SshNet.Agent code, not this one. You just need to compile SshNet.Agent with the latest SSH.Net as a reference.

Thanks. Since (theoretically) there's no need to build @darinkes 's forked copy of SSH.NET anymore, I removed the Git submodule, removed the reference to said submodule from SshNet.agent's csproj file, and installed SSH.NET 2023.0.0 using Nuget.

I'm seeing 

`Cannot resolve symbol Renci` 

and failure to find a bunch of classes that should be provided by the Nuget package.

If it matters, I'm on Windows 10, but I am not using Visual Studio; I'm using JetBrains Rider.

darinkes:
@stevesobol I updated the repo and reenabled the build, now using the official SSH.NET release version.
A beta nuget can be found here: https://www.nuget.org/packages/SshNet.Agent

You should open issues on the SshNet.Agent github project.

PR #795, opened 2021-03-09
Unix socket forwarding

darinkes:
The branch is based on the "agent_auth" branch, so ignore the first six commits.

This adds Unix-Socket-Forwarding to SSH.NET.

[UnixDomainSocketEndPoint](https://docs.microsoft.com/de-de/dotnet/api/system.net.sockets.unixdomainsocketendpoint?view=net-5.0) is available in .NET since netstandard2.1 and on Windows 10 since 1803.

Still WIP, but functional. Would like to get some feedback on it.


kanpov:
Any update on this?

kanpov:
For now, I just made a hack where curl is executed through ssh and with the --unix-socket option. Not a great solution though

PR #800, opened 2021-03-27
Optimized incoming bytes queue. Release lock asap.

SwarajKetan:
Current incoming queue keeps the bytes. It is expensive operation since the OnDataReceived event gets the data in form of byte[]. If we read the bytes from array and enqueue them into incoming queue, it will hold the lock on incoming queue for long. This in-turn blocks the read operation from the incoming queue. Hence, the change from Queue<byte>  -> Queue<byte[]> improves performance.


jscarle:
@IgorMilavec @WojciechNagorski This can be closed as ShellStream has changed too much since this PR.

Issue #802, opened 2021-04-08
Allow writing to SshCommand


kilyanni:
By default you can't write to the Standard Input of a running `SshCommand` and can only write to a process running in SSH by using `CreateShellSession`. This causes a few problems in my use-case (sending protobuf-formatted packets back and forth):

- The input is mirrored back in the output, useful in an interactive context but not for automation
- Sending arbitrary binary data can cause problems (like accidentally sending `\u001b` which is escape in UTF16 or similar)
- A newline needs to be sent in order to actually commit input to the program

I've created a [workaround for this](https://gist.github.com/blenderfreaky/3d2d03a9af6eaa30ec534437774c779d) that uses evil reflection hacks for sending stuff directly to `SshCommand`.

# Proposed changes

- Expose `IChannelSession`, `IChannelSession` and `SshCommand._channel` as public

or

- Add a public method to `SshCommand` that calls `_channel.SendData(byte[] data, int offset, int size)`

I could create a PR for either solution, although I'm not sure yet how I would write a test for this.

Rob-Hague:
This was done in #1293 by way of `SshCommand.CreateInputStream` (https://sshnet.github.io/SSH.NET/examples.html#stream-data-to-a-command)

PR #803, opened 2021-04-10
Add .SendData method for SshCommand

kilyanni:
See  #802.

This  adds two methods `SshCommand.SendData` that expose `SshCommand._channel.SendData`s two overloads.

No tests yet, not quite sure how to test this.


kilyanni:
I've added a test using `cat -` to mirror the stdin back to stdout.
It currently uses a random to test bigger amounts of random data, I'm not sure if this is optimal. Suggestions are welcome

Rob-Hague:
Closing since a similar change was done in #1293. Thanks

Issue #190, opened 2017-03-05
Add support for FIPS


gargml:
Hi There,
FIPS is the U.S. government standard of security for cryptography.

No place in this project have indication whether or not it support FIPS for communication with target devices.

will it be possible to add such support ?
Please advise

Many thanks !
Gargml

darkoperator:
FIPs as a standard only defines a certain set of algorithms that can be used. If enabled on a modern system it will downgrade the security since the standard references old algorithms proven to be vulnerable. In fact Enabling FIPS via GPO on windows breaks a lot of stuff in .Net like AES and SHA256. Why support this? 
https://blogs.technet.microsoft.com/secguide/2014/04/07/why-were-not-recommending-fips-mode-anymore/ 
https://technet.microsoft.com/en-us/library/cc750357.aspx

If you are referring for the library to be FIPS certified .. that is a lot of money and work. do not know if @drieseng wants to spend the money and time for that.

vidyasesh22:
I am using SSH.Net version 2016.0.0 and I'm getting the following error when trying to connect to SFTP server
Failed to connect to server. Exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.InvalidOperationException: This implementation is not part of the Windows Platform FIPS validated cryptographic algorithms.
   at System.Security.Cryptography.MD5CryptoServiceProvider..ctor()

Does the SSH.Net version 2016.0.0 support FIPS ? is there a work around for this issue?

darkoperator:
No because FIPS in windows eliminates most modern crypto algorithms making connections insecure so when enabled it cripples .Net crypto APIs used by SSH.Net

cocowalla:
@darkoperator as much as I agree about the pointlessness of FIPS-mode nowadays, there are, regrettably, plenty organisations that demand it (especially in the defense sector).

Therefore I think it makes sense to use the FIPS-compliant version of algorithms wherever possible - for example, use the FIPS-compliant `SHA256CryptoServiceProvider` rather than `SHA256Managed`.

Having said all that, @vidyasesh22, if you have FIPS-mode enabled then you cannot use *any* implementation of MD5, as MD5 is not a valid FIPS algorithm. Actually, given how broken it is, you really shouldn't be using it *at all* these days, let alone in a FIPS environment.

darinkes:
Then these organisations should pay for this pointless feature :)

mauroa:
[...]

PR #806, opened 2021-04-14
add toggle to use fips

iamkrillin:
Pulled the code from @mcclouden on issue #190.  Added a toggle to BaseClient to enable/disable fips mode

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ project.lock.json
 
 # Build outputs
 build/target/
+/.vs
--- a/src/Renci.SshNet/Abstractions/CryptoAbstraction.cs
+++ b/src/Renci.SshNet/Abstractions/CryptoAbstraction.cs
@@ -56,9 +56,12 @@ namespace Renci.SshNet.Abstractions
 #endif // FEATURE_RNG_CREATE || FEATURE_RNG_CSP
 
 #if FEATURE_HASH_MD5
-        public static System.Security.Cryptography.MD5 CreateMD5()
+        public static System.Security.Cryptography.HashAlgorithm CreateMD5()
         {
-            return System.Security.Cryptography.MD5.Create();
+            if (SftpClient.UseFIPS)
+                return new System.Security.Cryptography.SHA1CryptoServiceProvider();
+            else
+                return System.Security.Cryptography.MD5.Create();
         }
 #else
         public static global::SshNet.Security.Cryptography.MD5 CreateMD5()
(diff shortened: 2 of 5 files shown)


sunilvijendra:
Any plan to merge the FIPS toggle changes?

A9G-Data-Droid:

daniellangemann:
I'll add my vote - any plans to merge the FIPS toggle changes? This seems like a safe and straightforward solution.

drieseng:
I'd prefer going for a builder pattern where the "user" explicitly adds whatever algorithms he/she wants.

epatrick:
Any update on this PR? We've run into conflicts with Windows FIPS turned on as part of AWS' current STIG for Windows 2022 boxes.

A9G-Data-Droid:
@epatrick My `develop` branch works on FIPS systems: https://github.com/A9G-Data-Droid/SSH.NET

This library hasn't seen an update since Jan 24, 2021 so I'm not waiting around for the next version to support it.

okay9109:
Will this PR be included anytime soon?

Rob-Hague:
It should not be necessary. AFAIK, on .NET Framework the `Create` methods already return a `CryptoServiceProvider` instance, and .NET Core forwards implementations to the OS through which FIPS is controlled [(.NET Core Federal Information Processing Standard (FIPS) compliance)](https://learn.microsoft.com/en-us/dotnet/standard/security/fips-compliance). `CryptoServiceProvider` and other derived cryptographic types are obsolete ([SYSLIB0021](https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0021)).

WojciechNagorski:
@Rob-Hague So when we release a new version supporting .NET 7, could we close this PR?

Rob-Hague:
There is ~~maybe~~ still a need to disable MD5, but the rest of the changes (`CryptoServiceProvider` etc.) should not be needed.

Disclaimer: not an expert

okay9109:
> on .NET Framework the Create methods already return a CryptoServiceProvider instance

I see, and agree for SHA implementations.
From my understanding, on .NET Framework `MD5CryptoServiceProvider` [does not use an alternative implementation](https://github.com/microsoft/referencesource/blob/51cf7850defa8a17d815b4700b67116e3fa283c2/mscorlib/system/security/cryptography/md5cryptoserviceprovider.cs#L35) if the FIPS policy is enabled, and it uses `clr.dll` [underneath](https://github.com/microsoft/referencesource/blob/51cf7850defa8a17d815b4700b67116e3fa283c2/mscorlib/system/security/cryptography/utils.cs#L1073) (calling with QCall), which is not validated by NIST for FIPS compliance. Plus, the algorithm itself is not permitted.

So this:

> There is still a need to disable MD5, but the rest of the changes (CryptoServiceProvider etc.) should not be needed.

is true from my understanding.

Should there be a PR to treat MD5 only then?

Disclaimer: also not an expert, just trying to understand FIPS enforcement for .NET (and going insane haha)

Rob-Hague:
Ok I spoke too assertively and you may replace my comments with "I don't know" but what you have said makes sense, and my answer to

> Should there be a PR to treat MD5 only then?

is "I guess", but I'm not sure how easy it will be to test

A9G-Data-Droid:
In relation to this PR, you could eliminate both the `bool UseFIPS` and the MD5 methods entirely. That would make it work both in and outside of a FIPS mode OS. Some sort of PR is still required. You should be seeing those warnings about the MD5 methods being obsolete. We need to do something to remove them.

A9G-Data-Droid:
This PR can be closed as the latest version is working on FIPS enabled computers. I think this was closed by https://github.com/sshnet/SSH.NET/pull/1293

Issue #665, opened 2020-04-25
[RFC] Target framework support


drieseng:
**SSH.NET** has always supported a broad set of target frameworks. Even though this is important to our users and myself, our code base and release process are more and more hindered by this ideal.

Therefore I'd like to question whether we should continue to support the following target frameworks:

* .NET Framework 3.5
* Silverlight 4
* Silverlight 5
* Windows Phone 7.1
* Windows Phone 8.0
* Universal Windows Platform 10 [1]

Some of these target framework are no longer supported by **Microsoft** and require software that is no longer available for download.

This is by no means an announcement to drop support for these target frameworks, but it should be clear that the continued support comes with a cost.

The time I or anyone can spend on this project is limited. Any means to reduce complexity contributes to efficiency, and in turn brings profit to the **SSH.NET** user base.

I'd like to use this issue to sollicit feedback from our users as to how important it is to continue support for the afore mentioned target frameworks.

I encourage you to be vocal!

Thanks!
Gert

[1] UWP 10 - even though it can hardly be considered legacy - is not (or does not appear to be) supported in SDK projects. If someone can assist us with this, we can continue to support it without much extra effort.

darkoperator:
Im all for dropping the old frameworks. As you mentioned it takes time away from adding and fixing issues. Faster release cycles means over a million users of my PowerShell module can get the latest and greatest faster plus I can address faster their request for more modern frameworks so for me it is a win win.

chrgraefe:
Please go ahead and drop support of this TFMs.
You can branch it, if anybody needs it for fast code checking or forking. In rare cases it might be useful for backporting things, but this should be an exception.

brent-williams:
Strongly in favour of dropping older platforms.  If even MS has dropped support, let alone the software can't be downloaded, then surely you're in good company to do the same if it's hindering development on current platforms.

rjvdboon:
It seems to me that anyone still targeting those obsolete platforms is unlikely to upgrade their dependencies.
B.t.w. how many acknowledgements do you need before deciding that the 2.0.0.0 release will have a more limited (saner) set of target frameworks? The major release you are preparing would be the perfect cutoff point for the reduction in TFMs.

mrjfalk:
I'm still using the library while targeting Universal Windows Platform 10, and just updated to latest beta version. Would be great if UWP support could remain. Although as you also target .NET standard 2.0 that would probably be enough seeing that works for many versions of UWP as well.

[...]

PR #819, opened 2021-05-04
Add async support to SftpClient and SftpFileStream

### LINKED ISSUE #153: Add support for a Task based async API

IgorMilavec:
Core async support requested by #153.

Added async support to BaseClient: (update 2021-05-11)
 - ConnectAsync

Added async support to SftpClient:
- ListDirectoryAsync
- GetStatusAsync
- OpenAsync
- DeleteFileAsync
- RenameFileAsync
- DownloadFileAsync
- UploadFileAsync

Added async support to SftpFileStream:
- OpenAsync
- ReadAsync
- WriteAsync
- FlushAsync

Added net46 target to use async in .NET Framework.

This PR is intended to resolve #153. 
This PR is intended to supersede #300 and #661.


src/Renci.SshNet/Sftp/SftpFileStream.cs
```diff
@@ -166,6 +169,28 @@ public virtual byte[] Handle
         /// </value>
```
vanillajonathan, line 170: This property lacks documentation.

src/Renci.SshNet/Sftp/SftpFileStream.cs
```diff
@@ -166,6 +169,28 @@ public virtual byte[] Handle
...
         public TimeSpan Timeout { get; set; }
 
+        private SftpFileStream(ISftpSession session, string path, FileAccess access, int bufferSize, byte[] handle, long position)
+        {
+            Timeout = TimeSpan.FromSeconds(30);
```
vanillajonathan, line 175: This 30 seconds timeout should be mentioned in documentation.

src/Renci.SshNet/Sftp/SftpFileStream.cs
```diff
@@ -267,6 +292,110 @@ internal SftpFileStream(ISftpSession session, string path, FileMode mode, FileAc
             }
         }
 
+#if FEATURE_TAP
+        internal static async Task<SftpFileStream> OpenAsync(ISftpSession session, string path, FileMode mode, FileAccess access, int bufferSize, CancellationToken cancellationToken)
+        {
+            if (session == null)
+                throw new SshConnectionException("Client not connected.");
+            if (path == null)
+                throw new ArgumentNullException("path");
```
vanillajonathan, line 302: Use `nameof(path)` instead of string.

src/Renci.SshNet/Sftp/SftpFileStream.cs
```diff
@@ -267,6 +292,110 @@ internal SftpFileStream(ISftpSession session, string path, FileMode mode, FileAc
...
+            if (bufferSize <= 0)
+                throw new ArgumentOutOfRangeException("bufferSize");
```
vanillajonathan: Add message parameter "bufferSize must not be less than 0" or something like this.

src/Renci.SshNet/Sftp/SftpFileStream.cs
```diff
@@ -740,6 +1012,84 @@ public override void Write(byte[] buffer, int offset, int count)
...
+        public override async Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            if (buffer == null)
+                throw new ArgumentNullException("buffer");
+            if (offset < 0)
+                throw new ArgumentOutOfRangeException("offset");
+            if (count < 0)
+                throw new ArgumentOutOfRangeException("count");
+            if ((buffer.Length - offset) < count)
+                throw new ArgumentException("Invalid array range.");
```
vanillajonathan, line 1031: Specify second parameter to `ArgumentException`, `nameof(buffer)`.

src/Renci.SshNet/SftpClient.cs
```diff
@@ -476,6 +542,62 @@ public IEnumerable<SftpFile> ListDirectory(string path, Action<int> listCallback
...
+        /// <returns>
+        /// A <see cref="Task{IEnumerable}"/> that represents the asynchronous list operation.
+        /// The task result contains an enumerable collection of <see cref="SftpFile"/> for the files in the directory specified by <paramref name="path" />.
+        /// </returns>
+        /// <exception cref="ArgumentNullException"><paramref name="path" /> is <b>null</b>.</exception>
+        /// <exception cref="SshConnectionException">Client is not connected.</exception>
+        /// <exception cref="SftpPermissionDeniedException">Permission to list the contents of the directory was denied by the remote host. <para>-or-</para> A SSH command was denied by the server.</exception>
+        /// <exception cref="SshException">A SSH error where <see cref="Exception.Message" /> is the message from the remote host.</exception>
+        /// <exception cref="ObjectDisposedException">The method was called after the client was disposed.</exception>
+        public async Task<IEnumerable<SftpFile>> ListDirectoryAsync(string path, CancellationToken cancellationToken)
```
vanillajonathan, line 555: Consider returning `ICollection` or `IReadOnlyCollection` rather than `IEnumerable` as per the [Microsoft Guidelines for Collections](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/guidelines-for-collections).

> DO use Collection<T> or a subclass of Collection<T> for properties or return values representing read/write collections.

src/Renci.SshNet/Sftp/SftpFileStream.cs
IgorMilavec: No change here?

src/Renci.SshNet/Sftp/SftpFileStream.cs
IgorMilavec: Agree, but this is my "blame" only as a result of refactoring. The whole goal of my PR is to support async and stop using timeouts.

src/Renci.SshNet/Abstractions/DnsAbstraction.cs
```diff
@@ -87,5 +88,23 @@ public static IPAddress[] GetHostAddresses(string hostNameOrAddress)
...
+        /// Returns the Internet Protocol (IP) addresses for the specified host.
+        /// </summary>
+        /// <param name="hostNameOrAddress">The host name or IP address to resolve</param>
+        /// <returns>
+        /// A task with result of an array of type <see cref="IPAddress"/> that holds the IP addresses for the host that
+        /// is specified by the <paramref name="hostNameOrAddress"/> parameter.
+        /// </returns>
+        /// <exception cref="ArgumentNullException"><paramref name="hostNameOrAddress"/> is <c>null</c>.</exception>
+        /// <exception cref="SocketException">An error is encountered when resolving <paramref name="hostNameOrAddress"/>.</exception>
+        public static Task<IPAddress[]> GetHostAddressesAsync(string hostNameOrAddress)
```
vanillajonathan, line 106: Should this be a `ICollection<IPAddress>` instread of a `IPAddress[]`?

[Guidelines for Collections - Framework Design Guidelines | Microsoft Docs](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/guidelines-for-collections)

src/Renci.SshNet/Sftp/SftpFileStream.cs
vanillajonathan: Oh, I guess it it inherits the documentation from the `Stream` class.

schaveyt:
Any chance this could also target .net5 or .net6 as well

IgorMilavec:
The library already targets netstandard2.0, so you can reference it from net5 or net6 projects.
It would only be beneficial for the library to target net5+ if it would use any net5+ features, like Spans etc. But this is out of scope of this issue and this PR. IMO we should remove legacy targets (please see #665) before adding any modern code.

schaveyt:
I did not connect those dots. Thank you for clarifying 👍

paulmaybee:
This addition looks like it has a great deal of value. Is there a timetable or estimate for when the PR will be completed?

IgorMilavec:
This PR if functionally complete IMO. It should also be stable, I have been running it in production since July. Now we need @drieseng to review and merge it.

drieseng:
@IgorMilavec Thanks, I'll try to find time in the coming days.

IgorMilavec:
Figured this is the last chance to modify UploadFileAsync signature before this becomes a breaking change... :)

src/Renci.SshNet/SftpClient.cs
```diff
@@ -985,7 +985,7 @@ public async Task UploadFileAsync(Stream input, string path, bool canOverride, C
             CheckDisposedOrNotConnected();
             cancellationToken.ThrowIfCancellationRequested();
 
-            using (SftpFileStream output = await SftpFileStream.OpenAsync(_sftpSession, path, canOverride ? FileMode.Create : FileMode.CreateNew, FileAccess.Write, (int)_bufferSize, cancellationToken).ConfigureAwait(false))
+            using (SftpFileStream output = await SftpFileStream.OpenAsync(_sftpSession, path, (FileMode)createMode, FileAccess.Write, (int)_bufferSize, cancellationToken).ConfigureAwait(false))
```
zybexXL: Note that to resume an upload using Append mode, the caller will also need to position the input stream at the correct offset, or else the entire input stream will be appended to the remote file.
I added a similar feature to SftpClient in PR #864, but I used the current input offset to determine if the stream should append or not.

IgorMilavec: This feature is intended to support scenario where one needs to concatenate multiple local files to a single remote file.
Resumption could be done with the new interface, but it is quite hard to do correctly in a general sense. As you already mentioned, the caller would need to seek to correct offset, but the caller first needs to determine what this offset is. So yes, it is now supported but the application needs to handle it correctly.

zybexXL: Right. I have been using #864's resume feature for a couple of months, it works nicely for both upload and download. But the app does need to check the remote length first to set the proper stream offset.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -962,6 +962,24 @@ public SftpFileAttributes RequestFStat(byte[] handle, bool nullOnError)
...
+                    response => tcs.TrySetResult(response.Attributes),
+                    response => tcs.TrySetException(GetSftpException(response))));
+
+                return await tcs.Task.ConfigureAwait(false);
+            }
+        }
+#endif
+
         /// <summary>
         /// Performs SSH_FXP_SETSTAT request.
```
vanillajonathan, line 985: Consider wrapping `SSH_FXP_SETSTAT` in a `<c>` tag.

IgorMilavec:
Incorporated fix from #824 into Async code

src/Renci.SshNet/Session.cs
```diff
@@ -669,6 +672,129 @@ public void Connect()
...
+            try
+            {
+                // ToDo: Migrate to SemaphoreSlim to enable async awaiting
+                // ToDo: SHould we really limit this globally here, or leave it to the app?
+                AuthenticationConnection.Wait();
+
+                if (IsConnected)
+                    return;
+
+                //lock (this)   // ToDo: Why do we need double locking at all?
```
drieseng: We do not want multiple threads to establish a connection in parallel (using the same client instance, that is).

**Edit:**
As I mentioned in another comment, I'm ok with relaxing this if we clearly document this in the release notes and the /// doc.

src/Renci.SshNet/Renci.SshNet.csproj
```diff
@@ -7,7 +7,7 @@
     <AssemblyOriginatorKeyFile>../Renci.SshNet.snk</AssemblyOriginatorKeyFile>
     <LangVersion>5</LangVersion>
     <SignAssembly>true</SignAssembly>
-    <TargetFrameworks>net35;net40;netstandard1.3;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net35;net40;net46;netstandard1.3;netstandard2.0</TargetFrameworks>
```
drieseng: I see no reason for adding .NET 4.6 support at this stage.
You remark about **SshNet.Security.Cryptography** is not correct. It does have a strong name.
If you really want to add a new target framework for the desktop, then I'd add .NET Framework 4.8.

src/Renci.SshNet/Connection/ProxyConnector.cs
```diff
@@ -0,0 +1,72 @@
+﻿using Renci.SshNet.Common;
```
drieseng: Nit: unused using.

src/Renci.SshNet/Connection/ProxyConnector.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        public ProxyConnector(ISocketFactory socketFactory) :
+            base(socketFactory)
+        {
+        }
+
+        protected abstract void HandleProxyConnect(IConnectionInfo connectionInfo, Socket socket);
+
+#if FEATURE_TAP
+        // ToDo: Performs async/sync fallback, true async version should be implemented in derived classes
```
drieseng, line 23: Will you be implementing these?

src/Renci.SshNet/Connection/Socks5Connector.cs
```diff
@@ -2,6 +2,7 @@
 using Renci.SshNet.Common;
 using System;
 using System.Net.Sockets;
+using System.Threading;
```
drieseng: Nit: unused using.

src/Renci.SshNet/ISftpClient.cs
```diff
@@ -2,8 +2,12 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Text;
+using System.Threading;
```
drieseng: Please move this under `#if FEATURE_TAP`.

src/Renci.SshNet/Session.cs
```diff
@@ -669,6 +672,129 @@ public void Connect()
...
+        public async Task ConnectAsync(CancellationToken cancellationToken)
+        {
+            if (IsConnected)
+                return;
+
+            try
+            {
+                // ToDo: Migrate to SemaphoreSlim to enable async awaiting
+                // ToDo: SHould we really limit this globally here, or leave it to the app?
+                AuthenticationConnection.Wait();
```
drieseng: This was introduced by @olegkap in 3e19ae132ea3f6050b0c6b7231b05cac5d436667.
SSH servers limit the number of simultaneous unauthenticated connections (eg. https://man.openbsd.org/sshd_config#MaxStartups).

Personally, I'm ok with removing this for the async variant. If we do this, we should clearly document this in the release notes.

src/Renci.SshNet/Sftp/SftpFileStream.cs
drieseng: "'Cannot be less than or equal to zero."

src/Renci.SshNet/Sftp/SftpFileStream.cs
```diff
@@ -267,6 +292,110 @@ internal SftpFileStream(ISftpSession session, string path, FileMode mode, FileAc
...
+                    try
+                    {
+                        handle = await session.OpenAsync(path, flags | Flags.Truncate, cancellationToken).ConfigureAwait(false);
+                        flags |= Flags.Truncate;
+                    }
+                    catch (TaskCanceledException)// ToDo: when (cancellationToken.IsCancellationRequested)
+                    {
+                        throw;
+                    }
+                    catch (Exception)
```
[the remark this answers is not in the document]
drieseng: When **FileMode** is **Create**, I think we should just use **Flags.CreateNewOrOpen | Flags.Truncate**.
Let me run some checks later.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -362,15 +413,15 @@ public byte[] RequestOpen(string path, Flags flags, bool nullOnError = false)
             {
                 var request = new SftpOpenRequest(ProtocolVersion, NextRequestId, path, Encoding, flags,
                     response =>
-                        {
```
drieseng: Please undo these formatting changes.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -456,10 +525,10 @@ public void RequestClose(byte[] handle)
             {
                 var request = new SftpCloseRequest(ProtocolVersion, NextRequestId, handle,
                     response =>
-                        {
```
drieseng: Please undo these formatting changes.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -444,6 +495,24 @@ public byte[] EndOpen(SftpOpenAsyncResult asyncResult)
             }
         }
 
+#if FEATURE_TAP
+        public async Task<byte[]> OpenAsync(string path, Flags flags, CancellationToken cancellationToken)
```
drieseng: Please rename to **RequestOpenAsync** and move it next to the **RequestOpen** method.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -472,6 +541,33 @@ public void RequestClose(byte[] handle)
             }
         }
 
+#if FEATURE_TAP
+        public async Task CloseAsync(byte[] handle, CancellationToken cancellationToken)
```
drieseng: Please rename to **RequestCloseAsync** and move next to **RequestClose** method.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -695,15 +846,15 @@ public SftpFileAttributes RequestLStat(string path)
             {
                 var request = new SftpLStatRequest(ProtocolVersion, NextRequestId, path, Encoding,
                     response =>
-                        {
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -788,15 +939,15 @@ public SftpFileAttributes RequestFStat(byte[] handle, bool nullOnError)
             {
                 var request = new SftpFStatRequest(ProtocolVersion, NextRequestId, handle,
                     response =>
-                        {
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -824,10 +993,10 @@ public void RequestSetStat(string path, SftpFileAttributes attributes)
             {
                 var request = new SftpSetStatRequest(ProtocolVersion, NextRequestId, path, Encoding, attributes,
                     response =>
-                        {
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -853,10 +1022,10 @@ public void RequestFSetStat(byte[] handle, SftpFileAttributes attributes)
             {
                 var request = new SftpFSetStatRequest(ProtocolVersion, NextRequestId, handle, attributes,
                     response =>
-                        {
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -885,15 +1054,15 @@ public byte[] RequestOpenDir(string path, bool nullOnError = false)
             {
                 var request = new SftpOpenDirRequest(ProtocolVersion, NextRequestId, path, Encoding,
                     response =>
-                        {
-                            handle = response.Handle;
-                            wait.Set();
-                        },
+                    {
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -923,18 +1110,18 @@ public byte[] RequestOpenDir(string path, bool nullOnError = false)
             {
                 var request = new SftpReadDirRequest(ProtocolVersion, NextRequestId, handle,
                     response =>
-                        {
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -961,10 +1177,10 @@ public void RequestRemove(string path)
             {
                 var request = new SftpRemoveRequest(ProtocolVersion, NextRequestId, path, Encoding,
                     response =>
-                        {
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -989,10 +1233,10 @@ public void RequestMkDir(string path)
             {
                 var request = new SftpMkDirRequest(ProtocolVersion, NextRequestId, path, Encoding,
                     response =>
-                        {
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -1017,10 +1261,10 @@ public void RequestRmDir(string path)
             {
                 var request = new SftpRmDirRequest(ProtocolVersion, NextRequestId, path, Encoding,
                     response =>
-                        {
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -1051,15 +1295,15 @@ public void RequestRmDir(string path)
             {
                 var request = new SftpRealPathRequest(ProtocolVersion, NextRequestId, path, Encoding,
                     response =>
-                        {
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -1145,15 +1417,15 @@ public SftpFileAttributes RequestStat(string path, bool nullOnError = false)
             {
                 var request = new SftpStatRequest(ProtocolVersion, NextRequestId, path, Encoding,
                     response =>
-                        {
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -1239,10 +1511,10 @@ public void RequestRename(string oldPath, string newPath)
             {
                 var request = new SftpRenameRequest(ProtocolVersion, NextRequestId, oldPath, newPath, Encoding,
                     response =>
-                        {
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -1276,15 +1576,15 @@ public void RequestRename(string oldPath, string newPath)
             {
                 var request = new SftpReadLinkRequest(ProtocolVersion, NextRequestId, path, Encoding,
                     response =>
-                        {
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -1317,10 +1617,10 @@ public void RequestSymLink(string linkpath, string targetpath)
             {
                 var request = new SftpSymLinkRequest(ProtocolVersion, NextRequestId, linkpath, targetpath, Encoding,
                     response =>
-                        {
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -1355,10 +1655,10 @@ public void RequestPosixRename(string oldPath, string newPath)
             {
                 var request = new PosixRenameRequest(ProtocolVersion, NextRequestId, oldPath, newPath, Encoding,
                     response =>
-                        {
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -1395,15 +1695,15 @@ public SftpFileSytemInformation RequestStatVfs(string path, bool nullOnError = f
             {
                 var request = new StatVfsRequest(ProtocolVersion, NextRequestId, path, Encoding,
                     response =>
-                        {
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -1395,15 +1695,15 @@ public SftpFileSytemInformation RequestStatVfs(string path, bool nullOnError = f
...
-                            information = response.GetReply<StatVfsReplyInfo>().Information;
-                            wait.Set();
-                        },
+                    {
+                        information = response.GetReply<StatVfsReplyInfo>().Information;
+                        wait.Set();
+                    },
                     response =>
-                        {
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -1443,15 +1767,15 @@ internal SftpFileSytemInformation RequestFStatVfs(byte[] handle, bool nullOnErro
             {
                 var request = new FStatVfsRequest(ProtocolVersion, NextRequestId, handle,
                     response =>
-                        {
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -1487,10 +1811,10 @@ internal void HardLink(string oldPath, string newPath)
             {
                 var request = new HardLinkRequest(ProtocolVersion, NextRequestId, oldPath, newPath,
                     response =>
-                        {
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/SftpClient.cs
drieseng: This should be **IAsyncEnumerable\<SftpFile>**. It's only supported on .NET 4.6.1 or higher.
For .NET 4.6.1 and .NET Standard 2.0, you need to add a reference to **Microsoft.Bcl.AsyncInterfaces**.

src/Renci.SshNet/SftpClient.cs
```diff
@@ -640,6 +758,37 @@ public void DownloadFile(string path, Stream output, Action<ulong> downloadCallb
...
+        /// <param name="cancellationToken">The <see cref="CancellationToken"/> to observe.</param>
+        /// <returns>A <see cref="Task"/> that represents the asynchronous download operation.</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="path" /> is <b>null</b> -or- <paramref name="output" /> is <b>null</b>.</exception>
+        /// <exception cref="SshConnectionException">Client is not connected.</exception>
+        /// <exception cref="SftpPermissionDeniedException">Permission to perform the operation was denied by the remote host. <para>-or-</para> A SSH command was denied by the server.</exception>
+        /// <exception cref="SftpPathNotFoundException"><paramref name="path"/> was not found on the remote host.</exception>/// 
+        /// <exception cref="SshException">A SSH error where <see cref="Exception.Message" /> is the message from the remote host.</exception>
+        /// <exception cref="ObjectDisposedException">The method was called after the client was disposed.</exception>
+        public async Task DownloadFileAsync(string path, Stream output, CancellationToken cancellationToken)
```
drieseng: I propose to leave out this method. See my comment on `UploadFileAsync(...)` for more info.

src/Renci.SshNet/SftpClient.cs
```diff
@@ -809,6 +958,39 @@ public void UploadFile(Stream input, string path, bool canOverride, Action<ulong
...
+        /// <returns>A <see cref="Task"/> that represents the asynchronous upload operation.</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="input" /> is <b>null</b> -or- <paramref name="path" /> is <b>null</b>.</exception>
+        /// <exception cref="SshConnectionException">Client is not connected.</exception>
+        /// <exception cref="SftpPermissionDeniedException">Permission to upload the file was denied by the remote host. <para>-or-</para> A SSH command was denied by the server.</exception>
+        /// <exception cref="SshException">A SSH error where <see cref="Exception.Message" /> is the message from the remote host.</exception>
+        /// <exception cref="ObjectDisposedException">The method was called after the client was disposed.</exception>
+        /// <remarks>
+        /// Method calls made by this method to <paramref name="input" />, may under certain conditions result in exceptions thrown by the stream.
+        /// </remarks>
+        public async Task UploadFileAsync(Stream input, string path, UploadMode createMode, CancellationToken cancellationToken)
```
drieseng: At first, I was inclined to say "use System.IO.FileMode directly, and remove the UploadMode enum".
But I'd actually prefer to remove this method altogether. I don't see an advantage over using OpenAsync(string path, FileMode mode, FileAccess access, CancellationToken cancellationToken) and having users invoke Stream.CopyToAsync(...) themselves. This gives them also control over the buffer size that is used to read from the input stream. Right now, you use our "optimal" write buffer size for this.

src/Renci.SshNet/Session.cs
```diff
@@ -669,6 +672,129 @@ public void Connect()
...
+                    RegisterMessage("SSH_MSG_NEWKEYS");
+
+                    // Some server implementations might sent this message first, prior to establishing encryption algorithm
+                    RegisterMessage("SSH_MSG_USERAUTH_BANNER");
+
+                    // Mark the message listener threads as started
+                    _messageListenerCompleted.Reset();
+
+                    // Start incoming request listener
+                    // ToDo: Make message pump async, to not consume a thread for every session
```
drieseng: Are you planning on working on this?

src/Renci.SshNet/Sftp/SftpFileStream.cs
```diff
@@ -267,6 +292,110 @@ internal SftpFileStream(ISftpSession session, string path, FileMode mode, FileAc
...
+            if (mode == FileMode.Append)
+            {
+                try
+                {
+                    var attributes = await session.RequestFStatAsync(handle, cancellationToken).ConfigureAwait(false);
+                    position = attributes.Size;
+                }
+                catch
+                {
+                    session.RequestClose(handle);
```
drieseng: Use async variant.

src/Renci.SshNet/Sftp/SftpFileStream.cs
```diff
@@ -359,7 +513,7 @@ public override int Read(byte[] buffer, int offset, int count)
                     var bytesAvailableInBuffer = _bufferLen - _bufferPosition;
                     if (bytesAvailableInBuffer <= 0)
                     {
-                        var data = _session.RequestRead(_handle, (ulong) _position, (uint) _readBufferSize);
```
drieseng: Undo formatting change.

src/Renci.SshNet/Sftp/SftpFileStream.cs
```diff
@@ -438,6 +592,124 @@ public override int Read(byte[] buffer, int offset, int count)
...
+                throw new ArgumentNullException("buffer");
+            if (offset < 0)
+                throw new ArgumentOutOfRangeException("offset");
+            if (count < 0)
+                throw new ArgumentOutOfRangeException("count");
+            if ((buffer.Length - offset) < count)
+                throw new ArgumentException("Invalid array range.");
+
+            // Lock down the file stream while we do this.
+            //lock (_lock)  Do we need to be thread safe?
```
drieseng: I'm ok with relaxing this if we clearly document this in the release notes and the /// doc.
For the /// doc, you can just add `<threadsafety static="true" instance="false"/>` to the class-level doc.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -51,7 +54,7 @@ public uint NextRequestId
         {
             get
             {
-                return (uint) Interlocked.Increment(ref _requestId);
+                return (uint)Interlocked.Increment(ref _requestId);
```
drieseng: Undo formatting change.

src/Renci.SshNet/SftpClient.cs
```diff
@@ -2123,26 +2358,26 @@ private void InternalUploadFile(Stream input, string path, Flags flags, SftpUplo
 
                 if (bytesRead > 0)
                 {
-                    var writtenBytes = offset + (ulong) bytesRead;
+                    var writtenBytes = offset + (ulong)bytesRead;
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/SftpClient.cs
```diff
@@ -2066,7 +2301,7 @@ private void InternalDownloadFile(string path, Stream output, SftpDownloadAsyncR
 
                     output.Write(data, 0, data.Length);
 
-                    totalBytesRead += (ulong) data.Length;
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/SftpClient.cs
```diff
@@ -1403,7 +1638,7 @@ public SftpFileStream OpenWrite(string path)
         {
             CheckDisposed();
 
-            return new SftpFileStream(_sftpSession, path, FileMode.OpenOrCreate, FileAccess.Write, (int) _bufferSize);
+            return new SftpFileStream(_sftpSession, path, FileMode.OpenOrCreate, FileAccess.Write, (int)_bufferSize);
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/SftpClient.cs
```diff
@@ -1353,9 +1561,36 @@ public SftpFileStream Open(string path, FileMode mode, FileAccess access)
         {
             CheckDisposed();
 
-            return new SftpFileStream(_sftpSession, path, mode, access, (int) _bufferSize);
+            return new SftpFileStream(_sftpSession, path, mode, access, (int)_bufferSize);
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/SftpClient.cs
```diff
@@ -1162,7 +1370,7 @@ public SftpFileStream Create(string path)
         {
             CheckDisposed();
 
-            return new SftpFileStream(_sftpSession, path, FileMode.Create, FileAccess.ReadWrite, (int) _bufferSize);
+            return new SftpFileStream(_sftpSession, path, FileMode.Create, FileAccess.ReadWrite, (int)_bufferSize);
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/SftpClient.cs
```diff
@@ -1141,7 +1349,7 @@ public StreamWriter AppendText(string path, Encoding encoding)
             if (encoding == null)
                 throw new ArgumentNullException("encoding");
 
-            return new StreamWriter(new SftpFileStream(_sftpSession, path, FileMode.Append, FileAccess.Write, (int) _bufferSize), encoding);
+            return new StreamWriter(new SftpFileStream(_sftpSession, path, FileMode.Append, FileAccess.Write, (int)_bufferSize), encoding);
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/SftpClient.cs
```diff
@@ -2123,26 +2358,26 @@ private void InternalUploadFile(Stream input, string path, Flags flags, SftpUplo
 
                 if (bytesRead > 0)
                 {
-                    var writtenBytes = offset + (ulong) bytesRead;
+                    var writtenBytes = offset + (ulong)bytesRead;
 
                     _sftpSession.RequestWrite(handle, offset, buffer, 0, bytesRead, null, s =>
+                    {
+                        if (s.StatusCode == StatusCodes.Ok)
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/SftpClient.cs
```diff
@@ -2123,26 +2358,26 @@ private void InternalUploadFile(Stream input, string path, Flags flags, SftpUplo
...
-                                }
+                                //  Execute callback on different thread
+                                ThreadAbstraction.ExecuteThread(() => uploadCallback(writtenBytes));
                             }
-                        });
+                        }
+                    });
                     Interlocked.Increment(ref expectedResponses);
 
-                    offset += (ulong) bytesRead;
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/SftpClient.cs
```diff
@@ -2202,6 +2437,13 @@ protected override void Dispose(bool disposing)
             }
         }
 
+        private void CheckDisposedOrNotConnected()
```
drieseng: Please remove this, and perform both checks separately like to we do for the sync methods.
Also please perform the dispose check before any argument checks and the (dis)connected check afterwards.

drieseng (changes requested):

src/Renci.SshNet/SftpClient.cs
```diff
@@ -138,7 +141,7 @@ public int ProtocolVersion
                 CheckDisposed();
                 if (_sftpSession == null)
                     throw new SshConnectionException("Client not connected.");
-                return (int) _sftpSession.ProtocolVersion;
+                return (int)_sftpSession.ProtocolVersion;
```
drieseng: Undo formatting change.

src/Renci.SshNet/SftpClient.cs
```diff
@@ -61,7 +64,7 @@ public TimeSpan OperationTimeout
                 if (timeoutInMilliseconds < -1d || timeoutInMilliseconds > int.MaxValue)
                     throw new ArgumentOutOfRangeException("value", "The timeout must represent a value between -1 and Int32.MaxValue, inclusive.");
 
-                _operationTimeout = (int) timeoutInMilliseconds;
+                _operationTimeout = (int)timeoutInMilliseconds;
```
drieseng: Undo formatting change.

src/Renci.SshNet/Renci.SshNet.csproj
vanillajonathan: Or maybe drop some target frameworks?
_"netstandard2.0 ought to be enough for anybody"_
[.NET Framework 4.5.2, 4.6, 4.6.1 will reach End of Support on April 26, 2022](https://devblogs.microsoft.com/dotnet/net-framework-4-5-2-4-6-4-6-1-will-reach-end-of-support-on-april-26-2022/)
Is there reason to add net46 when .NET Framework 4.6 is .NET Standard 1.3 compatible, and .NET Framework 4.6.1 is .NET Standard 2.0 compatible?
https://docs.microsoft.com/en-us/dotnet/standard/net-standard

drieseng: `While NuGet considers .NET Framework 4.6.1 as supporting .NET Standard 1.5 through 2.0, there are several issues with consuming .NET Standard libraries that were built for those versions from .NET Framework 4.6.1 projects. For .NET Framework projects that need to use such libraries, we recommend that you upgrade the project to target .NET Framework 4.7.2 or higher`

I have no details on what exactly is broken.

drieseng:
@IgorMilavec I finally had time to do a high-level review of this PR. Overall it's looking good, but of course (unsurprisingly) still incomplete. Thanks for taking this first (but important) step!!

I hope you're not discouraged by the number of comments.
A lot of them are due to (unintended?) formatting changes. These make the PR harder to review, pollute history and are not inline with the code style that we use.

I propose we discuss the comments in the PR, but - if useful/necessary - we can organize a (voice?) chat.

IgorMilavec:
No worries, @drieseng, thanks for the review.
Let me first clear trivial items off the table and then we'll discuss further, OK? I think voice chat would be great at that point.

src/Renci.SshNet/Renci.SshNet.csproj
IgorMilavec: net46 target is needed to use ```TaskCreationOptions.RunContinuationsAsynchronously```. 
@vanillajonathan  I have a couple of projects targeting net472 where I've had problems if the referenced libraries targeted net45+ AND netstandard 2.0. There is also a discussion about specific intricacies on dotnet repo, but I don't have the link at hand right now.
This is quite a tricky topic that I feel should be discussed further in #665 .

src/Renci.SshNet/Connection/ProxyConnector.cs
IgorMilavec: We need this in net35 target (```!FEATURE_SOCKET_DISPOSE``` case, specifically). I'll clean it up to make it evident.

src/Renci.SshNet/Connection/ProxyConnector.cs
IgorMilavec: Can do, but I'd rather do it in a separate PR, to keep things clean. This PR is getting large as it is...

src/Renci.SshNet/Session.cs
IgorMilavec: The current implementation of connection limiting does not play well in the async world and this would need to be completely reengineered if we decide to support this functionality in the async version. If you agree, then I'll remove this from ```ConnectAsync``` and document in the XmlDoc that the caller needs to take care of this. Maybe we should add a topic about this in the general documentation?

src/Renci.SshNet/Session.cs
IgorMilavec: Yes, this is something that we should definitely do, but I'd rather leave this for a later PR. This PR adds a new client surface area and I think we should first finish this as I think this is stable. Async message pump on the other hand is an implementation detail that can be changed later. And we need to be clear on the compilation targets that we want to support, because I have a feeling that supporting both sync (for pre-net40 targets) and async message pump in the codebase will be a royal PITA.

src/Renci.SshNet/Sftp/SftpFileStream.cs
IgorMilavec: Well, this one is between a rock and a hard place: Because this is inside a catch block I cannot await an async operation. Then maybe I could just start the async operation and not await the task, but this could lead to race conditions when the caller wants to dispose the session and the ```RequestCloseAsync``` hasn't completed yet.

src/Renci.SshNet/Sftp/SftpFileStream.cs
IgorMilavec: I'm not sure about that, any of the three could be wrong, depending on the caller's intention. Anyhow, this is the same as the sync version, so we either need to change both or none...

src/Renci.SshNet/SftpClient.cs
IgorMilavec: I created the new enum just to be semantically explicit and for better developer experience.
I think that the use of optimal buffer size is the reason why we would have this method in the first place. As a developer without any knowledge of how the library works, I would probably choose my "go to" number 4096, which would lead to suboptimal performance in this case.
If I disregard buffer size selection, I agree that users could easily implement this themselves, but then the same argument could be had for the sync version ```UploadFile```.

IgorMilavec:
I believe I have addressed all the trivial items. @drieseng can you please re-review the code and my comments?

.vs/ProjectSettings.json
```diff
@@ -0,0 +1,3 @@
+{
```
drieseng: Please remove this file.

.vs/VSWorkspaceState.json
```diff
@@ -0,0 +1,10 @@
+{
```
drieseng: Please remove this file.

src/Renci.SshNet/Renci.SshNet.csproj
drieseng: @IgorMilavec, is it ok for you to target .NET Framework 4.7.2 (or 4.8) instead?

src/Renci.SshNet/Sftp/SftpFileStream.cs
```diff
@@ -257,8 +283,8 @@ internal SftpFileStream(ISftpSession session, string path, FileMode mode, FileAc
             // that ensures we always receive or send the max. number of bytes in a single SSH_FXP_READ
             // or SSH_FXP_WRITE message
 
-            _readBufferSize = (int) session.CalculateOptimalReadLength((uint) bufferSize);
-            _writeBufferSize = (int) session.CalculateOptimalWriteLength((uint) bufferSize, _handle);
+            _readBufferSize = (int)session.CalculateOptimalReadLength((uint)bufferSize);
```
drieseng: Please undo these formatting changes.

src/Renci.SshNet/Sftp/SftpFileStream.cs
drieseng: If you update the **Renci.SshNet** project to use C# 6(.0) as language, you can use await in catch and finally blocks.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -635,6 +731,34 @@ public byte[] RequestRead(byte[] handle, ulong offset, uint length)
             return data;
         }
 
+#if FEATURE_TAP
+        public async Task<byte[]> ReadAsync(byte[] handle, ulong offset, uint length, CancellationToken cancellationToken)
```
drieseng: Please name to **RequestRequestAsync**.

src/Renci.SshNet/Sftp/SftpSession.cs
```diff
@@ -679,6 +803,33 @@ public byte[] RequestRead(byte[] handle, ulong offset, uint length)
             }
         }
 
+#if FEATURE_TAP
+        public async Task WriteAsync(byte[] handle, ulong serverOffset, byte[] data, int offset, int length, CancellationToken cancellationToken)
```
drieseng: Please rename to **RequestWriteAsync**.

src/Renci.SshNet/SftpClient.cs
drieseng: We could indeed say the same abount the sync version, but removing that one now would be a breaking change.
For now, I'd the remove the **DownloadFileAsync** and **UploadFileAsync** methods and the enum that you introduced.

src/Renci.SshNet/Abstractions/DnsAbstraction.cs
```diff
@@ -42,8 +45,6 @@ public static IPAddress[] GetHostAddresses(string hostNameOrAddress)
             if (!asyncResult.AsyncWaitHandle.WaitOne(Session.InfiniteTimeSpan))
                 throw new SshOperationTimeoutException("Timeout resolving host name.");
             return Dns.EndGetHostAddresses(asyncResult);
-#elif FEATURE_DNS_TAP
```
drieseng: This should be reverted, if not we'll use the IPAddress.TryParse(hostNameOrAddress, out address) implementation for .NET Standard 1.3 and UAP 10. It's not like we're always using the `DnsAbstrations.GetHostAddressesAsync()` method for those target frameworks that support TAP.

src/Renci.SshNet/Abstractions/SocketAbstraction.cs
```diff
@@ -59,15 +62,22 @@ public static void Connect(Socket socket, IPEndPoint remoteEndpoint, TimeSpan co
...
         private static void ConnectCore(Socket socket, IPEndPoint remoteEndpoint, TimeSpan connectTimeout, bool ownsSocket)
         {
 #if FEATURE_SOCKET_EAP
             var connectCompleted = new ManualResetEvent(false);
             var args = new SocketAsyncEventArgs
-                {
-                    UserToken = connectCompleted,
-                    RemoteEndPoint = remoteEndpoint
-                };
+            {
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Abstractions/SocketAbstraction.cs
```diff
@@ -97,7 +107,7 @@ private static void ConnectCore(Socket socket, IPEndPoint remoteEndpoint, TimeSp
 
             if (args.SocketError != SocketError.Success)
             {
-                var socketError = (int) args.SocketError;
+                var socketError = (int)args.SocketError;
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Abstractions/SocketAbstraction.cs
```diff
@@ -124,7 +134,7 @@ private static void ConnectCore(Socket socket, IPEndPoint remoteEndpoint, TimeSp
                 throw new SshOperationTimeoutException(string.Format(CultureInfo.InvariantCulture,
                     "Connection failed to establish within {0:F0} milliseconds.", connectTimeout.TotalMilliseconds));
 #else
-            #error Connecting to a remote endpoint is not implemented.
+#error Connecting to a remote endpoint is not implemented.
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Abstractions/SocketAbstraction.cs
```diff
@@ -144,7 +154,7 @@ public static void ClearReadBuffer(Socket socket)
         public static int ReadPartial(Socket socket, byte[] buffer, int offset, int size, TimeSpan timeout)
         {
 #if FEATURE_SOCKET_SYNC
-            socket.ReceiveTimeout = (int) timeout.TotalMilliseconds;
+            socket.ReceiveTimeout = (int)timeout.TotalMilliseconds;
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Abstractions/SocketAbstraction.cs
```diff
@@ -197,7 +207,7 @@ public static int ReadPartial(Socket socket, byte[] buffer, int offset, int size
                 receiveCompleted.Dispose();
             }
 #else
-            #error Receiving data from a Socket is not implemented.
+#error Receiving data from a Socket is not implemented.
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Abstractions/SocketAbstraction.cs
```diff
@@ -259,7 +269,7 @@ public static void ReadContinuous(Socket socket, byte[] buffer, int offset, int
             if (readToken.Exception != null)
                 throw readToken.Exception;
 #else
-            #error Receiving data from a Socket is not implemented.
+#error Receiving data from a Socket is not implemented.
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Abstractions/SocketAbstraction.cs
```diff
@@ -290,7 +300,7 @@ public static int ReadByte(Socket socket, TimeSpan timeout)
         /// <exception cref="SocketException">The write failed.</exception>
         public static void SendByte(Socket socket, byte value)
         {
-            var buffer = new[] {value};
+            var buffer = new[] { value };
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Abstractions/SocketAbstraction.cs
```diff
@@ -488,7 +505,7 @@ public static void Send(Socket socket, byte[] data, int offset, int size)
                 sendCompleted.Dispose();
             }
 #else
-            #error Sending data to a Socket is not implemented.
+#error Sending data to a Socket is not implemented.
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Abstractions/SocketAbstraction.cs
```diff
@@ -508,7 +525,7 @@ public static bool IsErrorResumable(SocketError socketError)
 #if FEATURE_SOCKET_EAP
         private static void ConnectCompleted(object sender, SocketAsyncEventArgs e)
         {
-            var eventWaitHandle = (ManualResetEvent) e.UserToken;
+            var eventWaitHandle = (ManualResetEvent)e.UserToken;
```
drieseng: Please undo formatting changes.

src/Renci.SshNet/Sftp/SftpFileStream.cs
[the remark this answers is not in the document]
drieseng: I can confirm that it's ok to just add **Flags.CreateNewOrOpen | Flags.Truncate** when mode is **FileMode.Create**.

drieseng (changes requested):

src/Renci.SshNet/Abstractions/DnsAbstraction.cs
IgorMilavec: Ups, I was too quick to remove and missed that the fallback was just IPAddress.TryParse. I have reverted the change, but I propose to remove FEATURE_DNS_TAP in the future from this method as we're using Sync/Async, which could cause problems.

src/Renci.SshNet/SftpClient.cs
drieseng: @vanillajonathan You forgot to thumb down this one. Thanks for the constructive feedback.

src/Renci.SshNet/Renci.SshNet.csproj
IgorMilavec: Sure, I've raised the target to net472.

src/Renci.SshNet/Abstractions/DnsAbstraction.cs
drieseng: I agree. These conditional compilation symbols are a necessary evil. The less we have to maintain, the better.

src/Renci.SshNet/Session.cs
IgorMilavec: I have changed the code and updated XmlDoc.
Can you please suggest other places where we should document this?

src/Renci.SshNet/SftpClient.cs
IgorMilavec: Then we need to remove FEATURE_TAP from netstandard1.3. Also, we take two "external" dependencies (AsyncInterfaces and Task.Extensions).
What if we leave this one as it is and add ```IAsyncEnumerable<SftpFile> EnumerateDirectoryAsync(...)``` for netstandard2.1+ ?

drieseng:
@IgorMilavec Thanks for hanging on. I'll do my best to finish the review some time this week.

src/Renci.SshNet/SftpClient.cs
drieseng: I won't let this block this merge request, but I still strongly consider changing this method to return **IAsyncEnumerable\<SftpFile>**. I also don't like bringing in additional dependencies, but the alternative is not better.

drieseng:
We're still lacking (unit & integration) test coverage, but I'm ok with adding these as a follow PR.
I'd also like to discuss further how to proceed with `ISftpClient.ListDirectoryAsync(...)`. In its current state, this is not a true async method.

That said: Thx !!!!!!!!

Dimencia:
Working on a large project at work and just ran into this being a problem, that's awesome that hours before I (realized I) needed it, there was already a PR merged

I'll be pulling this down and trying it tomorrow, nice work, and thanks

IgorMilavec:
Thanks @drieseng !
I'll start working on unit tests. After these are complete I can tackle the message pump and make it Task based as we already discussed.
Regarding ```ISftpClient.ListDirectoryAsync(...)```: the current version is already async, but it is not "streaming". For directories with small amount of files (up to the amount that ```SftpReadDirRequest``` returns) there is no difference. For a directory with large amount of files, the current implementation only completes (asynchronously) when all the file infos are read. The ```IAsyncEnumerable``` implementation would allow the caller to process file names in chunks (of the amount returned by one ```SftpReadDirRequest```) even before all the file infos are read back. Let me implement another function with IAsyncEnumerable and then we can discuss further, OK?

drieseng:
@IgorMilavec Thanks for taking this up. I'm ok with having a second method for ISftpClient.ListDirectoryAsync(...) as POC.

noahmehl:
When we review this PR, it looks like these methods are not actually implemented:

- DownloadFileAsync
- UploadFileAsync

Are we just mis-understanding the PR itself?

IgorMilavec:
These functions were initially implemented as part of this PR, but during the code review we decided that these functions are not really necessary, as they can be easily implemented in the consuming code with a couple of lines of code.

If you have no special requirements, you can use this extension class to get these two functions:
```csharp
namespace Renci.SshNet
{
    public static class SftpClientExtensions
    {
        public static async Task DownloadFileAsync(this SftpClient sftpClient, string path, Stream output, CancellationToken cancellationToken)
        {
            using (Stream remoteStream = await sftpClient.OpenAsync(path, FileMode.Open, FileAccess.Read, cancellationToken).ConfigureAwait(false))
            {
                await remoteStream.CopyToAsync(output, 81920, cancellationToken).ConfigureAwait(false);
            }
        }

        public static async Task UploadFileAsync(this SftpClient sftpClient, Stream input, string path, FileMode createMode, CancellationToken cancellationToken)
        {
            using (Stream remoteStream = await sftpClient.OpenAsync(path, createMode, FileAccess.Write, cancellationToken).ConfigureAwait(false))
            {
                await input.CopyToAsync(remoteStream, 81920, cancellationToken).ConfigureAwait(false);
            }
        }
    }
}
```

noahmehl:
@IgorMilavec Thank you so much for your quick and thoughtful reply.  We will take a look at this asap.

jjxtra:
Are these methods in official nuget? Not seeing any async methods on the sftpclient...

vanillajonathan:
I think the `DownloadFileAsync` and `UploadFileAsync` methods ought to be included, because I think this should be a async-first library where async should be the default behavior users use, and it should be easy to do so.

To be honest, I would like to see the non-async methods get removed as well as the legacy Begin* and End* async functions.

WojciechNagorski:

snargledorf:
Something that just bit me is that the OpenAsync method expects an absolute path while the Begin/End async methods for Upload/Download operate relative to the working directory.

Not sure what the best way to handle this would be, but it might be less likely to make a mistake if there were replacement methods on the SftpClient (DownloadAsync/UploadAsync) that behaved in the same way as the Begin/End methods (Relative to working directory), while OpenAsync remains absolute as to not break existing code.

MarkEnverus:
```csharp
namespace Renci.SshNet
{
    public static class SftpClientExtensions
    {
        public static async Task DownloadFileAsync(this SftpClient sftpClient, string path, Stream output, CancellationToken cancellationToken)
        {
            using (Stream remoteStream = await sftpClient.OpenAsync(path, FileMode.Open, FileAccess.Read, cancellationToken).ConfigureAwait(false))
            {
                await remoteStream.CopyToAsync(output, 81920, cancellationToken).ConfigureAwait(false);
            }
        }

        public static async Task UploadFileAsync(this SftpClient sftpClient, Stream input, string path, FileMode createMode, CancellationToken cancellationToken)
        {
            using (Stream remoteStream = await sftpClient.OpenAsync(path, createMode, FileAccess.Write, cancellationToken).ConfigureAwait(false))
            {
                await input.CopyToAsync(remoteStream, 81920, cancellationToken).ConfigureAwait(false);
            }
        }
    }
}
```
@IgorMilavec  the issue here is that we cannot moq this code since we can't create a SftpFileStream (as it's not an interface and it's private).  Whereas we can moq UploadFile (void return) unless OpenAsync returned a generic Stream instead of a wrapped Stream

PR #823, opened 2021-05-13
Fix avoidable exception when data length is too long

### LINKED ISSUE #713: "Data longer than 2147483647 is not supported" exception when connecting to Tectia ssh server running on a mainframe
This is similar to issue #41 but not exactly the same.
When connecting to Tectia ssh server running on a mainframe, I intermittently got an exception: Data longer than 2147483647 is not supported.
Tracing into the code I found that it looks like sometimes the first 4 bytes in the SSH_MSG_IGNORE message are filled with arbitrary numbers instead of the actual data length. When the number is greater than int.MaxValue the exception is thrown.
Changing IgnoreMessage.LoadData to the following fixed the issue:
```
        protected override void LoadData()
        {
            var dataLength = ReadUInt32();
            if (dataLength > (DataStream.Length - DataStream.Position))
            {
                DiagnosticAbstraction.Log("SSH_MSG_IGNORE: Length exceeds data bytes, data ignored.");
                Data = Array<byte>.Empty;
            }
            else if (dataLength > int.MaxValue)
            {
                throw new NotSupportedException(string.Format(CultureInfo.CurrentCulture, "Data longer than {0} is not supported.", int.MaxValue));
            }
            else
            {
                Data = ReadBytes((int) dataLength);
            }
        }
```

The server version says "SSH-2.0-4.3.6500.1 SSH Secure Shell".

### LINKED ISSUE #1243: Tectia SSH Server on z/OS mainframe
I have seen this mentioned before in the issues, but when connecting to a Tectia SSH server on an IBM mainframe, this error immediately occurs at connection:

`System.NotSupportedException: Data longer than 2147483647 is not supported.`

I would be happy to investigate patching the issue myself if some initial guidance could be provided.

EDIT:

Stacktrace is below
```
   at Renci.SshNet.Messages.Transport.IgnoreMessage.LoadData()
   at Renci.SshNet.Common.SshData.LoadInternal(Byte[] value, Int32 offset, Int32 count)
   at Renci.SshNet.Common.SshData.Load(Byte[] data, Int32 offset, Int32 count)
   at Renci.SshNet.Session.LoadMessage(Byte[] data, Int32 offset, Int32 count)
   at Renci.SshNet.Session.ReceiveMessage(Socket socket)
   at Renci.SshNet.Session.MessageListener()
--- End of stack trace from previous location ---
   at Renci.SshNet.Session.WaitOnHandle(WaitHandle waitHandle, TimeSpan timeout)
   at Renci.SshNet.Session.ConnectAsync(CancellationToken cancellationToken)
   at Renci.SshNet.BaseClient.CreateAndConnectSessionAsync(CancellationToken cancellationToken)
   at Renci.SshNet.BaseClient.ConnectAsync(CancellationToken cancellationToken)
   at UserQuery.Main(), line 17
```

se006:
Fix #713 
Fix #1243

Data lengths that are both longer than int.MaxValue and data bytes are now ignored and do not throw an exception.

This would come up when connecting to a Tectia SFTP server running on a mainframe.  

Credit to @edmlin for the fix.  This is for issue #713


IgorMilavec (changes requested):
You should change/remove test Load_ShouldThrowNotSupportedExceptionWhenDataLengthIsGreaterThanInt32MaxValue for this PR to pass.

se006:
I have removed the unit test as I cannot think of a way to enter into the middle condition, 
```
 else if (dataLength > int.MaxValue)
            {
                throw new NotSupportedException(string.Format(CultureInfo.CurrentCulture, "Data longer than {0} is not supported.", int.MaxValue));
            }
```
Perhaps this condition should be removed entirely.

IgorMilavec:
With the current implementation of SshData I believe you cannot test this case.
This exception provides a more explicit error information than the OverflowException that would happen anyway, and should stay in the code IMO.

IgorMilavec:
@se006 can you please edit the PR description (first comment) to include text "Fix #713" so this PR will be linked to the issue? Tnx!

se006:
Fix #713 added to first comment as requested

Rob-Hague (approved):

scott-xu (approved):

WojciechNagorski (approved):
LGTM!
I wish there was a test, but I know it will be difficult.

WojciechNagorski:

PR #855, opened 2021-08-04
FEAT: Add the ability to choose originating address for TCP connections

daviddenis-stx:
Would you be interested by this kind of change ?

The idea is to be able to bind a particular local network interface. 

It's for example useful when you have several routable interfaces on a given machine (ex: server with several failovers) and you want to be seen connecting from one of these interfaces (ex: the interface holding an IP that is white listed by the remote server you want to connect to). 

I already posted a similar change to fluentftp here https://github.com/robinrodricks/FluentFTP/commit/a761d05f982f8ad00a583ebb924002c6b515e395

User would only have to set the LocalIpAddress to something valid on its ConnectionInfo instance

Currently testing the change in real life

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,5 +1,7 @@
 The MIT License (MIT)
 
+Copyright (c) Renci, Oleg Kapeljushnik, Gert Driesen and contributors
+
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
 in the Software without restriction, including without limitation the rights


src/Renci.SshNet.Tests/Classes/Connection/SocketFactoryTests.cs
```diff
@@ -0,0 +1,48 @@
...
+        [TestMethod]
+        public void CreateOnSocketFactoryWithLocalIpAddressShouldHaveLocalEndpoint()
+        {
+            var socketFactory = new SocketFactory();
+
+            var socket = socketFactory.Create(
+                AddressFamily.InterNetwork,
+                SocketType.Stream,
+                ProtocolType.Tcp,
+                IPAddress.Loopback);
```
zybexXL: Other Tests are failing after this one runs because you're not Disposing the socket.
You can perhaps look at the DirectConnectorTest_Connect_* testcases and add this one to that collection, inheriting from DirectConnectorTestBase and overriding Teardown() function.

src/Renci.SshNet.Tests/Classes/Connection/SocketFactoryTests.cs
```diff
@@ -0,0 +1,48 @@
...
+                SocketType.Stream,
+                ProtocolType.Tcp, 
+                null);
+
+            Assert.IsNull(socket.LocalEndPoint);
+
+            socket = socketFactory.Create(
+                AddressFamily.InterNetwork,
+                SocketType.Stream,
+                ProtocolType.Tcp);
```
zybexXL, line 43: same, please Dispose this socket

src/Renci.SshNet/ConnectionInfo.cs
```diff
@@ -174,6 +176,12 @@ public class ConnectionInfo : IConnectionInfoInternal
         /// </value>
         public int RetryAttempts { get; set; }
 
+        /// <summary>
+        /// Gets or sets the local IP address to bind.
+        /// If not null, bind local socket to given address (use this if you have several interfaces and want to be seen as connecting from a particular one).
+        /// </summary>
+        public IPAddress LocalIpAddress { get; set; }
```
zybexXL: Why not `IPEndpoint LocalEndpoint` ? This way the user can optionally specify a local port as well, which might be needed for some strict firewalls.

src/Renci.SshNet.Tests/Classes/Connection/SocketFactoryTests.cs
zybexXL: Hmm, actually, I think tests are failing due to port exhaustion on the Appveyor VM - sockets remain in TIME_WAIT state for a long time. This is happening randomly on other PRs that have nothing to do with sockets.

To avoid this the SocketFactory code may need a flag to tell it to use SO_REUSEADDR on the sockets. Or if possible, reducing the TIME_WAIT delay or increasing the pool of available dynamic ports on the VM. 

This is outside of your PR.

jy824212: so whom should be the person to reconfigure the VM as you mentioned this is outside of the PR

zybexXL: @drieseng has access to it, not sure if anyone else has.

PR #860, opened 2021-08-15
Renci optimize session issue #861

SwarajKetan:
Due to improper scope AuthenticationConnection semaphore lock is acquired more than once when connect() method is called more than once while a connect operation is still in progress. It exhausts the semaphore slot and keeps the other Connect operations waiting until one finishes. Change of scoping will help us acquireclock just before the connect operation begins.

### DIFF
--- a/src/Renci.SshNet/Session.cs
+++ b/src/Renci.SshNet/Session.cs
@@ -565,19 +565,15 @@ namespace Renci.SshNet
             if (IsConnected)
                 return;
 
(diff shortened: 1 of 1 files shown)


A9G-Data-Droid:

Rob-Hague:
Thanks! Closing as the AuthenticationConnection semaphore has been removed in #1304

WojciechNagorski:

PR #863, opened 2021-08-25
Ignore ObjectDisposedException while stopping a forwarded port

StephenCWills:
I ran a code scan using SonarQube, and it raised alerts for these lines of code. Based on the comments inside the conditional block of code, the condition should also be checking for the `ForwardedPortStatus.Stopping` value as well.


WojciechNagorski:
I'm closing the PR, the fix is already implemented.

PR #864, opened 2021-08-27
Add support for resuming SFTP file upload/download

zybexXL:
This adds resume capability to `SftpClient.UploadFile()` , `SftpClient.DownloadFile()`, as well as to the `BeginUploadFile()` and `BeginDownloadFile()` variants.

The API remains unchanged. Resume is triggered simply by changing the current Position of the input/output stream which is given as an argument to these functions:
- Setting a non-zero position on the input Stream for UploadFile() triggers an Append to the remote file. **It's up to the caller to first check the current size of the remote file and set the local position accordingly!**
- Setting a non-zero position on the output Stream for DownloadFile() triggers a resume of the download at the given position.

This also does NOT check if the server supports Resume, though all servers I've tested so far do support it.


src/Renci.SshNet/Sftp/SftpFileReader.cs
```diff
@@ -53,12 +53,14 @@ internal class SftpFileReader : ISftpFileReader
...
         /// <param name="fileSize">The size of the file, if known; otherwise, <c>null</c>.</param>
-        public SftpFileReader(byte[] handle, ISftpSession sftpSession, uint chunkSize, int maxPendingReads, long? fileSize)
+        /// <param name="offset">The offset to resume from</param>
+        public SftpFileReader(byte[] handle, ISftpSession sftpSession, uint chunkSize, int maxPendingReads, long? fileSize, ulong offset = 0)
         {
             _handle = handle;
             _sftpSession = sftpSession;
             _chunkSize = chunkSize;
             _fileSize = fileSize;
+            _offset = _readAheadOffset = offset;
```
zybexXL: note: on download, set the initial offset of the remote file to the same offset as the local output file. Read-ahead also starts at the same offset

src/Renci.SshNet/ServiceFactory.cs
```diff
@@ -132,7 +132,7 @@ public ISftpFileReader CreateSftpFileReader(string fileName, ISftpSession sftpSe
             {
                 var fileAttributes = sftpSession.EndLStat(statAsyncResult);
                 fileSize = fileAttributes.Size;
-                maxPendingReads = Math.Min(10, (int) Math.Ceiling((double) fileAttributes.Size / chunkSize) + 1);
+                maxPendingReads = Math.Min(10, (int) Math.Ceiling((double)(fileSize - (long)offset) / chunkSize) + 1);
```
zybexXL: note: this calculates maxPendingReads based on remaining size, not total filesize

src/Renci.SshNet/SftpClient.cs
```diff
@@ -801,10 +801,15 @@ public void UploadFile(Stream input, string path, bool canOverride, Action<ulong
 
             var flags = Flags.Write | Flags.Truncate;
 
-            if (canOverride)
-                flags |= Flags.CreateNewOrOpen;
+            if (input.Position > 0)
+                flags = Flags.Write | Flags.Append;
```
zybexXL, line 985: note: on upload, if the local stream position is not zero, then open the remote file in APPEND mode

src/Renci.SshNet/SftpClient.cs
```diff
@@ -931,10 +936,15 @@ public IAsyncResult BeginUploadFile(Stream input, string path, bool canOverride,
...
```
zybexXL, line 1128: File upload: If the local stream position is not zero, then open the remote file in APPEND mode

src/Renci.SshNet/SftpClient.cs
```diff
@@ -2106,7 +2116,7 @@ private void InternalUploadFile(Stream input, string path, Flags flags, SftpUplo
 
             var handle = _sftpSession.RequestOpen(fullPath, flags);
 
-            ulong offset = 0;
+            ulong offset = (ulong)input.Position;
```
zybexXL: File upload: Set the initial offset of the remote file to the same as the local file

zybexXL:

zybexXL:
The single failed testcase does not relate to this PR, looks like a spurious failure in ForwardedPortRemoteTest_Start_PortNeverStarted.ForwardedPortShouldAcceptNewConnections()

WojciechNagorski (changes requested):
I checked this PR and it looks great!

I just want you to provide two integration tests for upload and download.
You can use similar mechanism, like in
https://github.com/sshnet/SSH.NET/blob/2eec748b184e6ef3dfec93649c1ca1bdc27a627a/test/Renci.SshNet.IntegrationTests/ConnectivityTests.cs#L199

```c#
bool vmNetworkConnectionDisabled = false;
SshConnectionRestorer disruptor = null;
try
{
    using (var client = new SftpClient(_connectionInfoFactory.Create()))
    {
        client.Connect();
        
        // <-- Start upload/download file

        disruptor = _sshConnectionDisruptor.BreakConnections();
        vmNetworkConnectionDisabled = true;

        WaitForConnectionInterruption(client);
        // disconnect while network connectivity is lost
        client.Disconnect();

        Assert.IsFalse(client.IsConnected);
        
        disruptor.RestoreConnections();
        vmNetworkConnectionDisabled = false;

        // connect when network connectivity is restored
        client.Connect();

        // <-- Resume download/upload file

        client.Dispose();
    }
}
finally
{
    if (vmNetworkConnectionDisabled)
    {
        disruptor.RestoreConnections();
    }
    disruptor?.Dispose();
}
```

zybexXL:
@WojciechNagorski 
This type of Resume test won't work as the test file may well finish upload/download before the Disruptor kicks in, especially if the test server is in localhost.
Better to upload a small file, then reconnect and append some more data, then check if the filesize is as expected. The download resume test can download the file in 2 chunks and check if it's correct. OK ?

WojciechNagorski:
Sorry for delay. It's correct.

PR #865, opened 2021-08-30
Use hardware-accelerated AES CryptoServiceProvider

zybexXL:
This feature reduces CPU usage dramatically, allowing higher SFTP performance on slower machines. It does so by using the AesCryptoServiceProvider for AES CBC/CTR/ECB encryption/decryption, which taps into AES-NI accelerated OS functions.

### STFP download performance testcases:
**Test case A: 4 year-old i5-7300U 2-core/4-threads laptop, 250Mbps internet connection:**
- Before: 13.15MB/sec, one CPU thread fully saturated
- After: 24.50 MB/sec (+86%), limited by internet connection, main thread at around 20-30% (5-10% in Task Manager)

**Test case B: 4 year old I7-7700 4-core/8-threads, 500Mbps internet connection:**
- Before: 40.85 MB/sec, one CPU-thread fully saturated
- After: 66.70 MB/sec (+63%), limited by internet connection, main thread at around 30-40% (5-10% in Task Manager)

_Note: all values above were taken with PR #866 also merged in. Tested with sftpClient.DownloadFile()_

### SSH.NET vs AES-NI benchmarks
SSH.NET implements AES as pure managed code. CPUs nowadays have hardware acceleration for AES via the AES-NI instruction set, and .NET provides APIs to the OS functions that make use of it. There are 3 .NET providers for AES:

AesManaged : older managed-only implementation, not accelerated. Still faster than the current SSH.Net
AesCryptoServiceProvider: call into the OS and makes use of AES-NI if available
AesCNG (CryptoNewGen): newer Crypto API, available since 2018 in Win10. Also makes use of AES-NI and has less overhead

Here are some benchmarks (executed on the dual-core laptop...):
```
Provider    Mode  Iterations                                  Average
----------  ----  ---------------------------------------     ------------
SSH.NET     CTR     29.44   29.75   29.41   29.81   28.69  =>   29.42 MB/s
SSH.NET     CBC     24.72   28.47   28.34   28.59   28.81  =>   27.79 MB/s
SSH.NET     ECB     33.75   33.50   25.34   27.56   23.53  =>   28.74 MB/s
SSH.NET     CFB     27.81   26.41   27.41   27.69   27.56  =>   27.38 MB/s
SSH.NET     OFB     27.97   28.06   27.75   27.69   27.75  =>   27.84 MB/s
           
AesManaged  CBC     41.19   41.50   41.72   41.78   25.66  =>   38.37 MB/s
AesCSP      CBC    717.06  777.25  828.03  815.72  812.75  =>  790.16 MB/s   (AES-NI)
AesCNG      CBC    830.16  847.25  852.72  853.06  832.78  =>  843.19 MB/s   (AES-NI)
```

As you can see, the accelerated APIs are about 25x faster than the current managed code!

Since AesCNG is too new I chose to use AesCryptoServiceProvider. This supports CBC and ECB natively, but not CFB or OFB (which are not listed in the SSH.Net protocol list anyway). For CTR (used by AWS SFTP Transfer Family), my code uses ECB followed by XOR. This would be extremely fast on C/C++ (or using unsafe tag in C# to allow memory pointers), but on pure C# I had to do a few memory Blockcopy to convert between byte[] and uint[], so the result is not ideal - however, it's still about 10x faster than the previous code.

Here's the new AES benchmarks with this PR:
```
Provider    Mode  Iterations                                  Average
----------  ----  ---------------------------------------     ------------
SSH.NET+    CTR    316.53  317.69  315.72  305.00  308.53  =>  312.69 MB/s (1063%)
SSH.NET+    CBC    768.03  744.16  772.19  781.16  761.41  =>  765.39 MB/s (2754%)
SSH.NET+    ECB   2368.78 2370.50 2393.44 2368.03 2344.53  => 2369.06 MB/s (8243%)
SSH.NET+    CFB     27.31   27.78   26.91   27.81   26.66  =>   27.29 MB/s (100%)
SSH.NET+    OFB     28.41   27.91   28.25   28.38   27.97  =>   28.18 MB/s (101%)
```

These CTR and CBC values are now the new maximum theoretical SFTP performance of SSH.NET 😎


src/Renci.SshNet/Security/Cryptography/Ciphers/Modes/CtrCipherMode.cs
```diff
@@ -55,39 +73,65 @@ public override int EncryptBlock(byte[] inputBuffer, int inputOffset, int inputC
...
         /// </summary>
-        /// <param name="inputBuffer">The input data to decrypt.</param>
-        /// <param name="inputOffset">The offset into the input byte array from which to begin using data.</param>
-        /// <param name="inputCount">The number of bytes in the input byte array to use as data.</param>
-        /// <param name="outputBuffer">The output to which to write decrypted data.</param>
-        /// <param name="outputOffset">The offset into the output byte array from which to begin writing data.</param>
-        /// <returns>
-        /// The number of bytes decrypted.
-        /// </returns>
-        public override int DecryptBlock(byte[] inputBuffer, int inputOffset, int inputCount, byte[] outputBuffer, int outputOffset)
```
zybexXL, line 69: This is an exact copy of EncryptBlock() - the parent class implements it by default, no need to override it on this CipherMode

src/Renci.SshNet/Security/Cryptography/Ciphers/Modes/OfbCipherMode.cs
```diff
@@ -54,40 +60,5 @@ public override int EncryptBlock(byte[] inputBuffer, int inputOffset, int inputC
...
-        /// </summary>
-        /// <param name="inputBuffer">The input data to decrypt.</param>
-        /// <param name="inputOffset">The offset into the input byte array from which to begin using data.</param>
-        /// <param name="inputCount">The number of bytes in the input byte array to use as data.</param>
-        /// <param name="outputBuffer">The output to which to write decrypted data.</param>
-        /// <param name="outputOffset">The offset into the output byte array from which to begin writing data.</param>
-        /// <returns>
-        /// The number of bytes decrypted.
-        /// </returns>
-        public override int DecryptBlock(byte[] inputBuffer, int inputOffset, int inputCount, byte[] outputBuffer, int outputOffset)
```
zybexXL, line 69: This is an exact copy of EncryptBlock() - the parent class implements it by default, no need to override it on this CipherMode

jjxtra:
Well that's pretty good!

IgorMilavec:
```CTRArrayXOR``` can be optimized on supported platforms with the use of ```Vector<T>```. It's faster and allocation free:
|   Method | Size |      Mean |     Error |    StdDev |  Gen 0 | Allocated |
|--------- |----- |----------:|----------:|----------:|-------:|----------:|
| Original |  128 | 123.07 ns |  2.455 ns |  3.823 ns | 0.0968 |     304 B |
|   Vector |  128 |  21.87 ns |  0.101 ns |  0.090 ns |      - |         - |
| Original |  256 | 184.98 ns |  1.866 ns |  1.745 ns | 0.1783 |     560 B |
|   Vector |  256 |  38.55 ns |  0.096 ns |  0.085 ns |      - |         - |
| Original |  512 | 349.29 ns | 10.298 ns | 30.363 ns | 0.3414 |   1,072 B |
|   Vector |  512 |  81.09 ns |  0.251 ns |  0.235 ns |      - |         - |

Here is the code I used:
```` c#
byte[] CTRArrayXOR(byte[] counter, byte[] data, int offset, int length)
{
    for (int loopOffset = 0; length > 0; length -= Vector<byte>.Count)
    {
        var v = new Vector<byte>(counter, loopOffset) ^ new Vector<byte>(data, offset + loopOffset);
        if (length >= Vector<byte>.Count)
        {
            v.CopyTo(counter, loopOffset);
            loopOffset += Vector<byte>.Count;
        }
        else
        {
            for (int i = 0; i < length; i++)
            {
                counter[loopOffset++] = v[i];
            }
        }
    }
    return counter;
}
````

**Comment by zybexXL** [comment]
I've tested your snippet by replacing the XOR on my AES-CTR benchmark, which processes chunks of 32KB of random data (similar to what happens when downloading a file)
My current code: ~300MB/sec
Your code: ~150MB/sec

Note that my code copies the byte[] to uint[] so that the XOR is not done byte by byte. Does your benchmark take that into account? Is the "original" my PR code (accelerated), or the existing SSH-NET code?

~~You mention that vector\<T\> doesn't need to allocate memory... I think that's not exactly true. That FOR loop instantiates 2 Vector\<byte\> objects, each holding 4 bytes. Due to the limited scope the objects are immediately discarded after each loop. So it's still allocating 16384 objects in the stack in that FOR loop to XOR the two 32KB arrays. Instantiating objects is not free.~~

I see that vector\<T\> is likely optimized by the compiler to use CPU registers. Even so, it seems slower than my current code.

**Comment by zybexXL** [comment]
I recompiled the benchmark for NetCore 5.0 and added 3 versions of CTR:
- BlockCopy (current code in this PR, copying byte[] to uint[] before XOR)
- Vector\<T\> (IgorMilavec's suggestion)
- Span\<T\> (allowing cast from byte[] to uint[] without memcopy)

Running it as Release instead of Debug also kicked performance by a lot, and changes the ranking. So here are the results:

|Benchmark            |Average
|---------------------|-----------
|SSH.NET CTR 2020.0.1| 49.06 MB/s
|SSH.NET CTR BlockCopy|586.66 MB/s
|SSH.NET CTR Span\<T\>|671.85 MB/s
|SSH.NET CTR Vector\<T\>|731.97 MB/s
|AesCSP CBC |785.79 MB/s

So Vector is indeed the fastest, but available only since NetStandard 2.1 (like Span). Blockcopy is not that bad considering it's available for all platforms, and it's still 13x faster than base. We can add a separate PR with Vector/Span later on if needed.

Here's the snippet for SpanXOR - it assumes the arrays are 4-byte aligned for simplification:

```cs
        private byte[] SpanXOR(byte[] data, int offset, byte[] output)
        {
            int uOffset = offset / 4;
            Span<uint> uData = MemoryMarshal.Cast<byte, uint>(data);
            Span<uint> uOut = MemoryMarshal.Cast<byte, uint>(output);
            for (int i = 0; i < uOut.Length; i++)
                uOut[i] = uOut[i] ^ uData[i + uOffset];
            return output;
        }
```

IgorMilavec (approved):
@drieseng This looks fine. I've been using it in production for a couple of months without any side effects. With 10x performance improvement, I propose to merge this before the next release.

A9G-Data-Droid:

zybexXL:
I have been using this code since 2021 and I still see the huge performance gains this provides, together with #866

Rob-Hague:
I'm very much in favour of this change, but I think the design needs some work. In fact, I think it can be much simpler as a starting point: I think we can just replace the implementation of `AesCipher` with an `ICryptoTransform` in ECB mode and no padding.

That way, we still use the given `CipherMode` object (rather than subverting it) which is more predictable, testable and yet we should still get the majority of the performance gains. The `CipherMode`s can easily be optimized as a next step, and then further down the line (if we are still not satisfied) we can think about increasing the complexity in order to delegate to the BCL for the cipher modes too.

If you have any thoughts on this approach, or would like to work on it, let me know. Otherwise, I will play with it.

zybexXL:
All AES tests pass (including the new ones from #1232). The failed test is unrelated, seems spurious:

```
Failed CanWriteShouldReturnFalse [9 ms]
  Error Message:
   Initialization method Renci.SshNet.Tests.Classes.Sftp.SftpFileStreamTest_CanWrite_Closed_FileAccessReadWrite.SetUp threw exception. System.ArgumentOutOfRangeException: Cannot be less than or equal to zero. (Parameter 'bufferSize').
```

Is there a way to force re-run of the test suite without doing a dummy commit?

WojciechNagorski:
I've merged #1232 can you update this PR and check if all tests passed?

zybexXL:
DST messed up the order of comments, my comment above was actually made after Wojciech's.
AES tests passed.

zybexXL:
> I'm very much in favour of this change, but I think the design needs some work. In fact, I think it can be much simpler as a starting point: I think we can just replace the implementation of `AesCipher` with an `ICryptoTransform` in ECB mode and no padding.
> 
> That way, we still use the given `CipherMode` object (rather than subverting it) which is more predictable, testable and yet we should still get the majority of the performance gains. The `CipherMode`s can easily be optimized as a next step, and then further down the line (if we are still not satisfied) we can think about increasing the complexity in order to delegate to the BCL for the cipher modes too.
> 
> If you have any thoughts on this approach, or would like to work on it, let me know. Otherwise, I will play with it.

@Rob-Hague 
ECB by itself is not enough. My main goal with this code was to improve performance with AWS SFTP, which uses CTR.
When I added this PR 3 years ago, SSH.Net still included many legacy frameworks which did not support the accelerated AESCryptoProvider API, so I left the legacy/slow ECB code in place and surrounded the new code with the FEATURE_AES_CSP conditional.  

With the removal of the legacy frameworks perhaps we can get rid of the legacy code and just use this new one. However, I think we should first merge this in, and then rework and remove the legacy code in a separate PR.

Rob-Hague:
> With the removal of the legacy frameworks perhaps we can get rid of the legacy code and just use this new one. However, I think we should first merge this in, and then rework and remove the legacy code in a separate PR.

Let's get it right in this PR.

My biggest concern is the subversion of logic. When passing a `CipherMode` object, my expectation is that this object is going to be used in the cipher process (as it always has done). But with this change, the object is only used as a type check and the implementation is not touched. This is unexpected.

I have run some experiments with the following changes to the benchmarks:

<details>

```diff

diff --git a/test/Renci.SshNet.Benchmarks/Security/Cryptography/Ciphers/AesCipherBenchmarks.cs b/test/Renci.SshNet.Benchmarks/Security/Cryptography/Ciphers/AesCipherBenchmarks.cs
index ff414cc4..b91e3b8a 100644
--- a/test/Renci.SshNet.Benchmarks/Security/Cryptography/Ciphers/AesCipherBenchmarks.cs
+++ b/test/Renci.SshNet.Benchmarks/Security/Cryptography/Ciphers/AesCipherBenchmarks.cs
@@ -15,7 +15,7 @@ namespace Renci.SshNet.Benchmarks.Security.Cryptography.Ciphers
         {
             _key = new byte[32];
             _iv = new byte[16];
-            _data = new byte[256];
+            _data = new byte[32 * 1024];
 
             Random random = new(Seed: 12345);
             random.NextBytes(_key);
@@ -34,5 +34,29 @@ namespace Renci.SshNet.Benchmarks.Security.Cryptography.Ciphers
         {
             return new AesCipher(_key, new CbcCipherMode(_iv), null).Decrypt(_data);
         }
+
+        [Benchmark]
+        public byte[] Encrypt_CFB()
+        {
+            return new AesCipher(_key, new CfbCipherMode(_iv), null).Encrypt(_data);
+        }
+
+        [Benchmark]
+        public byte[] Decrypt_CFB()
+        {
+            return new AesCipher(_key, new CfbCipherMode(_iv), null).Decrypt(_data);
+        }
+
+        [Benchmark]
+        public byte[] Encrypt_CTR()
+        {
+            return new AesCipher(_key, new CtrCipherMode(_iv), null).Encrypt(_data);
+        }
+
+        [Benchmark]
+        public byte[] Decrypt_CTR()
+        {
+            return new AesCipher(_key, new CtrCipherMode(_iv), null).Decrypt(_data);
+        }
     }
 }

```
</details>

Results on develop branch 826222f6fe33450754b1bb:

| Method      | Mean     | Error   | StdDev  | Gen0    | Allocated |
|------------ |---------:|--------:|--------:|--------:|----------:|
| Encrypt_CBC | 320.4 μs | 0.47 μs | 0.36 μs | 15.6250 |  32.41 KB |
| Decrypt_CBC | 352.3 μs | 2.37 μs | 1.85 μs | 15.6250 |  32.41 KB |
| Encrypt_CFB | 335.6 μs | 0.43 μs | 0.34 μs | 15.6250 |  32.45 KB |
| Decrypt_CFB | 367.9 μs | 4.23 μs | 3.95 μs | 15.6250 |  32.45 KB |
| Encrypt_CTR | 310.4 μs | 0.38 μs | 0.31 μs | 15.6250 |  32.45 KB |
| Decrypt_CTR | 315.3 μs | 3.96 μs | 3.70 μs | 15.6250 |  32.45 KB |


Results on my suggestion to replace the `AesCipher` implementation with the BCL https://github.com/Rob-Hague/SSH.NET/commit/8619062975f4a3e502397cc1c9781f995c2ace62

| Method      | Mean     | Error   | StdDev  | Gen0    | Allocated |
|------------ |---------:|--------:|--------:|--------:|----------:|
| Encrypt_CBC | 221.2 μs | 2.01 μs | 1.88 μs | 15.8691 |   32.7 KB |
| Decrypt_CBC | 220.0 μs | 2.73 μs | 2.56 μs | 15.8691 |  32.71 KB |
| Encrypt_CFB | 251.3 μs | 0.33 μs | 0.27 μs | 15.6250 |  32.75 KB |
| Decrypt_CFB | 255.3 μs | 0.67 μs | 0.60 μs | 15.6250 |  32.75 KB |
| Encrypt_CTR | 201.3 μs | 0.41 μs | 0.38 μs | 15.8691 |  32.75 KB |
| Decrypt_CTR | 209.3 μs | 0.22 μs | 0.19 μs | 15.8691 |  32.75 KB |


Results on your branch https://github.com/zybexXL/SSH.NET/commit/a9f68fbed70b94728869202c44f69869f2eb7693

| Method      | Mean       | Error     | StdDev    | Gen0    | Allocated |
|------------ |-----------:|----------:|----------:|--------:|----------:|
| Encrypt_CBC |  29.637 μs | 0.1278 μs | 0.1067 μs | 16.1133 |  33.14 KB |
| Decrypt_CBC |   6.579 μs | 0.0160 μs | 0.0142 μs | 16.1285 |  33.14 KB |
| Encrypt_CFB | 333.761 μs | 0.1710 μs | 0.1335 μs | 15.6250 |  32.52 KB |
| Decrypt_CFB | 333.899 μs | 0.2090 μs | 0.1955 μs | 15.6250 |  32.52 KB |
| Encrypt_CTR |  24.630 μs | 0.0327 μs | 0.0273 μs | 78.1250 |  161.2 KB |
| Decrypt_CTR |  24.544 μs | 0.0817 μs | 0.0683 μs | 78.1250 |  161.2 KB |


So my suggestion gets moderate gains but clearly we can do a lot better with an approach to delegate the entire encryption to the BCL (rather than block-by-block).

So here is what I propose:

We keep the behaviour of the existing constructor on `AesCipher` as is. That is, if you pass a `CipherMode` and/or `CipherPadding` object then they will be used as expected, with encryption happening block-by-block using the BCL as in my branch.

We add a new constructor which will allow delegating to the BCL for the entire encryption process. This will achieve the much better performance *and* separation from the existing behaviour.

```csharp

public AesCipher(
    System.Security.Cryptography.CipherMode cipherMode,
    System.Security.Cryptography.PaddingMode paddingMode,
    bool ctrMode = false)
{ }

```

Since `S.S.C.CipherMode` does not have a `CTR` value, there is a `ctrMode` parameter. The constructor would throw if `ctrMode` is true and `cipherMode` != ECB. A bit awkward, but it should work fine. Other ideas would be to define our own `CipherMode` enum which does contain a `CTR` value, or to allow passing e.g. `(S.S.C.CipherMode)-1` which we interpret as CTR.

In terms of the design, we define nested classes in `AesCipher` which represent the different implementations. Essentially, `AesCipher` becomes:

```c#

public sealed class AesCipher : BlockCipher
{
    private readonly BlockCipher _impl;

    public AesCipher(byte[] key, CipherMode mode, CipherPadding padding)
        : base(key, 16, mode, padding)
    {
        _impl = new SshNetCipherModeImpl(key, mode, padding);
    }

    public AesCipher(
        byte[] key,
        byte[] iv,
        System.Security.Cryptography.CipherMode cipherMode,
        System.Security.Cryptography.PaddingMode paddingMode,
        bool ctrMode = false)
    {
        _impl = new BclImpl(/* ... */);
    }

    // AesCipher just forwards all implementation to _impl

    public override int EncryptBlock(byte[] inputBuffer, int inputOffset, int inputCount, byte[] outputBuffer, int outputOffset)
    {
        return _impl.EncryptBlock(inputBuffer, inputOffset, inputCount, outputBuffer, outputOffset);
    }

    public override int DecryptBlock(byte[] inputBuffer, int inputOffset, int inputCount, byte[] outputBuffer, int outputOffset)
    {
        return _impl.DecryptBlock(inputBuffer, inputOffset, inputCount, outputBuffer, outputOffset);
    }

    public override byte[] Encrypt(byte[] input, int offset, int length)
    {
        return _impl.Encrypt(input, offset, length);
    }

    public override byte[] Decrypt(byte[] input)
    {
        return _impl.Decrypt(input);
    }

    public override byte[] Decrypt(byte[] input, int offset, int length)
    {
        return _impl.Decrypt(input, offset, length);
    }

    // Implementations

    // The block-by-block implementation using instantiated SshNet.CipherMode, SshNet.CipherPadding objects
    private sealed class SshNetCipherModeImpl : BlockCipher
    {
        private readonly Aes _aes;
        private ICryptoTransform _encryptor;
        private ICryptoTransform _decryptor;

        public SshNetCipherModeImpl(byte[] key, CipherMode mode, CipherPadding padding)
            : base(key, 16, mode, padding)
        {
            // Initialise _aes in ECB mode
        }

        public override int EncryptBlock(byte[] inputBuffer, int inputOffset, int inputCount, byte[] outputBuffer, int outputOffset)
        {
            _encryptor ??= _aes.CreateEncryptor();

            return _encryptor.TransformBlock(inputBuffer, inputOffset, inputCount, outputBuffer, outputOffset);
        }

        public override int DecryptBlock(byte[] inputBuffer, int inputOffset, int inputCount, byte[] outputBuffer, int outputOffset)
        {
            _decryptor ??= _aes.CreateDecryptor();

            return _decryptor.TransformBlock(inputBuffer, inputOffset, inputCount, outputBuffer, outputOffset);
        }
    }

    private sealed class BclImpl : BlockCipher
    {
        // This overrides Encrypt/Decrypt for full BCL perf.

        public override byte[] Decrypt(byte[] input)
        {
            throw new NotImplementedException();
        }

        public override byte[] Decrypt(byte[] input, int offset, int length)
        {
            throw new NotImplementedException();
        }

        public override int DecryptBlock(byte[] inputBuffer, int inputOffset, int inputCount, byte[] outputBuffer, int outputOffset)
        {
            throw new NotImplementedException();
        }

        public override byte[] Encrypt(byte[] input, int offset, int length)
        {
            throw new NotImplementedException();
        }

        public override int EncryptBlock(byte[] inputBuffer, int inputOffset, int inputCount, byte[] outputBuffer, int outputOffset)
        {
            throw new NotImplementedException();
        }
    }
}

```

Comments appreciated. Most important to me is that we do not subvert provided `SshNet.CipherMode` instances.

Rob-Hague:
> Most important to me is that we do not subvert provided SshNet.CipherMode instances.

One other thought: because the proposed new constructor is a bit awkward for CTR mode, I am in principle OK with subverting our own `SshNet.CipherMode` instances with an exact type check e.g. `mode.GetType() == typeof(CtrCipherMode)` instead of an `is` check (which would pass also for derived types which we definitely should not subvert). Then we can stick with one constructor which decides on the implementation, and the awkward logic is not exposed publicly.

zybexXL:
@Rob-Hague Please check my latest rework implementing most of your suggestions:

- All AES modes now benefit from CNG acceleration using the BCL (not just CTR and CBC as before)
- _System.Security.Cryptography.CipherMode_ and derived classes now all use the Aes BCL implementation
- The _CipherMode_ class no longer needs a reference to the _BlockCipher_ object. This means it can now be instantiated on its own if needed.
- _System.Security.Cryptography.AesCipher_ cleaned up, always uses one of the CipherModes for actual work.
- Added an _EcbCipherMode_ class, used as default when AesCipher is instantiated without a specific mode (same behavior as before this PR).
- 
All tests from your new AesCipherTest suite are passing. Thank you for those, they are excellent for debugging :)

Performance of this commit on my laptop:

| Method      | Mean      | Error     | StdDev    | Gen0    | Allocated |
|------------ |----------:|----------:|----------:|--------:|----------:|
| Encrypt_CBC | 31.400 us | 0.5509 us | 0.4884 us |  3.9673 |  32.78 KB |
| Decrypt_CBC |  7.141 us | 0.1381 us | 0.1890 us |  3.9978 |  32.79 KB |
| Encrypt_CFB | 48.552 us | 0.9110 us | 0.8522 us |  3.9673 |  32.78 KB |
| Decrypt_CFB | 42.692 us | 0.8191 us | 1.0060 us |  3.9673 |  32.79 KB |
| Encrypt_CTR | 25.344 us | 0.4998 us | 1.4420 us | 19.5923 | 160.88 KB |
| Decrypt_CTR | 26.970 us | 0.5389 us | 1.2491 us | 19.5923 | 160.88 KB |
| Encrypt_ECB |  6.645 us | 0.1257 us | 0.1176 us |  3.9978 |  32.78 KB |
| Decrypt_ECB | 11.662 us | 0.2306 us | 0.4443 us |  3.9978 |  32.79 KB |

zybexXL:
Doh! This change breaks 3DES and Blowfish because they use CBC with 64 bit blocksize, not supported by the BCL.
I'll fix it tomorrow, it should not impact the AesCipher performance when blocksize=128.

Rob-Hague:
Indeed, I do not think any changes are necessary in `BlockCipher`, `Cipher` or `CipherMode`. We should be able to keep all changes constrained to `AesCipher` (with the occasional exception like changing `CipherMode.IV` to `protected internal`).

We should be able to just have the suggested `AesCipher.SshNetCipherModeImpl` whose implementation is as in #1235, and `AesCipher.BclImpl` whose implementation is basically your `AesCipherCSP` (but without the SSH.NET `CipherMode` and `CipherPadding` classes).

I would recommend trying to achieve that without the other changes. Of course, I could have missed something when sketching it out which would require additional changes.

zybexXL:
_EDIT: I thinking what we're saying is not so different - make AesCipher use the BCL and ignore the CipherModes just on Aes. That should work, but improving the existing CipherModes would also lift other non-AES ciphers that use them._

-- original comment:
That was my previous approach, if you [look at it](5cc5d4a72096b3a083e7c20831e82579cb20ee5f). It's just not possible to leave those classes untouched and get the 10x performance on all modes - just compare your numbers with my numbers. When you start trying to get that performance only by changing AesCipher, you hit some limits of the current design - try it. Also, not all AES modes are supported by the different frameworks, there needs to be some #IF blocks.

Adding Aes to CipherModes lifts all boats, not just AES itself. Other ciphers also use those modes, though of course AES is the most relevant.

One solution is to get rid of the custom CipherModes (or not use them for AES) and just use the BCL enumeration, with CTR being a special case. The CipherPadding is also kind of useless, as PKCS5 is the same as PKCS7, so there's only one Padding implementation - it could just be a boolean.

I'll work on fixing this today/tomorrow while keeping the same performance levels.

Rob-Hague:
> One solution is to get rid of the custom CipherModes and just use the BCL enumeration (with CTR being a special case).

Yes, that is my suggestion with the new constructor which takes [`System.Security.Cryptography.CipherMode`](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.ciphermode) and which avoids the `EncryptBlock` loop in `BlockCipher.Encrypt`. Note, that is the BCL enum type, not the `Renci.SshNet.Security.Cryptography.CipherMode` class defined in SSH.NET... it is confusing.

 ```cs
     public AesCipher(
         byte[] key,
         byte[] iv,
         System.Security.Cryptography.CipherMode cipherMode,
         System.Security.Cryptography.PaddingMode paddingMode,
         bool ctrMode = false) // Or some magic value for cipherMode parameter
     {
         _impl = new BclImpl(/* ... */); // This is your AesCipherCSP
     }
 ```

However, the existing constructor taking an SSH.NET `CipherMode` should still go through the `EncryptBlock` loop in case someone has provided a derived `CipherMode` type (and because it is more obvious). This is covered by `SshNetCipherModeImpl` in the proposal.

> make AesCipher use the BCL and ignore the CipherModes just on Aes

To reiterate, we should not ignore an SSH.NET `CipherMode` object when we have been given one, hence my suggestion for the new constructor.

> That should work, but improving the existing CipherModes would also lift other non-AES ciphers that use them.

Agreed but this is less important (than keeping this change as simple as possible)

> Also, not all AES modes are supported by the different frameworks, there needs to be some #IF blocks.

OFB is not but the rest are I think? It's fine to let the BCL throw an error if we take the new constructor approach. OFB is not important.

zybexXL:
CFB is not available on NetStandard 2.1, but is on Net6/7
OFB is only available in Net462 and NetStandard 2.0
CTR is not in the BCL, so it likely needs a separate constructor

Rob-Hague:
BTW, I am aware that I could be being totally unclear in my suggested implementation. I am happy to clear things up in any way. I would also be happy to submit a PR to your PR to get it in the suggested shape.

zybexXL:
@Rob-Hague 
Alright, I moved all code into `AesCipher.cs` as you suggested. Some comments:
- I added an `AesCipherMode` enum. Let me know if it should be moved into a different file
- The optimized CTR code is also in `AesCipher.cs`, though it would perhaps be cleaner to add an `AesCtrCipher.cs` that would inherit from AesCipher and contain the CTR-specific code. However, it would be strange to have this for CTR and not the others
- Further improved CTR by using the `Vector<byte>` suggestion from @IgorMilavec (with fixes) and reducing the memory allocation (can't be further reduced, minimum is 2x input buffer size for this optimization)
- The existing PaddingMode implementation is broken - I will open an Issue to track it. I didn't change it (except on the new Aes implementation) because some existing testcases actually look for the bad padding. I did have to disable padding [here](https://github.com/sshnet/SSH.NET/pull/865/files#diff-5ff67ce9967b2cc4fc2bbeeb0668079c1f46873bdb011ce12ea67bc81c112dbcR525) for those test cases to pass with the new code.

**Performance on my machine:**

Results on develop branch https://github.com/sshnet/SSH.NET/commit/826222f6fe33450754b1bb58c6a8e033e912b5ff:
| Method      | Mean     | Error   | StdDev  | Gen0   | Allocated |
|------------ |---------:|--------:|--------:|-------:|----------:|
| Encrypt_CBC | 298.7 us | 5.77 us | 5.66 us | 3.9063 |  32.41 KB |
| Decrypt_CBC | 299.9 us | 4.14 us | 3.45 us | 3.9063 |  32.41 KB |
| Encrypt_CFB | 312.6 us | 6.18 us | 6.61 us | 3.9063 |  32.45 KB |
| Decrypt_CFB | 316.8 us | 5.53 us | 4.90 us | 3.9063 |  32.45 KB |
| Encrypt_CTR | 296.6 us | 6.00 us | 5.61 us | 3.9063 |  32.45 KB |
| Decrypt_CTR | 298.6 us | 4.21 us | 3.94 us | 3.9063 |  32.45 KB |
| Encrypt_ECB | 265.6 us | 4.70 us | 4.17 us | 3.9063 |  32.37 KB |
| Decrypt_ECB | 262.4 us | 3.48 us | 3.26 us | 3.9063 |  32.37 KB |

Results on this PR: (https://github.com/zybexXL/SSH.NET/commit/63bfc766c80893029488a9bc8edbef4c4df0533e)
| Method      | Mean      | Error     | StdDev    | Gen0   | Allocated |Speedup%|
|------------ |----------:|----------:|----------:|-------:|----------:|--------|
| Encrypt_CBC | 31.023 us | 0.3831 us | 0.3584 us | 4.0283 |  33.09 KB |   863% |
| Decrypt_CBC |  7.757 us | 0.1110 us | 0.0866 us | 4.0436 |  33.09 KB |  3766% |
| Encrypt_CFB | 48.047 us | 0.6268 us | 0.5556 us | 4.0283 |  33.09 KB |   551% |
| Decrypt_CFB | 42.820 us | 0.3457 us | 0.3234 us | 4.0283 |  33.09 KB |   640% |
| Encrypt_CTR | 17.283 us | 0.3206 us | 0.3149 us | 8.0872 |   66.2 KB |  1616% |
| Decrypt_CTR | 17.248 us | 0.3321 us | 0.4656 us | 8.0872 |   66.2 KB |  1631% |
| Encrypt_ECB |  7.054 us | 0.1356 us | 0.1714 us | 4.0283 |  33.05 KB |  3665% |
| Decrypt_ECB | 12.527 us | 0.2423 us | 0.3475 us | 4.0283 |  33.05 KB |  1995% |

Up to 38x speedup on CBC/ECB and 17x in CTR 😎

zybexXL:
@Rob-Hague Please review the changes to [AesCipherTest.Gen.cs.txt](https://github.com/sshnet/SSH.NET/pull/865/files#diff-43b6771b6cc38a872368b5e5c03a437f834c9bee5de56e40684755249f669346) - I changed the script but did not run it to check for a valid output. I fixed `AesCipherTest.cs` manually.

Rob-Hague:
Thanks. I'll send a review over the next few days.

zybexXL:
Reduced memory usage for CTR under Net 6+ using `Span<T>` and buffer reuse. Also increased performance a notch:

| Method      | Mean     | Error    | StdDev   | Gen0   | Allocated |
|------------ |---------:|---------:|---------:|-------:|----------:|
| Encrypt_CTR | 15.38 us | 0.303 us | 0.394 us | 4.1504 |  34.17 KB |
| Decrypt_CTR | 14.29 us | 0.273 us | 0.364 us | 4.1504 |  34.17 KB |

WojciechNagorski:
@Rob-Hague, @zybexXL Which of these two PRs should be checked first? #1235 #865 They are related and I don't really know where to start.

Rob-Hague:
Please merge #1235, it will reduce the diff for this one.

I have opened https://github.com/zybexXL/SSH.NET/pull/1 as initial feedback for this one. I will send my approval once it is ready.

zybexXL:
@Rob-Hague @WojciechNagorski 
The PRs do the same basic thing (AES hardware acceleration), but #1235 achieves 0.5x performance gain while #865 gives 6x to 36x gain. 1235 is just a proof of concept (not needed, since this PR was pending for 2 years and in production use for 3 years in my company, and with other comments above saying that it works...). 865 is a full optimized implementation covering all use cases and AES modes.

I've restructured this PR to fit Rob's concept of "nice code". If there are further concerns, add a code review. I know everyone else's code always smells, but enough is enough - this is a collaboration, not a competition.

Rob-Hague:
> this is a collaboration, not a competition.

Indeed, where has it been stated otherwise?

> I've restructured this PR to fit Rob's concept of "nice code". If there are further concerns, add a code review.

I don't believe you did structure it as I suggested, but I was aware that I could be unclear. That's why I opened [zybex](https://github.com/zybexXL/SSH.NET/pull/1).

src/Renci.SshNet/ConnectionInfo.cs
```diff
@@ -360,11 +360,13 @@ public ConnectionInfo(string host, int port, string username, ProxyTypes proxyTy
...
-                    { "aes192-cbc", new CipherInfo(192, (key, iv) => new AesCipher(key, new CbcCipherMode(iv), padding: null)) },
-                    { "aes256-cbc", new CipherInfo(256, (key, iv) => new AesCipher(key, new CbcCipherMode(iv), padding: null)) },
+                    { "aes128-cbc", new CipherInfo(128, (key, iv) => new AesCipher(key, iv, AesCipherMode.CBC, pkcs7Padding: false)) },
+                    { "aes192-cbc", new CipherInfo(192, (key, iv) => new AesCipher(key, iv, AesCipherMode.CBC, pkcs7Padding: false)) },
+                    { "aes256-cbc", new CipherInfo(256, (key, iv) => new AesCipher(key, iv, AesCipherMode.CBC, pkcs7Padding: false)) },
+                    { "aes128-ctr", new CipherInfo(128, (key, iv) => new AesCipher(key, iv, AesCipherMode.CTR, pkcs7Padding: false)) },
+                    { "aes192-ctr", new CipherInfo(192, (key, iv) => new AesCipher(key, iv, AesCipherMode.CTR, pkcs7Padding: false)) },
+                    { "aes256-ctr", new CipherInfo(256, (key, iv) => new AesCipher(key, iv, AesCipherMode.CTR, pkcs7Padding: false)) },
```
WojciechNagorski: Where the `Dispose` method of `AesClipher` is call? I've checked that other Clipher do not implement `IDisposable` right now.

WojciechNagorski (changes requested):

src/Renci.SshNet/ConnectionInfo.cs
zybexXL: I can remove the IDisposable, it's not really needed for the managed AES objects as they're GC collected anyway.
This can be revisited on a separate PR.

WojciechNagorski: You're right. AES objects only clear data in array in the dispose method.

WojciechNagorski:
@zybexXL I will review (and merge if OK) this pull request as soon as you update this pull request.

I merged #1235 PR before this because it reduced the diff for this one.

src/Renci.SshNet/ConnectionInfo.cs
```diff
@@ -360,11 +360,13 @@ public ConnectionInfo(string host, int port, string username, ProxyTypes proxyTy
 
             Encryptions = new Dictionary<string, CipherInfo>
```
WojciechNagorski, line 361: @Rob-Hague @drieseng Should we change the order of encriptions? I don't know why they were sorted in this way but I think there was a reason.

Rob-Hague: The order is important. We should have aes128-ctr,aes192-ctr,aes256-ctr first, then leave the rest as they are (I'm not sure we should use most of the rest...)

For reference, these are the values coming from `man ssh_config`:

```bash
 The supported ciphers are:

       3des-cbc
       aes128-cbc
       aes192-cbc
       aes256-cbc
       aes128-ctr
       aes192-ctr
       aes256-ctr
       aes128-gcm@openssh.com
       aes256-gcm@openssh.com
       chacha20-poly1305@openssh.com

 The default is:

       chacha20-poly1305@openssh.com,
       aes128-ctr,aes192-ctr,aes256-ctr,
       aes128-gcm@openssh.com,aes256-gcm@openssh.com
```

Note: a `Dictionary` doesn't guarantee order by contract, but without removing anything it will enumerate in the order the items were added

WojciechNagorski:
@zybexXL are you going to update this PR. I would see this PR as one of the most important for the next release.

zybexXL:
@WojciechNagorski Yes, I've just been busy this week. I'll update it today or tomorrow.

src/Renci.SshNet/ConnectionInfo.cs
zybexXL: [RFC4344](https://datatracker.ietf.org/doc/html/rfc4344#section-4) recommends AES-CTR modes, and also 3DES-CTR which we don't have yet. It says AES modes are preferred going forward.
The older [RFC4253](https://datatracker.ietf.org/doc/html/rfc4253#section-6.3) required 3DES-CBC and recommended AES128-CBC.

I've reordered the cyphers accordingly. 3DES-CTR can be added on a separate PR if needed.

A9G-Data-Droid: Not having 3DES is good.

If anyone is going to add 3DES it would need to come with a `[Deprecated]` flag. It should only be used for interoperability with legacy systems that cannot be updated to use modern encryption. I would recommend following OpenSSL:

[OpenSSL](https://en.wikipedia.org/wiki/OpenSSL) does not include 3DES by default since version 1.1.0 (August 2016) and considers it a "weak cipher".[[24]](https://en.wikipedia.org/wiki/Triple_DES#cite_note-24)

zybexXL: I'll leave that for a separate PR then.

zybexXL:
@WojciechNagorski PR is now updated, ready to merge.

src/Renci.SshNet/Security/Cryptography/Ciphers/AesCipher.CtrImpl.cs
```diff
@@ -0,0 +1,230 @@
...
+                    {
+                        uint j = 3;
+                        do
+                        {
+                            _packedIV[j] = SwapEndianness(SwapEndianness(_packedIV[j]) + 1);
+                        }
+                        while (_packedIV[j] == 0 && --j >= 0);
+                    }
+                }
+            }
```
Rob-Hague: The `_packedIV` logic looks unnecessarily complicated. I would be surprised if it is any better than simply:

```suggestion
            private static byte[] CTRCreateCounterArray(byte[] iv, int blocks)
            {
                Debug.Assert(iv.Length == 16);
                
                var counter = new byte[blocks * 16];
                
                for (int i = 0; i < blocks; i++)
                {
                    Buffer.BlockCopy(iv, 0, counter, i * 16, 16);
            
                    var j = 16;
                    while (--j >= 0 && ++iv[j] == 0)
                    {
                        // Intentionally empty block
                    }
                }
                
                return counter;
            }
```

Which would work on all platforms. This is not blocking feedback.

src/Renci.SshNet/Security/Cryptography/Ciphers/AesCipher.CtrImpl.cs
```diff
@@ -0,0 +1,230 @@
...
+                    else
+                    {
+                        for (var i = 0; i < length; i++)
+                        {
+                            buffer[loopOffset] ^= data[offset + loopOffset];
+                            loopOffset++;
+                        }
+                    }
+                }
+            }
```
Rob-Hague: This can be:

```suggestion
            private static void ArrayXOR(byte[] buffer, byte[] data, int offset, int length)
            {
                int i = 0;
            
#if NETSTANDARD2_1_OR_GREATER || NET
                int oneVectorFromEnd = length - Vector<byte>.Count;
                
                for (; i <= oneVectorFromEnd; i += Vector<byte>.Count)
                {
                    var v = new Vector<byte>(buffer, i) ^ new Vector<byte>(data, offset + i);
                    v.CopyTo(buffer, i);
                }
#endif
                
                for (; i < length; i++)
                {
                    buffer[i] ^= data[offset + i];
                }
            }
```

Which would work on all platforms (the .NET Framework version looks like a deoptimization at first glance). This is not blocking feedback.

src/Renci.SshNet/Security/Cryptography/Ciphers/AesCipher.cs
```diff
@@ -1,86 +1,114 @@
...
 
+using Renci.SshNet.Security.Cryptography.Ciphers.Modes;
+using Renci.SshNet.Security.Cryptography.Ciphers.Paddings;
+
 namespace Renci.SshNet.Security.Cryptography.Ciphers
 {
+    /// <summary>
+    /// Custom AES Cipher Mode, follows System.Security.Cryptography.CipherMode.
+    /// </summary>
+    public enum AesCipherMode
```
Rob-Hague: We should name this something without "Aes", but I don't have any great suggestions.

1. `CipherModeType` / `CipherType` ?
2. Nest it in `Cipher` / `BlockCipher` / `CipherMode` as `Mode` / `Type` ? e.g.
```diff
namespace Renci.SshNet.Security.Cryptography;

public class Cipher
{
+   public enum Mode
+   {
+       CBC = 1,
    // etc
+   }
}
```
Then we have `new AesCipher(key, iv, Cipher.Mode.CBC);`

3. Remove it, replace it with `S.S.C.CipherMode` and define a constant value on `CipherMode` e.g.
```diff
namespace Renci.SshNet.Security.Cryptography.Ciphers;

public abstract class CipherMode
{
+   public const System.Security.Cryptography.CipherMode CTR = (System.Security.Cryptography.CipherMode)(-1);
}
```
Then we have `new AesCipher(key, iv, S.S.C.CipherMode.CBC)` and `new AesCipher(key, iv, Renci..CipherMode.CTR)`

I'm not hot on any of them.

Rob-Hague (commented):
Nice. My only "important" feedback is on the naming of `AesCipherMode`. The others are not necessary to address.

src/Renci.SshNet/Security/Cryptography/Ciphers/AesCipher.CtrImpl.cs
zybexXL: Current code:
| Method      | Mean     | Error    | StdDev   | Gen0   | Allocated |
|------------ |---------:|---------:|---------:|-------:|----------:|
| Encrypt_CTR | 12.39 us | 0.245 us | 0.319 us | 3.9978 |   32.8 KB |
| Decrypt_CTR | 12.52 us | 0.249 us | 0.417 us | 3.9978 |   32.8 KB |

Your proposed code:
| Method      | Mean     | Error    | StdDev   | Gen0   | Allocated |
|------------ |---------:|---------:|---------:|-------:|----------:|
| Encrypt_CTR | 22.93 us | 0.449 us | 0.480 us | 3.9978 |   32.8 KB |
| Decrypt_CTR | 22.32 us | 0.290 us | 0.257 us | 3.9978 |   32.8 KB |

I had already tried that, it's the naive implementation. The NetFramework fallback code is an optimization of that, and the Net7 code is twice as fast. Blockcopy is actually slow for small buffers, working with int32 is much faster. I also tested int64, there's no significant difference.

src/Renci.SshNet/Security/Cryptography/Ciphers/AesCipher.CtrImpl.cs
zybexXL: Current ArrayXOR under Net Framework:
| Method      | Mean     | Error    | StdDev   | Gen0    | Allocated |
|------------ |---------:|---------:|---------:|--------:|----------:|
| Encrypt_CTR | 19.97 us | 0.398 us | 0.595 us | 11.8103 |  96.84 KB |
| Decrypt_CTR | 18.88 us | 0.378 us | 0.388 us | 11.8103 |  96.84 KB |

Your code under Net Framework (just this change):
| Method      | Mean     | Error    | StdDev   | Gen0   | Allocated |
|------------ |---------:|---------:|---------:|-------:|----------:|
| Encrypt_CTR | 29.97 us | 0.553 us | 0.517 us | 3.9978 |   32.8 KB |
| Decrypt_CTR | 29.73 us | 0.569 us | 0.585 us | 3.9978 |   32.8 KB |

I did mention in some other comment that I traded memory usage for performance for the NetFramework code. I think that's OK as it's a dying platform anyway, and buffer size won't be larger than 32K.

I do like your Net7 refactoring - the performance is the same but it's simpler. I'll commit it but will leave the NetFramework variant in place as it's faster.

src/Renci.SshNet/Security/Cryptography/Ciphers/AesCipher.cs
zybexXL: AesCipherMode is not ideal, but I think it's OK as it's only used with AesCipher.cs. There's already another class called CipherMode.

zybexXL:
Appveyor failed on unrelated test (ForwardedPortShouldAcceptNewConnections).

I've noticed that some tests always fail if there are 2 or more Appveyor sessions running at the same time, usually because some port is already in use. Maybe there's a way to queue the appveyor sessions so that only one runs at any given time?

Edit - consider adding this to appveyor.yml:
```
# Maximum number of concurrent jobs for the project
max_jobs: 1

```

src/Renci.SshNet/Security/Cryptography/Ciphers/AesCipher.CtrImpl.cs
Rob-Hague: Fair enough

src/Renci.SshNet/Security/Cryptography/Ciphers/AesCipher.CtrImpl.cs
Rob-Hague: Ok, fair. I personally prefer simplicity and maintainability for older frameworks. We are also only considering 1 packet in isolation. When there are lots of packets on the wire, do we really want to be spending 3x the memory?

src/Renci.SshNet/Security/Cryptography/Ciphers/AesCipher.cs
Rob-Hague: Right now it's only being used for the AES implementation, but a block cipher mode is not something specific to AES. Maybe we will use it for 3DES?

 We can let Wojtek be the arbiter.

src/Renci.SshNet/Security/Cryptography/Ciphers/AesCipher.CtrImpl.cs
zybexXL: These are short lived buffers, it's just a bit more work for the GC.

I ran a benchmark of encrypting 1024 x 32KB buffers (32 MB per iteration). In Net462, total active process memory hovered around 12 MB from start to finish. Under Net7 memory usage was actually higher at 22 MB. Both cases showed stable memory usage, looks like GC is doing a good job in both frameworks despite total allocated memory being 3x larger in Net462.

src/Renci.SshNet/Security/Cryptography/Ciphers/AesCipher.CtrImpl.cs
Rob-Hague: 1. It seems like the `else` block does not get hit in the AES tests. Could you please add one which does? e.g. using an IV of 0xFF..FF

2. The endianness logic needs to cater for big-endian systems, i.e. using [BitConverter.IsLittleEndian](https://learn.microsoft.com/en-us/dotnet/api/system.bitconverter.islittleendian). Something like

```c#
_packedIV[j] = BitConverter.IsLittleEndian
    ? SwapEndianness(SwapEndianness(_packedIV[j]) + 1)
    : _packedIV[j] + 1;
```
And the same for the `if` block above.

Alternatively... https://github.com/Rob-Hague/SSH.NET/commit/25172b241cf706c5f3190366731d977eda52003a

Rob-Hague:
The [appveyor](https://www.appveyor.com/pricing/) open source licence only allows 1 concurrent job, and I don't see any overlapping runs in the history. The failed test is not using any sockets. It is probably a timing issue like #1185

zybexXL:
> The [appveyor](https://www.appveyor.com/pricing/) open source licence only allows 1 concurrent job, and I don't see any overlapping runs in the history. The failed test is not using any sockets. It is probably a timing issue like #1185

I'm not talking about this particular failure. The fact is, whenever I push to multiple branches/PRs, Appveyor testing fails. I've seen port-related issues at least twice.

src/Renci.SshNet/Security/Cryptography/Ciphers/AesCipher.CtrImpl.cs
zybexXL: Are you telling me that the rest of SSH.Net works on Big Endian ARM v6 or others? I find that unlikely, given lots of code that fills in packet structures by simply assigning an int32. There are no tests for that, except for a PackTest that tests conversion functions that are mostly only used in that test class. Nowadays even ARM is little endian by default.

zybexXL: > Alternatively... [Rob-Hague@25172b2](https://github.com/Rob-Hague/SSH.NET/commit/25172b241cf706c5f3190366731d977eda52003a)

Please provide performance numbers for this change.

zybexXL: > 1. It seems like the `else` block does not get hit in the AES tests. Could you please add one which does? e.g. using an IV of 0xFF..FF

Please provide a PR for this.

WojciechNagorski:
To be honest, I prefer GitHub Actions but we can't use it because there is a limit.

src/Renci.SshNet/Security/Cryptography/Ciphers/AesCipher.CtrImpl.cs
Rob-Hague: > Are you telling me that the rest of SSH.Net works on Big Endian ARM v6 or others? I find that unlikely, given lots of code that fills in packet structures by simply assigning an int32.

Could you please provide an example? On a cursory glance I could not find any offenders.

I don't know if the library works on big endian architecture, but I know that this code doesn't, so why not get it right?

> Please provide performance numbers for this change.

```

BenchmarkDotNet v0.13.9+228a464e8be6c580ad9408e98f18813f6407fb5a, Windows 11 (10.0.22621.2715/22H2/2022Update/SunValley2)
AMD Ryzen 7 5700U with Radeon Graphics, 1 CPU, 16 logical and 8 physical cores
.NET SDK 8.0.100
  [Host]     : .NET 7.0.14 (7.0.1423.51910), X64 RyuJIT AVX2
  DefaultJob : .NET 7.0.14 (7.0.1423.51910), X64 RyuJIT AVX2


```
d34f521b4acfe6baa02e53a
| Method      | Mean     | Error    | StdDev   | Gen0    | Allocated |
|------------ |---------:|---------:|---------:|--------:|----------:|
| Encrypt_CTR | 11.66 μs | 0.125 μs | 0.111 μs | 15.8691 |   32.8 KB |
| Decrypt_CTR | 11.66 μs | 0.088 μs | 0.078 μs | 15.8691 |   32.8 KB |

25172b241cf706c5f3190366
| Method      | Mean     | Error     | StdDev    | Gen0    | Allocated |
|------------ |---------:|----------:|----------:|--------:|----------:|
| Encrypt_CTR | 9.179 μs | 0.1522 μs | 0.1424 μs | 15.8691 |  32.77 KB |
| Decrypt_CTR | 8.909 μs | 0.0592 μs | 0.0462 μs | 15.8691 |  32.77 KB |

zybexXL: Nice performance bump, I'll merge the code.

> I don't know if the library works on big endian architecture, but I know that this code doesn't, so why not get it right?

Because the level of nitpicking on this PR is already absurd. You know you can add another PR once this one is merged right?


> Could you please provide an example? On a cursory glance I could not find any offenders.

This [here](https://github.com/sshnet/SSH.NET/blob/develop/src/Renci.SshNet/Messages/Message.cs#L107) calls the stream.Write method [here](https://github.com/sshnet/SSH.NET/blob/develop/src/Renci.SshNet/Common/SshDataStream.cs#L65) which [converts to BigEndian](https://github.com/sshnet/SSH.NET/blob/develop/src/Renci.SshNet/Common/Pack.cs#L174) without checking the current arch.

Rob-Hague: > This [here](https://github.com/sshnet/SSH.NET/blob/develop/src/Renci.SshNet/Messages/Message.cs#L107) calls the stream.Write method [here](https://github.com/sshnet/SSH.NET/blob/develop/src/Renci.SshNet/Common/SshDataStream.cs#L65) which [converts to BigEndian](https://github.com/sshnet/SSH.NET/blob/develop/src/Renci.SshNet/Common/Pack.cs#L174) without checking the current arch.

AFAIU this code will work fine. Bitwise operations are endianness-agnostic.

I'll let @WojciechNagorski finish off this PR.

zybexXL: > AFAIU this code will work fine. Bitwise operations are endianness-agnostic.

0x12345678 -> reverse = 0x78563412. When you then store it in memory, a little endian CPU will write 12 34 56 78 (as needed there), but a big endian CPU will now write 78 56 34 12. To obtain the same result the BigEndian CPU should not reverse endianness.

zybexXL:
@WojciechNagorski Please merge this PR, I think all issues are resolved.

src/Renci.SshNet/Security/Cryptography/Ciphers/AesCipher.cs
drieseng: I'd keep **AesCipherMode**, and rename it if/when we decide to use it for other ciphers.
Breaking changes on this level shouldn't hurt many users.

drieseng: Please move to separate file.

drieseng (changes requested):

src/Renci.SshNet/Security/Cryptography/Ciphers/AesCipher.cs
zybexXL: done

drieseng (approved):

drieseng:
@WojciechNagorski, can this one go in for you?

WojciechNagorski (approved):
LGTM! This is an amazing contribution. 

@Rob-Hague I know you had some comments about AesCipherMode, but to be honest, it doesn't block merging. Please feel free to open the next PR. We still have some time until the next release.

PR #876, opened 2021-09-20
AES-GCM Cipher Suite for NETCOREAPP3_0 || NETSTANDARD2_1

jy824212:


jy824212:
Why I keep seeing "drieseng and others added 26 commits on Oct 15, 2017" from  above, is this the correct way to create a new PR?

PR #877, opened 2021-09-21
AES-GCM support under Netstandard2.1, Netstandard2.0 and Net Core 3.0

jy824212:
This is to support AES-GCM algo and currently only support netstandard 2.0,  netstandard 2.1 and netcoreapp3.0


jy824212:
Dear @drieseng ,

I wonder if you can look into this error as I didn't encounter any of these errors during my compilation.

Regards,
Jy

zybexXL:
AppVeyor is failing on many PRs, I think this is the cause - https://github.com/sshnet/SSH.NET/pull/855#discussion_r701714465

The error is "SocketException: Only one usage of each socket address (protocol/network address/port) is normally permitted."

The testsuite is now running thousands of testcases, and many of them open a connection. After the connection is closed, the port stays in TIME_WAIT for a long time (default on windows is 240 seconds). Dynamic connections use a port from 1024 to 5000, so there are "only" 4000 ports available during that 240s window... I think AppVeyor/SSHNet is hitting this limit.

The fix is to add some flag to make the tests create sockets using SO_REUSEADDR, or to change the AppVeyor VM config to reduce the 240s timeout (setting in the Registry on Windows, somewhere else on Linux).

I don't recommend using SO_REUSEADDR by default as that may cause some pending packets to be dropped when the connection is closed.

jy824212:
@zybexXL thanks for the reply and much appreciated. Do you know who is able to re configure this on the AppVeyor VM or is there something I can do from my local appveyor.yml

jy824212:
Hi @drieseng,

Will you be able to take a look at this ?

Regards,
Jy

src/Renci.SshNet/Security/Cryptography/EcdsaDigitalSignature.cs
```diff
@@ -39,7 +39,7 @@ public override bool Verify(byte[] input, byte[] signature)
             // for 521 sig_size is 132
             var sig_size = _key.KeyLength == 521 ? 132 : _key.KeyLength / 4;
             var ssh_data = new SshDataSignature(signature, sig_size);
-#if NETSTANDARD2_0
+#if NETSTANDARD2_0 || NETSTANDARD2_1
```
IgorMilavec, line 42: Please use ```#if NETSTANDARD2_0_OR_GREATER || NETCOREAPP3_1_OR_GREATER``` in all instances, so this will be compiled when we target higher frameworks.

src/Renci.SshNet/Security/Cryptography/Ciphers/AEADCipher.cs
```diff
@@ -0,0 +1,423 @@
...
+            var plainText = new ReadOnlySpan<byte>(inputBuffer, PACKET_LENGTH_FIELD_SZ, cipherText.Length);
+
+#if DEBUG_AESGCM
+
+            string var_iv = GetHexStringFrom(_iv);
+            string var_key = GetHexStringFrom(_key);
+            string var_pt = GetHexStringFrom(plainText.ToArray());
+            string var_aad = GetHexStringFrom(associatedData.ToArray());
+
+            Console.WriteLine("[DEBUG] In GcmEncrypt()");
```
IgorMilavec, line 176: Have you considered other possibilities for outputting debug information? ```ILogger``` is not viable yet, but I would prefer ```System.Diagnostics.Debug.WriteLine``` over ```Console.WriteLine```.

src/Renci.SshNet/Security/Cryptography/EcdsaDigitalSignature.cs
jy824212: certainly I can make such adjustment

src/Renci.SshNet/Security/Cryptography/Ciphers/AEADCipher.cs
jy824212: yes, I will follow your suggestion as at the time I wrote the code I wasn't too familiar with the C# DEBUG utility. Thank you for the feedback.

jy824212: I am getting this error 
Severity	Code	Description	Project	File	Line	Suppression State
Error	CS0246	The type or namespace name 'CngKey' could not be found (are you missing a using directive or an 
Severity	Code	Description	Project	File	Line	Suppression State
Error	CS0246	The type or namespace name 'CngAlgorithm' could not be found (are you missing a using directive or an assembly reference?)

Is because these are out of date in netstandard2.0 and netstandard2.1 ?

fmargueirat:
@jy824212 I was wondering if you were able to have this code running on later versions of Renci.SSH.NET. I try to add your code to the latest one, but I'm facing some issues.

PR #892, opened 2021-10-25
Mac/776 dotnetcore issue

mmacagno:
Fix the issue #776 
With same PR, I have also removed .NET Framework 3.5 target because the 3.5 SDK can no longer be downloaded from Microsoft as no longer supported for several years now.


WojciechNagorski:
The .NET 3.5 has been removed and the new frameworks have been added in #1109 PR.

Issue #613, opened 2020-01-12
Support For Other Fingerprint Algorithms


StrangeWill:
Currently SSH.NET only supports MD5, with the big-ish struggles right not being that `FingerPrint` being just a bite array, so something we can't easily add new algorithm support to.

We're currently calculating our own fingerprint for SHA256 support, but ideally having the library support this out of the box would be best.

To prevent breaking backwards comparability we can:

- Make additional properties per algorithm (not the best, easy though, accessible, matches current "convention")
- Make a new property that allows us to look up by algorithm, all algorithms need to be checked up-front and stored, but we can add more
- Make a method that takes an algorithm that we'll return the thumbprint for, internal implementation may cache this

What probably makes the most sense? Or does everyone just roll their own and it's not really a big deal?

0xced:
I just submitted #1163 which should address this issue by adding two new (well documented) properties: `MD5FingerPrint` and `SHA256FingerPrint`.

0xced:
Note that you can already compute the SHA256 fingerprint as of SSH.NET 2020.0.2. The fingerprint is nothing more than a hash of the host key. Here's how you can compute it:

```csharp
const string expectedFingerPrint = "ohD8VZEXGWo6Ez8GSEJQ9WpafgLFsOfLOtGGQCQo6Og";
using var client = new SshClient(host: "ssh.dev.azure.com", username: "git");
client.HostKeyReceived += (_, e) =>
{
	var sha256FingerPrint = Convert.ToBase64String(SHA256.HashData(e.HostKey)).TrimEnd('=');
	e.CanTrust = sha256FingerPrint == expectedFingerPrint;
};
client.Connect();
```

The pull request I submitted merely addresses discoverability (thanks to autocompletion) and standardization (hex format for MD5 and unpadded base64 format for SHA256).

WojciechNagorski:
Version 2023.0.0 has been published https://www.nuget.org/packages/SSH.NET/2023.0.0

PR #897, opened 2021-11-15
Simplify sample code for verifying host identify

0xced:
* Include `using` directives to make it clear that `System.Linq` and `Renci.SshNet` are required
* Remove unnecessary `new byte[]` in `expectedFingerPrint` declaration
* Use [discard](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/discards) (i.e. `_`) instead of unused `sender` argument (C# 7.0)
* Use [Enumerable.SequenceEqual](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.sequenceequal) instead of manually comparing byte arrays
* `using var client` without braces (C# 8.0 feature)

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,33 +143,17 @@ using (var client = new SftpClient(connectionInfo))
 Establish a SSH connection using user name and password, and reject the connection if the fingerprint of the server does not match the expected fingerprint:
 
 ```cs
-byte[] expectedFingerPrint = new byte[] {
-                                            0x66, 0x31, 0xaf, 0x00, 0x54, 0xb9, 0x87, 0x31,
-                                            0xff, 0x58, 0x1c, 0x31, 0xb1, 0xa2, 0x4c, 0x6b
-                                        };
+using System.Linq;
+using Renci.SshNet;
 
-using (var client = new SshClient("sftp.foo.com", "guest", "pwd"))
-{
-    client.HostKeyReceived += (sender, e) =>
-        {
-            if (expectedFingerPrint.Length == e.FingerPrint.Length)
-            {
-                for (var i = 0; i < expectedFingerPrint.Length; i++)
-                {
-                    if (expectedFingerPrint[i] != e.FingerPrint[i])
-                    {
-                        e.CanTrust = false;
-                        break;
-                    }
-                }
-            }
-            else
-            {
-                e.CanTrust = false;
-            }
-        };
-    client.Connect();
-}
+byte[] expectedFingerPrint = {
+    0x66, 0x31, 0xaf, 0x00, 0x54, 0xb9, 0x87, 0x31,
+    0xff, 0x58, 0x1c, 0x31, 0xb1, 0xa2, 0x4c, 0x6b
+};
+
+using var client = new SshClient("sftp.foo.com", "guest", "pwd");
+client.HostKeyReceived += (_, e) => e.CanTrust = expectedFingerPrint.SequenceEqual(e.FingerPrint);
+client.Connect();
 ```
 
 ## Supporting SSH.NET


A9G-Data-Droid:
I would like a note about how to get the key into `byte[]` format.

Related reading: https://winscp.net/eng/docs/faq_hostkey

That only works if you have the ability to get the public key from a known good source.

I like to store it in a Base64 string. So I'm pulling it from a config file like this:

```C#
byte[] expectedFingerPrint = Convert.FromBase64String(ServerFingerPrint);
```

If you want to pull the fingerprint from your code so you can accept the server fingerprint you need something like this:

```C#
client.HostKeyReceived += (_, e) =>
{
    ServerFingerPrint = Convert.ToBase64String(e.FingerPrint);
    e.CanTrust = expectedFingerPrint.SequenceEqual(e.FingerPrint);
};
```

Then when `.CanTrust` is false you can give the user the option of saving this fingerprint, which is a normal workflow in most SSH\SFTP clients.

This at least serves to notify you when the fingerprint has changed, which is better than nothing.

WojciechNagorski:
Old PR, closing for now.

0xced:
Sure it's old but I think it's still relevant, isn't it?

If there are conflicts with the current state of the `develop` branch I'd be happy to resolve them.

I'd understand if you don't want to merge this because you'd prefer to keep the sample code compatible with an older C# version or maybe for another reason that I haven't thought about. But closing simply because the pull request old is really disappointing and will likely discourage me to further contribute to this project. 😞

WojciechNagorski:
@0xced regarding this comment https://github.com/sshnet/SSH.NET/pull/897#issuecomment-1274934059

the check should look like below, right?
```c#
client.HostKeyReceived += (_, e) =>
{
    ServerFingerPrint = Convert.ToBase64String(e.FingerPrint);
    e.CanTrust = expectedFingerPrint.SequenceEqual(e.FingerPrint);
};
```

If you prepare a good PR, I will definitely merge it.

0xced:
Well, Adam said
> I like to store it in a Base64 string.

Personally, I store them in this format: `97:70:33:82:fd:29:3a:73:39:af:6a:07:ad:f8:80:49` because it's an MD5 hash (and that's how Microsoft displays it in their [Azure DevOps SSH - Public Keys page](https://learn.microsoft.com/en-us/azure/devops/repos/git/use-ssh-keys-to-authenticate?view=azure-devops#step-2-add-the-public-key-to-azure-devops)):

Azure DevOps Server Fingerprints (both MD5 and SHA256)

```csharp
using System.Globalization;
using System.Linq;
using Renci.SshNet;

const string expectedFingerPrint = "97:70:33:82:fd:29:3a:73:39:af:6a:07:ad:f8:80:49";
using var client = new SshClient(host: "ssh.dev.azure.com", port: 22, username: "git", password: "");
client.HostKeyReceived += (_, e) =>
{
    var md5FingerPrint = string.Join(':', e.FingerPrint.Select(x => x.ToString("x2", NumberFormatInfo.InvariantInfo)));
    e.CanTrust = md5FingerPrint == expectedFingerPrint;
};
client.Connect();
```

It would even be misleading to use base64 since it's an MD5 hash and usually the fingerprint displayed as base64 are SHA256 hashes, see also [GitHub's SSH key fingerprints](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/githubs-ssh-key-fingerprints).

I still think my pull request is good as is. 🤷

A9G-Data-Droid:
@0xced That's a great point! In newer versions of OpenSSH, Base64 encoded SHA-256 is the default instead of hexadecimal MD5. This topic is related to #613.

I agree that the way you have shown is industry standard, as OpenSSH also reports this colon separated hex format.

The example you give in this PR is good, better than it was. I like your new example better. I would use `BitConverter` instead of rolling my own.

```c#
using System.Globalization;
using System.Linq;
using Renci.SshNet;

const string expectedFingerPrint = "97:70:33:82:fd:29:3a:73:39:af:6a:07:ad:f8:80:49";
using var client = new SshClient("sftp.foo.com", "guest", "pwd")
client.HostKeyReceived += (_, e) =>
{
    var md5FingerPrint = BitConverter.ToString(e.FingerPrint).Replace("-", ":");
    e.CanTrust = md5FingerPrint == expectedFingerPrint;
};

client.Connect();
```

0xced:
Related: I just submitted #1163 which adds two new (well documented) properties: `MD5FingerPrint` and `SHA256FingerPrint`. Inspired by this discussion of course. 😉

And if #1163 gets merged the sample code could be improved even further.

About `BitConverter` vs `string.Join()` we should benchmark it (both speed and allocations) before choosing the _best_ implementation.

WojciechNagorski:
closed by #1186. Readme has been updated.

WojciechNagorski:

PR #901, opened 2021-11-21
Start MessageListener with ThreadAbstraction.ExecuteThreadLongRunning

### LINKED ISSUE #824: Thread pool exhaustion due to MessageListener running on ThreadPool
Session.Connect() (and Session.ConnectAsync() in #819) start MessageListener through ThreadAbstraction that runs the connection message pump on a ThreadPool thread. This causes thread pool exhaustion when the number of connected sessions reaches the maximum number of thread pool threads. Message loop should not be run through thread pool, as this consumes a (limited) shared resource. Also, running a thread for each session impacts system performance and limits scalability of the application.
Preferably we should avoid consuming an entire thread and make the message loop async.

IgorMilavec:
Changed the ```Session.Connect``` to use ```ThreadAbstraction.ExecuteThreadLongRunning``` to start the ```MessageListener```. 

The commit 5d173c8 also includes cleanup of ```ThreadAbstraction```, which I did because I wasn't sure anymore which target was used as a dependency. I feel this cleanup is beneficial in the long run, but if you feel this is too much for this PR let me know and I'll roll it back.

Fixes #824

### DIFF
--- a/src/Renci.SshNet/Abstractions/ThreadAbstraction.cs
+++ b/src/Renci.SshNet/Abstractions/ThreadAbstraction.cs
@@ -21,12 +21,16 @@ namespace Renci.SshNet.Abstractions
 
         public static void ExecuteThreadLongRunning(Action action)
         {
+            if (action == null)
+                throw new ArgumentNullException("action");
 #if FEATURE_THREAD_TAP
             var taskCreationOptions = System.Threading.Tasks.TaskCreationOptions.LongRunning;
             System.Threading.Tasks.Task.Factory.StartNew(action, taskCreationOptions);
 #else
-            var thread = new System.Threading.Thread(() => action());
-            thread.Start();
+            new System.Threading.Thread(() => action())
+            {
+                IsBackground = true
+            }.Start();
 #endif
         }
 
--- a/src/Renci.SshNet/Abstractions/ThreadAbstraction.cs
+++ b/src/Renci.SshNet/Abstractions/ThreadAbstraction.cs
@@ -10,28 +10,17 @@ namespace Renci.SshNet.Abstractions
         /// <param name="millisecondsTimeout">The number of milliseconds for which the thread is suspended.</param>
         public static void Sleep(int millisecondsTimeout)
         {
-#if FEATURE_THREAD_SLEEP
             System.Threading.Thread.Sleep(millisecondsTimeout);
-#elif FEATURE_THREAD_TAP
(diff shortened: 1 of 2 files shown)


src/Renci.SshNet/Abstractions/ThreadAbstraction.cs
```diff
@@ -36,16 +29,10 @@ public static void ExecuteThreadLongRunning(Action action)
         /// <param name="action">The action to execute.</param>
         public static void ExecuteThread(Action action)
         {
-#if FEATURE_THREAD_THREADPOOL
```
drieseng, line 39: I'm afraid you'll need to revert these changes. ThreadPool is not available for **UAP10.0** TFM (see https://github.com/sshnet/SSH.NET/wiki/Conditional-compilation-symbols for more information).

src/Renci.SshNet/Abstractions/ThreadAbstraction.cs
```diff
@@ -10,24 +10,17 @@ internal static class ThreadAbstraction
...
-#elif FEATURE_THREAD_TAP
-            System.Threading.Tasks.Task.Delay(millisecondsTimeout).GetAwaiter().GetResult();
-#else
-            #error Suspend of the current thread is not implemented.
-#endif
         }
 
         public static void ExecuteThreadLongRunning(Action action)
         {
-#if FEATURE_THREAD_TAP
```
drieseng, line 24: I propose to leave this as is. Eventually, we'll need to change this as we need the thread or task that we started for our message listener in order to avoid thread leaks (there's an issue for this).

src/Renci.SshNet/Abstractions/ThreadAbstraction.cs
```diff
@@ -10,24 +10,17 @@ internal static class ThreadAbstraction
         /// <param name="millisecondsTimeout">The number of milliseconds for which the thread is suspended.</param>
         public static void Sleep(int millisecondsTimeout)
         {
-#if FEATURE_THREAD_SLEEP
```
drieseng, line 13: I'm afraid you'll need to revert these changes. `Thread.Sleep(...)` is not available for **UAP10.0** TFM (see https://github.com/sshnet/SSH.NET/wiki/Conditional-compilation-symbols for more information).

drieseng (changes requested):

IgorMilavec:
Sorry, I've made an error to "Fetch upstream" from GitHub and it messed up the commit history.
So it was more clean to just do a new PR #902. I'm closing this one.

PR #902, opened 2021-11-22
Start MessageListener with ThreadAbstraction.ExecuteThreadLongRunning #2

### LINKED ISSUE #824: Thread pool exhaustion due to MessageListener running on ThreadPool

IgorMilavec:
Changed the Session.Connect to use ThreadAbstraction.ExecuteThreadLongRunning to start the MessageListener.

This is a replacement PR for #901.

Fixes #824

### DIFF
--- a/src/Renci.SshNet/Abstractions/ThreadAbstraction.cs
+++ b/src/Renci.SshNet/Abstractions/ThreadAbstraction.cs
@@ -21,12 +21,17 @@ namespace Renci.SshNet.Abstractions
 
         public static void ExecuteThreadLongRunning(Action action)
         {
+            if (action == null)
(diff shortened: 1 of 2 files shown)


WojciechNagorski:

PR #910, opened 2022-01-05
Fix Seek Operations in SftpFileStream

### LINKED ISSUE #909: Incorrect Seek Operations in SftpFileStream
When `SftpFileStream.Seek()` is called with the `origin` parameter set to `SeekOrigin.End` and a negative offset, the resulting position is greater than the size of the stream. This is because the negative offset is subtracted from the total size, instead of added.
https://github.com/sshnet/SSH.NET/blob/7bdfc9e615b736b2762f5cd83c31a5f669663ff6/src/Renci.SshNet/Sftp/SftpFileStream.cs#L802
https://github.com/sshnet/SSH.NET/blob/7bdfc9e615b736b2762f5cd83c31a5f669663ff6/src/Renci.SshNet/Sftp/SftpFileStream.cs#L853

```
size   offset
1000 - (-42) = 1042
1000 + (-42) = 958
```
Failed test using `SftpFileStream.Seek()`:
```
FluentAssertions.Execution.AssertionFailedException
  HResult=0x80131500
  Message=Expected stream.Seek(-5, SeekOrigin.End) to be 22L, but found 32L.
```
Furthermore, the tests for `SftpFileStream` only cover cases with the `origin` set to `SeekOrigin.Begin`, and not `SeekOrigin.Current` or `SeekOrigin.End`.

### LINKED ISSUE #1018: SFTP file stream not working with System.IO.Compression.Zip
I am getting the following error when trying to instantiate a [ZipArchive ](https://learn.microsoft.com/en-us/dotnet/api/system.io.compression.ziparchive.-ctor?view=net-6.0#system-io-compression-ziparchive-ctor(system-io-stream-system-io-compression-ziparchivemode)) object using Sftp file stream. 
```
message='Exception: System.IO.InvalidDataException: Central Directory corrupt.
[2022-10-21T14:16:36.733Z]  ---> System.IO.IOException: Zip file corrupt: unexpected end of stream reached.
[2022-10-21T14:16:36.734Z]    at System.IO.Compression.ZipHelper.ReadBytes(Stream stream, Byte[] buffer, Int32 bytesToRead)
[2022-10-21T14:16:36.734Z]    at System.IO.Compression.ZipHelper.SeekBackwardsAndRead(Stream stream, Byte[] buffer, Int32& bufferPointer)
[2022-10-21T14:16:36.736Z]    at System.IO.Compression.ZipHelper.SeekBackwardsToSignature(Stream stream, UInt32 signatureToFind, Int32 maxBytesToRead)
[2022-10-21T14:16:36.736Z]    at System.IO.Compression.ZipArchive.ReadEndOfCentralDirectory()
[2022-10-21T14:16:36.737Z]    --- End of inner exception stack trace ---
[2022-10-21T14:16:36.737Z]    at System.IO.Compression.ZipArchive.ReadEndOfCentralDirectory()
[2022-10-21T14:16:36.738Z]    at System.IO.Compression.ZipArchive..ctor(Stream stream, ZipArchiveMode mode, Boolean leaveOpen, Encoding entryNameEncoding)
[2022-10-21T14:16:36.739Z]    at System.IO.Compression.ZipArchive..ctor(Stream stream, ZipArchiveMode mode, Boolean leaveOpen)
```

Looking at the backtrace, looks like when trying to read from the end of file it is getting this error: **Zip file corrupt: unexpected end of stream reached.**

my code is simple:
```
using(var remoteStream = sftpclient.OpenRead(filepath)
using(var zip = new ZipArchive(stream: remoteStream,  mode: ZipArchiveMode.Read, leaveOpen: true))
{
...
}
```

lemonyte:
Changed `SftpFileStream.Seek()` to add offset instead of subtracting when `origin` is set to `SeekOrigin.End`. Fixes #909, fixes #1018.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,3 @@ project.lock.json
 
 # Build outputs
 build/target/
-
-# Visual Studio cache/options directory
-.vs/
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,6 @@ project.lock.json
 
 # Build outputs
 build/target/
+
+# Visual Studio cache/options directory
+.vs/
--- a/src/Renci.SshNet.Tests/Classes/Sftp/SftpFileStreamTest_Seek_PositionedAtBeginningOfStream_OriginEndAndOffsetNegative.cs
+++ b/src/Renci.SshNet.Tests/Classes/Sftp/SftpFileStreamTest_Seek_PositionedAtBeginningOfStream_OriginEndAndOffsetNegative.cs
@@ -16,11 +16,12 @@ namespace Renci.SshNet.Tests.Classes.Sftp
         private int _bufferSize;
         private uint _readBufferSize;
         private uint _writeBufferSize;
+        private int _length;
         private byte[] _handle;
         private SftpFileStream _target;
         private int _offset;
         private SftpFileAttributes _attributes;
-        private EndOfStreamException _actualException;
+        private long _actual;
 
         protected override void SetupData()
         {
@@ -29,29 +30,38 @@ namespace Renci.SshNet.Tests.Classes.Sftp
             _random = new Random();
             _path = _random.Next().ToString();
             _fileMode = FileMode.OpenOrCreate;
-            _fileAccess = FileAccess.Read;
+            _fileAccess = FileAccess.Write;
             _bufferSize = _random.Next(5, 1000);
             _readBufferSize = (uint)_random.Next(5, 1000);
             _writeBufferSize = (uint)_random.Next(5, 1000);
+            _length = _random.Next(5, 10000);
             _handle = GenerateRandom(_random.Next(1, 10), _random);
-            _offset = _random.Next(int.MinValue, -1);
+            _offset = _random.Next(-_length, -1);
             _attributes = SftpFileAttributes.Empty;
         }
 
         protected override void SetupMocks()
         {
             SftpSessionMock.InSequence(MockSequence)
-                           .Setup(p => p.RequestOpen(_path, Flags.Read | Flags.CreateNewOrOpen, false))
+                           .Setup(session => session.RequestOpen(_path, Flags.Write | Flags.CreateNewOrOpen, false))
                            .Returns(_handle);
             SftpSessionMock.InSequence(MockSequence)
-                           .Setup(p => p.CalculateOptimalReadLength((uint)_bufferSize))
+                           .Setup(session => session.CalculateOptimalReadLength((uint)_bufferSize))
                            .Returns(_readBufferSize);
             SftpSessionMock.InSequence(MockSequence)
-                           .Setup(p => p.CalculateOptimalWriteLength((uint)_bufferSize, _handle))
+                           .Setup(session => session.CalculateOptimalWriteLength((uint)_bufferSize, _handle))
                            .Returns(_writeBufferSize);
-            SftpSessionMock.InSequence(MockSequence).Setup(p => p.IsOpen).Returns(true);
             SftpSessionMock.InSequence(MockSequence)
-                           .Setup(p => p.RequestFStat(_handle, false))
+                           .Setup(session => session.IsOpen)
+                           .Returns(true);
+            SftpSessionMock.InSequence(MockSequence)
+                           .Setup(session => session.RequestFStat(_handle, false))
+                           .Returns(_attributes);
+            SftpSessionMock.InSequence(MockSequence)
+                           .Setup(session => session.RequestFSetStat(_handle, _attributes));
+            SftpSessionMock.InSequence(MockSequence).Setup(session => session.IsOpen).Returns(true);
+            SftpSessionMock.InSequence(MockSequence)
+                           .Setup(session => session.RequestFStat(_handle, false))
                            .Returns(_attributes);
         }
 
@@ -60,43 +70,34 @@ namespace Renci.SshNet.Tests.Classes.Sftp
             base.Arrange();
 
             _target = new SftpFileStream(SftpSessionMock.Object, _path, _fileMode, _fileAccess, _bufferSize);
+            _target.SetLength(_length);
         }
 
         protected override void Act()
         {
-            try
-            {
-                _target.Seek(_offset, SeekOrigin.End);
-                Assert.Fail();
-            }
-            catch (EndOfStreamException ex)
-            {
-                _actualException = ex;
-            }
+            _actual = _target.Seek(_offset, SeekOrigin.End);
         }
 
         [TestMethod]
-        public void SeekShouldHaveThrownEndOfStreamException()
+        public void SeekShouldHaveReturnedOffset()
         {
-            Assert.IsNotNull(_actualException);
-            Assert.IsNull(_actualException.InnerException);
-            Assert.AreEqual("Attempted to read past the end of the stream.", _actualException.Message);
+            Assert.AreEqual(_attributes.Size + _offset, _actual);
         }
 
         [TestMethod]
-        public void IsOpenOnSftpSessionShouldHaveBeenInvokedOnce()
+        public void IsOpenOnSftpSessionShouldHaveBeenInvokedTwice()
         {
-            SftpSessionMock.Verify(p => p.IsOpen, Times.Once);
+            SftpSessionMock.Verify(session => session.IsOpen, Times.Exactly(2));
         }
 
         [TestMethod]
-        public void PositionShouldReturnZero()
+        public void PositionShouldReturnOffset()
         {
-            SftpSessionMock.InSequence(MockSequence).Setup(p => p.IsOpen).Returns(true);
+            SftpSessionMock.InSequence(MockSequence).Setup(session => session.IsOpen).Returns(true);
 
-            Assert.AreEqual(0L, _target.Position);
+            Assert.AreEqual(_attributes.Size + _offset, _target.Position);
 
-            SftpSessionMock.Verify(p => p.IsOpen, Times.Exactly(2));
+            SftpSessionMock.Verify(session => session.IsOpen, Times.Exactly(3));
         }
     }
 }
--- a/src/Renci.SshNet.Tests/Classes/Sftp/SftpFileStreamTest_Seek_PositionedAtBeginningOfStream_OriginEndAndOffsetNegative.cs
+++ b/src/Renci.SshNet.Tests/Classes/Sftp/SftpFileStreamTest_Seek_PositionedAtBeginningOfStream_OriginEndAndOffsetNegative.cs
@@ -1,0 +1,102 @@
+﻿using System;
+using System.IO;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using Renci.SshNet.Sftp;
+
+namespace Renci.SshNet.Tests.Classes.Sftp
+{
+    [TestClass]
+    public class SftpFileStreamTest_Seek_PositionedAtBeginningOfStream_OriginEndAndOffsetNegative : SftpFileStreamTestBase
+    {
+        private Random _random;
+        private string _path;
+        private FileMode _fileMode;
+        private FileAccess _fileAccess;
+        private int _bufferSize;
+        private uint _readBufferSize;
+        private uint _writeBufferSize;
+        private byte[] _handle;
+        private SftpFileStream _target;
+        private int _offset;
+        private SftpFileAttributes _attributes;
+        private EndOfStreamException _actualException;
+
+        protected override void SetupData()
+        {
+            base.SetupData();
+
+            _random = new Random();
+            _path = _random.Next().ToString();
+            _fileMode = FileMode.OpenOrCreate;
+            _fileAccess = FileAccess.Read;
+            _bufferSize = _random.Next(5, 1000);
+            _readBufferSize = (uint)_random.Next(5, 1000);
+            _writeBufferSize = (uint)_random.Next(5, 1000);
+            _handle = GenerateRandom(_random.Next(1, 10), _random);
+            _offset = _random.Next(int.MinValue, -1);
+            _attributes = SftpFileAttributes.Empty;
+        }
+
+        protected override void SetupMocks()
+        {
+            SftpSessionMock.InSequence(MockSequence)
+                           .Setup(p => p.RequestOpen(_path, Flags.Read | Flags.CreateNewOrOpen, false))
+                           .Returns(_handle);
+            SftpSessionMock.InSequence(MockSequence)
+                           .Setup(p => p.CalculateOptimalReadLength((uint)_bufferSize))
+                           .Returns(_readBufferSize);
+            SftpSessionMock.InSequence(MockSequence)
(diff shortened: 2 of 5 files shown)


IgorMilavec (commented):
Good catch!

Can you also cover cases for invalid origin and trying to move before the beginning of the file? Here are the exceptions I feel would better describe the problems (and are somewhat aligned with the ```FileStream``` behavior):
```csharp
throw new ArgumentException("Invalid seek origin.", "origin")
throw new IOException("An attempt was made to move the file pointer before the beginning of the file.");
throw new EndOfStreamException("An attempt was made to move the file pointer past the end of the file.");
```

I see that you have changed formatting of the code. Based on my past experience the maintainer will request you to undo these changes.

Can you add tests for the cases you mentioned were not covered?

lemonyte:
> I see that you have changed formatting of the code.

I am not sure what formatting changes you are referring to. The only change other than replacing `-` with `+` was the removal of a duplicate identical code block, as below:
```cs
if (condition)
{
    // Condition specific code

    switch (origin)
    {
        // ...
    }

    if (newPosn == -1)
    {
        throw new EndOfStreamException("End of stream.");
    }
    _position = newPosn;
}
else
{
    // Condition specific code

    switch (origin)
    {
        // ...
    }

    if (newPosn < 0)
    {
        throw new EndOfStreamException();
    }
    _position = newPosn;
}
```

Refactored to:
```cs
if (condition)
{
    // Condition specific code
}
else
{
    // Condition specific code
}

switch (origin)
{
    // ...
}
if (newPosn < 0)
{
    throw new EndOfStreamException("End of stream.");
}
_position = newPosn;
```

```cs
throw new EndOfStreamException("An attempt was made to move the file pointer past the end of the file.");
```

I think a seek operation past the end of a file is a valid operation, but correct me if I am wrong. Most platforms allow seeking past the end of a file, only raising an exception when a read is attempted after EOF. A case for seeking before the beginning of a file already exists:
https://github.com/sshnet/SSH.NET/blob/7bdfc9e615b736b2762f5cd83c31a5f669663ff6/src/Renci.SshNet/Sftp/SftpFileStream.cs#L808
https://github.com/sshnet/SSH.NET/blob/7bdfc9e615b736b2762f5cd83c31a5f669663ff6/src/Renci.SshNet/Sftp/SftpFileStream.cs#L859

> Can you add tests for the cases you mentioned were not covered?

I am not very experienced in C#, being primarily a Python developer, but I'll see what I can reverse-engineer from the existing tests. Some guidance on how to use the available tests would be much appreciated.

IgorMilavec:
To run the tests in VS 2019, open Test Explorer and you will see a list of tests. Filter Traits (there is a filter icon at the column heading) and deselect "integration" and "LongRunning". Then run all the tests from Test Explorer.

To cover this scenario, you create new test classes ```SftpFileStreamTest_Seek_PositionedAtBeginningOfStream_OriginEndAndOffset*``` by copiing ```SftpFileStreamTest_Seek_PositionedAtBeginningOfStream_OriginBeginAndOffset*``` and adjusting appropriately.

Also, due to the changes in this PR, the ```SeekShouldHaveThrownEndOfStreamException``` test is failing, I will flag the probable cause for this in the code review.

src/Renci.SshNet/Sftp/SftpFileStream.cs
```diff
@@ -838,29 +818,28 @@ public override long Seek(long offset, SeekOrigin origin)
...
+                    case SeekOrigin.End:
+                        var attributes = _session.RequestFStat(_handle, false);
+                        newPosn = attributes.Size + offset;
+                        break;
+                }
 
-                    _position = newPosn;
+                if (newPosn < 0)
+                {
+                    throw new EndOfStreamException("End of stream.");
```
IgorMilavec: You are providing a non-default exception message, causing the test SeekShouldHaveThrownEndOfStreamException to fail. I beleive if you just use ```throw new EndOfStreamException();``` this will be fixed.

lemonyte: Thanks for pointing that out. The exception message is already present in another part of the code, so I assumed I should leave it in there.

IgorMilavec: @drieseng There have been suggestions in other issues that we should localize exception messages. While that is probably too much work at this time, I propose to use default exception messages where possible and to stop checking error messages in the tests and rely on only asserting exception type (and other structured info in the exception) in the tests. Do you agree?

lemonyte: I have noticed that the exception message format is different across .NET versions. For example:
.NET Framework 3.5
```
ArgumentException error message.
Parameter name: ParamName
```
.NET Core 3.0
```
ArgumentException error message. (Parameter 'ParamName')
```
This causes some tests to pass on one framework and fail on another.

IgorMilavec:
> Changed `SftpFileStream.Seek()` to add offset instead of subtracting when `origin` is set to `SeekOrigin.End`. (#909)

@LemonPi314 can you please edit the above to include "Fix #909" so this PR will be linked to the issue? Tnx!

.gitignore
```diff
@@ -22,3 +22,6 @@ project.lock.json
 
 # Build outputs
 build/target/
+
+# Visual Studio cache/options directory
+.vs/
```
drieseng: Please add leading forward slash since this ignore only applies to the root directory of the repo.

src/Renci.SshNet/Sftp/SftpFileStream.cs
```diff
@@ -838,29 +818,31 @@ public override long Seek(long offset, SeekOrigin origin)
...
+                        break;
+                    case SeekOrigin.Current:
+                        newPosn = _position + offset;
+                        break;
+                    case SeekOrigin.End:
+                        var attributes = _session.RequestFStat(_handle, false);
+                        newPosn = attributes.Size + offset;
+                        break;
+                    default:
+                        throw new ArgumentException("Invalid seek origin.", "origin");
```
daviburg: nit: I like to use named parameters when there is more than one argument with the same type, as the compiler won't catch out of order parameters and human reviewers may not know or remember the order of the arguments.

src/Renci.SshNet/Sftp/SftpFileStream.cs
```diff
@@ -838,29 +818,31 @@ public override long Seek(long offset, SeekOrigin origin)
...
+                }
 
-                    _position = newPosn;
+                if (newPosn < 0)
+                {
+                    throw new EndOfStreamException();
```
daviburg, line 842: nit: .NET reference source show use of [IOException](https://referencesource.microsoft.com/mscorlib/system/io/ioexception.cs.html#db370b64091ac4df)
at https://referencesource.microsoft.com/#mscorlib/system/io/memorystream.cs,1a4dcb744a23ba6f

```csharp
if (offset < 0 || tempPosition < _origin)
                    throw new IOException(Environment.GetResourceString("IO.IO_SeekBeforeBegin"));
```

Should SSH.NET use the same exception? EndOfStreamException for seeking before start feels off.

daviburg:
Test case for the error(s) been fixed is still needed in this PR.

daviburg:
> Can you also cover cases for invalid origin and trying to move before the beginning of the file? Here are the exceptions I feel would better describe the problems (and are somewhat aligned with the `FileStream` behavior):
> 
> ```cs
> throw new ArgumentException("Invalid seek origin.", "origin")
> throw new IOException("An attempt was made to move the file pointer before the beginning of the file.");
> throw new EndOfStreamException("An attempt was made to move the file pointer past the end of the file.");
> ```

I second this suggestion to use specific and descriptive exceptions.

daviburg:
> I see that you have changed formatting of the code. Based on my past experience the maintainer will request you to undo these changes.

I am ambivalent about that. If the formatting change is specific to the code been already updated for the fix, and covered by the tests, I find it a good opportunity to clean up.
If the formatting change is extensive and makes it difficult to understand the functional change, I can see how that may best be done in a separate PR.

src/Renci.SshNet/Sftp/SftpFileStream.cs
drieseng: @IgorMilavec, in general I prefer very specific asserts. To be honest, I think we have more important stuff to deal with than localization. But I'm ok to discuss this as a team, and decide where we want to go with this.

drieseng: @LemonPi314, that's indeed something I was already aware of and which I dealt with in other (professional) projects.

src/Renci.SshNet/Sftp/SftpFileStream.cs
drieseng: That's something I've been struggling with professionaly as well.

I prefer using named arguments only when the value is a literal, but even then you end with lots of line wrapping. Try creating a DateTime with the `DateTime (int year, int month, int day, int hour, int minute, int second, int millisecond, DateTimeKind kind)`. This is not the best example of course, as it's something you only tend to use in unit tests.

By the way, [Meziantou.Analyzer](https://github.com/meziantou/Meziantou.Analyzer) has some great rules amongst which [MA0003](https://github.com/meziantou/Meziantou.Analyzer/blob/main/docs/Rules/MA0003.md).

We should also use **nameof** everywhere once we've dropped support for the legacy TFMs.

src/Renci.SshNet/Sftp/SftpFileStream.cs
drieseng: I'm ok with **IOException**, but with useful error message. Please update the /// doc to mention that we also throw **IOException** for this scenario.

drieseng:
> I am ambivalent about that. If the formatting change is specific to the code been already updated for the fix, and covered by the tests, I find it a good opportunity to clean up. If the formatting change is extensive and makes it difficult to understand the functional change, I can see how that may best be done in a separate PR.

In general, I prefer a separate PR unless we're talking about a very minimal change.
To avoid discussions on what such a minimal change is, it may be better to not mix formatting and non-formatting changes at all.

src/Renci.SshNet/Sftp/SftpFileStream.cs
lemonyte: Resolved

src/Renci.SshNet/Sftp/SftpFileStream.cs
lemonyte: I agree, though this may be out of the scope of this PR as I would need to modify every other seek test as well.

lemonyte:
I've added tests for seeking from `SeekOrigin.End`.
Regarding formatting changes, to be frank I do not see what formatting exactly is causing an issue, but I will keep your feedback in mind for future contributions.
As for exceptions and exception messages, those are out of the scope of this PR, and would best be addressed separately in a new PR.

I do not want to pull this thread any more off-topic than it already is, but I think it is worth noting that while developing this PR I've encountered several issues with the tests. Notably, failures in tests unrelated to the changed portions (see AppVeyor checks [45419708](https://ci.appveyor.com/project/drieseng/ssh-net/builds/45419708/tests) and [45419751](https://ci.appveyor.com/project/drieseng/ssh-net/builds/45419751/tests)), and failures caused by exception messages not matching (see below and [this comment](https://github.com/sshnet/SSH.NET/pull/910#discussion_r808522963)).

```
 Ctor_PartialSuccessLimit_Zero
   Source: ClientAuthenticationTest.cs line 19
   Duration: < 1 ms

  Message: 
    Assert.AreEqual failed. Expected:<Cannot be less than one.
    Parameter name: partialSuccessLimit>. Actual:<Cannot be less than one. (Parameter 'partialSuccessLimit')>. 

  Stack Trace: 
    ClientAuthenticationTest.Ctor_PartialSuccessLimit_Zero() line 31
```

If there is anything else I can do to improve this PR, let me know. In any case, hopefully this fix is merged soon.

src/Renci.SshNet.Tests/Classes/Sftp/SftpFileStreamTest_Seek_PositionedAtBeginningOfStream_OriginEndAndOffsetNegative.cs
```diff
@@ -0,0 +1,102 @@
...
+            _writeBufferSize = (uint)_random.Next(5, 1000);
+            _handle = GenerateRandom(_random.Next(1, 10), _random);
+            _offset = _random.Next(int.MinValue, -1);
+            _attributes = SftpFileAttributes.Empty;
+        }
+
+        protected override void SetupMocks()
+        {
+            SftpSessionMock.InSequence(MockSequence)
+                           .Setup(p => p.RequestOpen(_path, Flags.Read | Flags.CreateNewOrOpen, false))
```
daviburg: Could the variable name be something more descriptive than 'p'?

src/Renci.SshNet.Tests/Classes/Sftp/SftpFileStreamTest_Seek_PositionedAtBeginningOfStream_OriginEndAndOffsetNegative.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            _target = new SftpFileStream(SftpSessionMock.Object, _path, _fileMode, _fileAccess, _bufferSize);
+        }
+
+        protected override void Act()
+        {
+            try
+            {
+                _target.Seek(_offset, SeekOrigin.End);
+                Assert.Fail();
```
daviburg: consider using Assert.ThrowsException<EndOfStreamException>(() => { ... });

https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.assert.throwsexception?view=visualstudiosdk-2022

What you can't do with this built-in is verify the message of the exception - that still needs a try-catch with message validation in the catch (or your capture of the exception). You can then do a re-throw in the catch to keep satisfying the assert expectation.

daviburg (approved):

src/Renci.SshNet.Tests/Classes/Sftp/SftpFileStreamTest_Seek_PositionedAtBeginningOfStream_OriginEndAndOffsetNegative.cs
lemonyte: Renamed to `session`.

src/Renci.SshNet.Tests/Classes/Sftp/SftpFileStreamTest_Seek_PositionedAtBeginningOfStream_OriginEndAndOffsetNegative.cs
lemonyte: Turns out the exception check was not necessary.

drieseng:
@LemonPi314, your changes are looking good. Thanks!

I'd only like to hold off on merging until I (or someone else) had time to add a few integration tests for this.
Should this take too long, do not hesitate to spam me :-)

drieseng:
I'll submit a PR for the integration test changes once https://github.com/sshnet/IntegrationTests/pull/8 has landed.

drieseng (approved):

PR #914, opened 2022-01-22
Feature/timestamp sync

kendallb:
Use both file length and timestamp difference to sync files


kendallb:
Ugh, I realized when debugging this some more, we need to set the last write time for the file on the remote server after it is uploaded, however it appears that is not currently supported by the library? Are there any plans to add that support?

PR #918, opened 2022-02-14
Fix SocketAsyncEventArgsAwaitable calling continuation multiple times

### LINKED ISSUE #913: System.InvalidOperationException: Renci.SshNet.Abstractions.SocketExtensions+_ConnectAsync_d__1.MoveNext
We are seeing crashes when an async connect operation is canceled. Analyses of the crash leads me to believe that SocketAsyncEventArgsAwaitable.SetCompleted is being called twice: once during the handling of the cancel and once when the IOCompletion fires. However the code that runs the continuationAction does not handle this condition appropriately and runs the continuationAction twice, causing the InvalidOperationException.  Changing this line in SetCompleted:
```
var continuation = continuationAction ?? Interlocked.CompareExchange(ref continuationAction, SENTINEL, null);
```
to:
```
var continuation = Interlocked.Exchange(ref continuationAction, SENTINEL);
```
seems to fix the problem.

IgorMilavec:
Is some cases ```SocketAsyncEventArgsAwaitable``` was calling continuations multiple times due to improper state transition.
This will fix #913


IgorMilavec:
This has been running in my test environment for a couple of days and it seems there are no regressions.

dlangell:
What are the chances of getting this merged into the main development branch?  I haven't personally tested it, but it would be helpful if the fix actually works.  The ConnectAsync function is basically unusable without it since it can cause an unhandled exception.  I've had to stick with the Synchronous Connect instead because I was seeing the same issue noted in #913.  Cancelling the token while waiting for ConnectAsync to return for a device that is not reachable will cause the exception.

Rob-Hague (approved):

PR #922, opened 2022-02-19
Optimize AppVeyor network stack for number of dynamic connections

### LINKED ISSUE #921: AppVeyor test runs fail with "Only one usage of each socket address is normally permitted"
Sometimes AppVeyor test runs fail with ```System.Net.Sockets.SocketException: Only one usage of each socket address (protocol/network address/port) is normally permitted.```. After one occurrence of this exception, most consequent tests fail with the same exception.
A typical example of such behavior is [this build](https://ci.appveyor.com/project/drieseng/ssh-net/builds/40252697?fullLog=true) of #855.

IgorMilavec:
Many internet sources indicate that the root cause for ```System.Net.Sockets.SocketException: Only one usage of each socket address (protocol/network address/port) is normally permitted.``` is port exhaustion. In case of SSH.NET tests, this could be caused by a large number of connections opened and closed in quick succession. This was already discussed in https://github.com/sshnet/SSH.NET/pull/877#issuecomment-923982783.

In this PR I have both increased the number of ports that can be used for client side connections and decreased the time that the network stack needs to wait before reusing bindings. More information on these settings [here](https://docs.microsoft.com/en-us/biztalk/technical-guides/settings-that-can-be-modified-to-improve-network-performance).
As this only affects the AppVeyor build VM, I do not expect any regressions or side effects.

This will (hopefully) fix #921.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,9 @@
 os: Visual Studio 2019
 
+init:
+ - netsh int ipv4 set dynamicport tcp start=1025 num=64510
+ - powershell "New-ItemProperty -Path 'HKLM:\SYSTEM\CurrentControlSet\Services\Tcpip\Parameters' -Name 'TcpTimedWaitDelay' -Value '15' -PropertyType 'DWord'"
+
 before_build:
   - nuget restore src\Renci.SshNet.VS2019.sln
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,7 +2,6 @@ os: Visual Studio 2019
 
 init:
  - netsh int ipv4 set dynamicport tcp start=1025 num=64510
- - powershell "New-ItemProperty -Path 'HKLM:\SYSTEM\CurrentControlSet\Services\Tcpip\Parameters' -Name 'TcpTimedWaitDelay' -Value '15' -PropertyType 'DWord'"
 
 before_build:
   - nuget restore src\Renci.SshNet.VS2019.sln


appveyor.yml
```diff
@@ -1,5 +1,9 @@
 os: Visual Studio 2019
 
+init:
+ - netsh int ipv4 set dynamicport tcp start=1025 num=64510
+ - powershell "New-ItemProperty -Path 'HKLM:\SYSTEM\CurrentControlSet\Services\Tcpip\Parameters' -Name 'TcpTimedWaitDelay' -Value '15' -PropertyType 'DWord'"
```
zybexXL: This setting requires this other key to be set or it will have no effect: ([Source](https://support.solarwinds.com/SuccessCenter/s/article/Tweaking-Windows-Server-performance-to-prevent-port-exhaustion-when-using-Orion-modules?language=en_US#:~:text=NOTE%3A%C2%A0You%C2%A0must%C2%A0set%20the%20StrictTimeWaitSeqCheck%20to%201%2C%20or%20the%20TcpTimedWaitDelay%20value%20will%20have%20no%20effect%20on%20the%20system.))

```
[HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\Tcpip\Parameters]
"StrictTimeWaitSeqCheck"=dword:00000001
```

It also only takes effect after reboot, so if the AppVeyor VM reverts to base settings on each reboot this will also have no effect.

Third, both commands require Elevation. Does AppVeyor execute this as Admin? I find it unlikely.

IgorMilavec: Yes, AppVeyor build VM is started fresh from image. So there is no point in setting this registry setting... do you know of any other method for setting wait delay that would be effective immediately?

zybexXL: Perhaps we can set the per-socket timeout when running the test Suite?

```
// sets the amount of time to linger after closing, using the LingerOption public property.
LingerOption lingerOption = new LingerOption (true, 10);
```

SO_REUSEPORT or SO_REUSEADDR seem to be more appropriate, but it looks like Windows doesn't really care about these flags.

IgorMilavec: Cool, this would work. But I think this amount of changes is beyond this PR and should be discussed further. For now I'll just  remove the wait delay change as is is ineffective and leave the enlargement of client port pool in this PR.

appveyor.yml
```diff
@@ -1,5 +1,8 @@
 os: Visual Studio 2019
 
+init:
+ - netsh int ipv4 set dynamicport tcp start=1025 num=64510
```
zybexXL, line 4: "The requested operation requires elevation (Run as administrator)."

Will AppVeyor execute this as admin?

IgorMilavec: This is OK, AppVeyor does not utilize UAC. AppVeyor log for this PR shows that the command is executed successfully.

Issue #917, opened 2022-02-14
Cannot access a disposed object (ShellStream)


PatrickSchmidtSE:
I have this weird behaviour , only happening with one device ( and only once a day).

"Cannot access a disposed object (ShellStream)"
at Renci.SshNet.ShellStream.Write(String text)

What struggles me is that IsConnected is showing true AND no errorOccured Event was thrown ( so i know the connection was suddenly interupted).

i use just this: 
SshCommand cmdRes = client.RunCommand(cmd);

ShellStream is created via this call
stream = client.CreateShellStream("dumb", 10000, 24, 800, 10000, 4086);

The dispose must happen inside the Renci part. The question is, how to find out why the ShellStream is disposed. 
And if i get this error, how can i react? Just create a new ShellStream and re-run the command?

IgorMilavec:
I have created a discussion #926 for all ShellStream related issues.

jwlondon98:
Any update on this? I can't seem to get past this issue.

Issue #833, opened 2021-05-30
Return matched group in ShellStream.Expect


NN---:
Currently it returns the strings which requires me to call Regex.Match second time.

IgorMilavec:
I have created a discussion #926 for all ShellStream related issues.

IgorMilavec:
@NN I'm trying to better understand usage patterns related to Expect API. Can you please explain how you use it and if possible provide a proto code for it?

NN---:
The API reads the data and tries to
match it with the given pattern.
Once it is matched the method returns.
The problem here is that I want the matched groups what API doesn't give.

If you talk generally about the API, it would be better to have a Stream abstraction over SSH, then such code could be implemented for any stream.

IgorMilavec:
Yes, I know what the Expect API is supposed to do. :)
What I'm asking is how you are using it? I agree that having such functionality for any Stream would be beneficial, but to design it properly, we first need to understand how it is used in the client code. So if you could describe why and how you are using it would help me very much.

NN---:
One of usages is to wait for password prompt when `sudo` is called and put a password.

IgorMilavec:
So in this case you would be Expecting ```$".*\[sudo\] password for {RegexEscape(username)}: "```?

NN---:
Yes, this is the idea.
It is one of the possible usages of Expect of course.

NN---:
In general any command requiring input needs to wait for the prompt and only then put the text.

jscarle:
@IgorMilavec @WojciechNagorski Although the conversation is unclear, I believe this was also fixed by #1207 and #1319.

Issue #672, opened 2020-05-02
Close of channel should not dispose ShellStream


drieseng:
We currently dispose the **ShellStream** when the underlying channel is closed.
This can lead to a **NullReferenceException** in, for example, `ShellStream.Expect(TimeSpan timeout, params ExpectAction[] expectActions)`.

In case of `ReadLine(TimeSpan timeout)`, we should return **null** when the channel has been closed.

WojciechNagorski:
This issue has been fixed in the 2024.0.0 version.

Issue #670, opened 2020-04-30
Expect doesn't wake up after disconnect


sterlind:
If the connection is lost while we're waiting in `ShellStream.Expect()`, Expect() doesn't wake up, either hanging forever or until the timeout elapses.

Here's a simple repro. I configured the target server (a Cisco NCS-5K) to end sessions that have been idle for 30 seconds, and then I set an expect that never matches. The connection dies 30 seconds in, but the Supercalifragilisticexpialidocious Expect waits 5 excruciating minutes to return:
```csharp
            var ssh = new SshClient(info);
            ssh.Connect();
            using (var shell = ssh.CreateShellStream("xterm", 80, 80, 80, 80, 1024))
            {
                shell.Expect("#", TimeSpan.FromSeconds(30));
                Trace.TraceInformation("Shell has started!");
                var time = Stopwatch.StartNew();
                var result = shell.Expect("Supercalifragilisticexpialidocious!", TimeSpan.FromMinutes(5));
                Trace.TraceInformation($"Shell took {time.Elapsed} to time out..");
            }
```

Expect() should return immediately upon disconnect.

WojciechNagorski:
It's done in https://github.com/sshnet/SSH.NET/pull/1274

WojciechNagorski:
The 2023.0.1 version has been released to Nuget: https://www.nuget.org/packages/SSH.NET/2023.0.1

PR #923, opened 2022-02-22
Fix ShellStream.Read returning 0 prematurely

### LINKED ISSUE #63: ShellStream.Read not blocking when no data available
ShellStream.Read returns immediately with 0 bytes read when no data is available instead of blocking the call.
Consequently the other functions implemented in the parent class `Stream` that implement the asynchronous patterns based on the Read function (BeginRead/EndRead, ReadAsync) return also immediately with a 0 byte read instead of being suspended until data is available.
This prevent using correct asynchronous pattern with these functions.

The following modified function waiting on `_dataReceived` when no data is available seems to work:

``` c#
public override int Read(byte[] buffer, int offset, int count)
{

    var i = 0;
    while (true)
    {
        lock (_incoming)
        {
            for (; i < count && _incoming.Count > 0; i++)
            {
                buffer[offset + i] = _incoming.Dequeue();
            }
        }

        if (i != 0) 
            return i;

        _dataReceived.WaitOne();
    }
}
```

IgorMilavec:
Fixes ```ShellStream.Read``` to block  until more data is received or the session is closed.
This will fix #63

### DIFF
--- a/src/Renci.SshNet/ShellStream.cs
+++ b/src/Renci.SshNet/ShellStream.cs
@@ -215,17 +215,31 @@ namespace Renci.SshNet
         /// <exception cref="T:System.ObjectDisposedException">Methods were called after the stream was closed.</exception>
         public override int Read(byte[] buffer, int offset, int count)
         {
-            var i = 0;
-
-            lock (_incoming)
+            if (_isDisposed)
             {
-                for (; i < count && _incoming.Count > 0; i++)
-                {
-                    buffer[offset + i] = _incoming.Dequeue();
-                }
+                throw new ObjectDisposedException("ShellStream");
             }
 
-            return i;
+            var i = 0;
+            while (true)
+            {
+                lock (_incoming)
+                {
+                    for (; i < count && _incoming.Count > 0; i++)
+                    {
+                        buffer[offset + i] = _incoming.Dequeue();
+                    }
+                }
+
+                if (i != 0)
+                    return i;
+
+                _dataReceived.WaitOne();
+                if (_incoming.Count == 0)
+                {
+                    return 0; // The session was closed
+                }
+            }
         }
 
         /// <summary>
@@ -766,6 +780,10 @@ namespace Renci.SshNet
 
         private void Session_Disconnected(object sender, EventArgs e)
         {
+            // Release thread blocked in a ShellStream.Read() call
+            if (_dataReceived != null)
+                _dataReceived.Set();
+
             if (_channel != null)
                 _channel.Dispose();
         }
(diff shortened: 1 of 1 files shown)


IgorMilavec:
@drieseng , I also have a fix ready for #672, which is based on this PR. Should I amend this PR or open a new one? I think the cleanest option would be if you review and merge this PR and I rebase and open a new one for #672. (edit: and #453, #670, #917, #833, #78, #140, #383)

drieseng:
@IgorMilavec This will block indefinitely if the last "line" does not end with a line ending. I have an integration test locally that allows you to reproduce this, but I'd really like you to have a go at this too. Let me know if you need any help or guidance.

IgorMilavec:
```ShellStream.Read()``` should be agnostic about any content, I don't see how it could fail in this way? Do you have an example so I can test it?

drieseng:
A few other remarks:
* `Read()` still returns immediately.
* Do we want a **ReadTimeout** setting (like you find on **NetworkStream**)?

drieseng:
@IgorMilavec Yes, I have a test for this but - as I said - I'd really prefer you also take a stab at this hereby hoping to convince you of the added value of these integration tests.

drieseng:
It reproduces when you use a **StreamReader** to read content of a **ShellStream** that is closed while reading.

IgorMilavec:
Yes, I am aware that there are issues with closing the stream. Currently this PR only fixes #63. As I said, I already have other fixes ready, but I need your decision whether to add these changes to this PR?

drieseng:
@IgorMilavec We shouldn't merge this without a minimal fix that avoids a "hang" when the channel is closed, and corresponding tests.

IgorMilavec:
@drieseng I have committed the changes required to handle session/channel closing. There were quite a bit of changes, mostly due to the fact that all Read* and Expect* methods directly handle internal state. Can you please have a look?

IgorMilavec:
```ShellStream``` has a race condition where only ```_incoming``` is locked and ```_dataReceived``` is not. Really solving this would require a substantial rewrite of inner workings. I tried to at least make the current implementation usable in https://github.com/sshnet/SSH.NET/pull/923/commits/b4ffc604fbbda899e26ef7b2151a34d7983e63cc, but introduced a regression in Expect (timeout never expired).  This should now be fixed.

glorialearnscoding:
Hello @IgorMilavec @drieseng , any plan to publish new version with bugs mentioned above fixed?

TomBartDrone:
I repeat the @glorialearnscoding question, do you plan to publish fixes for that issues?

jscarle:
@jscarle Note to self.

@IgorMilavec I'd like to understand what your original plan was with this PR as ShellStream has changed enough that the diff is quite messy and I'd like to take a look at addressing this.

PR #924, opened 2022-02-23
Fix PipeStream.Read returning 0 prematurely 

### LINKED ISSUE #12: PipeStream returns 0 instead of blocking when running a command with intermittent output

### LINKED ISSUE #164: Cannot get streaming stdout/stderr and wait for remote process to exit
My scenario is that I want to start a process on a remote machine and wait for it to either crash, or exit gracefully. While I am waiting for it to crash/exit I want to propagate it's stdout/stderr to the host machine so I can parse it, or redirect it to file once I get the data I need from it.

My attempt is: https://gist.github.com/alanmcgovern/c12867f3accc6cf5870d5212d3425974#file-test-cs-L8

The issue with this approach is that the very first `(line = reader.ReadLine ())` returns null if you do it 'too quickly'. PipeStream thinks the stream has completed even though the process is still running on the remote machine. If i put a Thread.Sleep before that line I can read some output before i get a null, but eventually i get a null when reading, even though the process hasn't exited.

The only 'reliable' approach to detect when my process exits is to create a ShellStream, then write this command: `path/to/long_lived_application; echo TOMBSTONE;` and then exit my read-loop when i see the string `TOMBSTONE` appear in the output.

This approach is here: https://gist.github.com/alanmcgovern/c7bf0fab9760cfcd324573df51c98fbb

Is there a better way to asynchronously run a command and gather it's stdout/stderr?

### LINKED ISSUE #440: PipeStream not supporting Read offsets > 0
PipeStream seems to throw an exception when Read(byte[] buffer, int offset, int count) is called, with an offset above zero. 
https://github.com/sshnet/SSH.NET/blob/master/src/Renci.SshNet/Common/PipeStream.cs#L183

Is there a reason it cannot support this?  Seems like it would be easy enough to modify the for loop to start at the desired offset, and still calculate the amount read properly.

If it's just something that hasn't been done yet, I can perhaps do the work, but I wanted to check first to see if there's some edge case that makes this feature N/A that I'm not seeing.

Thanks!

### LINKED ISSUE #650: SshCommand: OutputStream & ExtendedOutputStream should only return 0 when channel is closed
Invoking `Read(Byte[], Int32, Int32)` on SshCommand*'s **OutputStream** or **ExtendedOutputStream** should blocking until either data is available, or the channel is closed (in which case zero should be returned).

We should strongly consider applying breaking changes to the *SshCommand* API as well:
* Change signature of `public string Execute()` to `public void Execute()`.
* Have *ExitCode* throw when we have not received an *ExitStatusRequestInfo*.
* Remove the **Result** property (users should use **OutputStream** instead).
* Remove the **Error** property (users should use **ExtendedOutputStream** instead).

IgorMilavec:
Fixes PipeStream.Read to block until more data is received or the session is closed.
This will fix #12, fix #164, fix #440, fix #650

### DIFF
--- a/src/Renci.SshNet.Tests/Classes/Common/PipeStreamTest.cs
+++ b/src/Renci.SshNet.Tests/Classes/Common/PipeStreamTest.cs
@@ -63,7 +63,7 @@ namespace Renci.SshNet.Tests.Classes.Common
 
             var readBuffer = new byte[2];
             var bytesRead = target.Read(readBuffer, 0, readBuffer.Length);
-            Assert.AreEqual(2, bytesRead);
+            Assert.AreEqual(readBuffer.Length, bytesRead);
             Assert.AreEqual(0x0a, readBuffer[0]);
             Assert.AreEqual(0x0d, readBuffer[1]);
 
@@ -78,8 +78,18 @@ namespace Renci.SshNet.Tests.Classes.Common
 
             target.BlockLastReadBuffer = true;
             readBuffer = new byte[2];
-            bytesRead = target.Read(readBuffer, 0, readBuffer.Length);
-            Assert.AreEqual(2, bytesRead);
+            int bufferOffset = 0;
+            while (bufferOffset < readBuffer.Length)
+            {
+                bytesRead = target.Read(readBuffer, bufferOffset, readBuffer.Length -  bufferOffset);
+                if (bytesRead == 0)
+                {
+                    Assert.Fail("PipeStream.Read returned 0");
+                }
+
+                bufferOffset += bytesRead;
+            }
+            Assert.AreEqual(readBuffer.Length, bufferOffset);
             Assert.AreEqual(0x09, readBuffer[0]);
             Assert.AreEqual(0x05, readBuffer[1]);
         }
--- a/src/Renci.SshNet.Tests/Classes/Common/PipeStreamTest.cs
+++ b/src/Renci.SshNet.Tests/Classes/Common/PipeStreamTest.cs
@@ -76,6 +76,7 @@ namespace Renci.SshNet.Tests.Classes.Common
                     });
             writeToStreamThread.Start();
 
+            target.BlockLastReadBuffer = true;
             readBuffer = new byte[2];
             bytesRead = target.Read(readBuffer, 0, readBuffer.Length);
             Assert.AreEqual(2, bytesRead);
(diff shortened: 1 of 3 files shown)


drieseng:
@IgorMilavec I definitely want this change in, but there are still a few things to consider:
- I don't think we deal with the situation where a connection or channel is closed while performing a blocking read.
- Do we want to allow a (read) timeout to be configured?
- Do we want to remove the *BlockLastReadBuffer* property?
- I'd like a few integration tests that cover this. Should be pretty easy to add some for **SshCommand**.
  Perhaps a good way to familiarize yourself with (and/or criticize) the work I did in the **IntegrationTests* repository.

IgorMilavec:
```PipeStream``` is used to communicate between two threads and is not related to a connection or channel. I don't even know why we need it, I just wanted to fix the issue reported. :)
```BlockLastReadBuffer``` is definitely something I don't like. The way this is implemented interferes with detecting the end of stream. For "read the length I need and only return less data is end of stream is reached" functionality I propose this extension:
```csharp
public static int ReadBlock(this Stream stream, byte[] buffer, int offset, int count)
{
    #region Validate arguments
    if (stream is null)
    {
        throw new ArgumentNullException(nameof(stream));
    }
    #endregion

    int bufferLength = 0;
    while (bufferLength < count)
    {
        int readLength = stream.Read(buffer, offset + bufferLength, count - bufferLength);
        if (readLength == 0)
        {
            break;
        }
        bufferLength += readLength;
    }

    return bufferLength;
}
```
Works for every stream, I also have an async version...

But generally, ```PipeStream``` is so terribly inefficient and complex that I propose to rewrite it completely. At this time I propose to only fix the most pressing bugs, release the final .NET 3.5 version and then refactor/rewrite such code when we only have to support modern frameworks.

IgorMilavec:
Just found out that ```ReadBlock()``` and ```ReadBlockAsync()``` will be added in dotnet 7: https://docs.microsoft.com/en-us/dotnet/api/system.io.streamreader.readblock?view=net-6.0.
So we can easily "pollyfill" them for earlier versions and use dotnet implementation from net7.0 onwards.
@drieseng should I add them in this PR and mark ```BlockLastReadBuffer``` deprecated?

I checked the integration tests and ```Ssh_Command_IntermittendOutput_OutputStream``` which is now ignored should now pass.

Issue #931, opened 2022-03-11
Add async support for Connect() function?


kendallb:
Connecting to an SSH server is quite time consuming with all the key exchanges etc. I did not find an async version of that function, so was wondering if it was possible to add an async version of that (old school async is fine, as we can then wrap it for task based async). Or is there some reason that function is not async?

For the moment when I need to open multiple connections I do them in separate task threads, so I can run them in parallel.

IgorMilavec:
Async support was just added for SFTP, SSH part didn't get the love yet. :)
There is no particular technical reason, it just hasn't been done yet.

kendallb:
Ok good to know

IgorMilavec:
Just had another look and ```ConnectAsync(CancellationToken cancellationToken)``` is implemented on the ```BaseClient``` from which ```SshClient``` is derived.
So if you only need this, it is already available on the development branch. No NuGet package yet, so you'll have to build SSH.NET yourself.

kendallb:
That is good to know. I can wait for the next NuGet build though, as I already solved our problem simply by running the connection code in a separate task. Once this is public, I can remove that functionality and use the async versions.

kendallb:
Hopefully we can get task based async stuff added at the same time so we can avoid using a third party library like this one:

https://github.com/JohnTheGr8/Renci.SshNet.Async

Maybe I should take a look at the latest code and do a PR to integrate it.

IgorMilavec:
That would be great. Please have a look at how it is implemented in ```SftpClient```. The above mentioned extension library uses TAP over APM approach. Unfortunately the current APM implementation in SSH.NET uses ThreadPool to queue "async" requests, which creates even more pressure that running caller code on ThreadPool. Ideally in the future we could just remove the APM or replace it with APM over TAP.

I expect that at least these should be added at this time:
 - SshCommand.ExecuteAsync()
 - SshClient.RunCommandAsync()
 - Shell.StartAsync()
 - Shell.StopAsync()

Do you see anything else? 

ShellStream is problematic in "many, many wonderfully ways", please see #926. I think it should be refactored before adding any async functionality.

kendallb:
That looks like about it for me. I don't use ShellStream (need to research what that is - could be useful!) but yes, checking that problem reports for that I suggest we leave that alone for the moment.

kendallb:
What is the process of getting all the tests to run. Do I need to set up a test SSH server someone and setup the credentials to access it?

[...]

PR #942, opened 2022-03-25
Add SshCommand.ExecuteAsync()

IgorMilavec:
This is my attempt at #931, reusing some code from #937.

```Channel.Close()``` was modified to always trigger ```Channel.Closed``` event, allowing us to detect that the channel was closed locally.

The ```SshCommand.ExecuteAsync()``` aborts the initial task in case cancellation is requested:
 - Is sends a SIGTERM signal to the server to try to abort the command run at the server
 - It closes the ```Channel```
 - It notifies the caller about cancellation

Because this code uses TAP over APM and ```CommandAsyncResult``` has no means to convey any exceptions thrown during execution, there are some workarounds to properly notify the caller that the task was cancelled.

@kendallb can you have a look at this and check if it works for you?

### DIFF
--- a/src/Renci.SshNet/Channels/Channel.cs
+++ b/src/Renci.SshNet/Channels/Channel.cs
@@ -562,14 +562,11 @@ namespace Renci.SshNet.Channels
                     // this also ensures don't raise the Closed event more than once
                     IsOpen = false;
 
-                    if (_closeMessageReceived)
+                    // raise event signaling that both ends of the channel have been closed
+                    var closed = Closed;
+                    if (closed != null)
                     {
-                        // raise event signaling that both ends of the channel have been closed
-                        var closed = Closed;
-                        if (closed != null)
-                        {
-                            closed(this, new ChannelEventArgs(LocalChannelNumber));
-                        }
+                        closed(this, new ChannelEventArgs(LocalChannelNumber));
                     }
                 }
             }
--- a/src/Renci.SshNet/SshCommand.cs
+++ b/src/Renci.SshNet/SshCommand.cs
@@ -8,6 +8,9 @@ using Renci.SshNet.Messages.Connection;
 using Renci.SshNet.Messages.Transport;
 using System.Globalization;
 using Renci.SshNet.Abstractions;
+#if FEATURE_TAP
+using System.Threading.Tasks;
+#endif
 
 namespace Renci.SshNet
 {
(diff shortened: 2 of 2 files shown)


kendallb:
Sure thing I will take a look.

kendallb:
I commented in your commit, but I will bring it in here for visibility. Your code works with one minor change. You need to run BeginExecute prior to setting up the cancellation token as actually starting the request can take a while (more than 100ms in my case) so when the token fires there is no channel to cancel and you get an NRE. Granted my cancellation is very short, but if the SSH server on the other end is taking a while to respond a timeout could theoretically trigger before that.

I patched your changes into my branch so I could run my test and now it passes. Those changes are pushed to my PR.

I did leave EndExecuteWithStatus as public in my version, as I think that is a useful function for others to use (does not consume the result stream)?

Also BeginExecute is not async. I think we should consider making that async as well as it does open the channel and send commands to it, and all of those could easily time out also (and hence the reason my test with a 100ms timeout failed without running it prior to setting up the cancelation token callback).

kendallb:
Also if we actually did make async connection open and execute, then we probably do not need to use a TAP wrapper at all over APM for this? How much work is it to make async versions of channel open and execute?

IgorMilavec:
Good catch with BeginExecute! :)
Sure, ideally we should make ExecuteAsync be async all the way down. But this is a lot of effort, especially at this point when we still need to support .NET 3.5 and 4.0. I think we have achieved quite a lot for now: we have defined a good API that we won't need to change if we change implementation details and conserved a thread in the process (apart from first 100ms). After the next release, the plan is to abandon support for .NET 3.5 and 4.0 and we can revisit ExecuteAsync then, OK?

kendallb:
Yep, sounds good. I did take a look at the internals and yeah, it's a lot of work to make that async all the way :) So this is a good first step and the connection part is not something that takes that long, and the main point of the async version of execute or for longer running commands.

So I think this is a good first step, especially if the fixes for the PipeStream stuff is also done.

We do need to also decide on what to do with .Result and .Error. Deprecate them or leave them as is.

IgorMilavec:
Regarding Result and Error: Gert already commented that he agrees to mark them deprecated. Can you please add that to your PR? Maybe provide a direct guideline on how to replace them in the Obsolete message?
What we still need to discuss with Gert is the accessibility / signature of EndExecute. 
I'll close this draft now and we can continue on your PR, OK?

PR #946, opened 2022-04-03
Fix netconf framing protocol

schaveyt:
## Problem

The implementation was not compliant to the RFC6242. Although the client does NOT support the framing protocol, if the server reported that is did, the client would still try to using the framing protocol. The result is that client and server are unable to communicate at the RPC XML layer and the session times out.

## Solution

The solution is to simply never used framing protocol regardless of whether the server does or not.

## Additional Minor Changes

1. Address null pointer exception when the netconf_session is disposing.
1. Add an example of the RPC XML to use when transmitting calls.
 

## Experimental Release for .NET Standard 2.1

https://github.com/schaveyt/SSH.NET/releases/tag/2021.04.02


Rob-Hague:
I think we should take this one on good faith. I have no knowledge of netconf and we have no testing but it seems reasonable to me, and with #639 and #1197 there are at least 3 people for whom it works.

ddombrowsky:
It appears that at least for Cisco WLCs, the implementation of the framing protocol is indeed broken.  I need to hardcode `_usingFramingProtocol = false;` to get `SendReceiveRpc` to do anything but timeout.

@WojciechNagorski  Please merge this PR.

Rob-Hague (approved):

Rob-Hague:
I took #639 and the remaining comments from this one

Issue #859, opened 2021-08-13
Accessing a hash algorithm by manipulating the HashName property is not supported on this platform.


alvaromolinaGL:
Hi, I am using the library to send some files through an sftp, I am connecting to the site via the SftpClient class:

var client = new SftpClient(host, port, username, password);

client.Connect();
client.UploadFile(zipFile, outPath);
client.Disconnect();

This code is running in an azure function, it was working fine but it stopped working back in May and its now throwing this message:

Accessing a hash algorithm by manipulating the HashName property is not supported on this platform. Instead, you must instantiate one of the supplied subtypes (such as HMACSHA1.)

Any  ideas??

Thanks.

IgorMilavec:
Can you provide the call stack of the exception?

dd-morphi:
Hi, 
Recently I've hit an issue with the same error message but in a different scenario.
My code is connecting to devices through SSH. One of those devices is an older Linux.
When connecting to it on Windows with .net5 it throws the exception:
```
System.PlatformNotSupportedException: Accessing a hash algorithm by manipulating the HashName property is not supported on this platform. Instead, you must instantiate one of the supplied subtypes (such as HMACSHA1.)
   at System.Security.Cryptography.HMAC.HashFinal()
   at Renci.SshNet.Security.Cryptography.HMACSHA1.HashFinal()
   at System.Security.Cryptography.HashAlgorithm.CaptureHashCodeAndReinitialize()
   at System.Security.Cryptography.HashAlgorithm.ComputeHash(Byte[] buffer)
   at Renci.SshNet.Session.SendMessage(Message message)
   at Renci.SshNet.Session.Connect()
   at Renci.SshNet.BaseClient.CreateAndConnectSession()
   at Renci.SshNet.BaseClient.Connect()
```
I was trying to debug your library and discovered that the issue is on this device because it returns the only hash algorithm `hmac-sha1-96`. It looks that your library on `netstandard2.0` has an issue with this. When tried with `framework` it works.

I reproduced the issue with Ubuntu by adding the line `MACs hmac-sha1-96` to `/etc/ssh/sshd_config` and restarting the sshd service. 

The easiest **workaround** for it is removing `FEATURE_HMAC_SHA1` from the list of `DefineConstants` and then rebuilding it. It uses then SshNet.Security.Cryptography which works correctly in this scenario.

---------
[...]

PR #947, opened 2022-04-05
Change HMACSHA1 implementation to external library

dd-morphi:
This is a workaround for https://github.com/sshnet/SSH.NET/issues/859 as described inside. It lasts for more than half a year and it will be good to have it in production.


dd-morphi:
Is there a possibility to merge it?

noahmehl:
+1 from us on this one

Rob-Hague:
Thanks for the contribution (and the initial diagnosis). Closing this as the underlying issue is now fixed in 2023.0.0 (by virtue of updated target framework support).

Issue #956, opened 2022-05-04
Permission denied (publickey) with SSH.NET - Can connect via FileZilla or Powershell.


wfrx:
I'm trying to extend our SFTP capability to allow connection via key-authentication.

The code examples are very straightforward (thanks for that) and I'm pretty sure I've done it right:
`connectionInfo = New ConnectionInfo(e.Host, e.Port, e.Username, GetAuthenticationMethods(e))`
with GetAuthenticationMethods(e) giving me an array of AuthenticationMethods, depending on the provided information:

```
[...]
authMethod.Add(New PasswordAuthenticationMethod(e.Username, e.Password))
[...]
If e.FilePassphrase = "" Then
  authMethod.Add(New PrivateKeyAuthenticationMethod(e.Username, New PrivateKeyFile(keyFile)))
ElseIf e.FilePassphrase <> "" Then
  authMethod.Add(New PrivateKeyAuthenticationMethod(e.Username, New PrivateKeyFile(keyFile, e.FilePassphrase)))
End If
[...]
m_Client = New SftpClient(connectionInfo)
m_Client.Connect()
```

I have a Server set up and generated keys. Originally, I had a ssh-rsa key, giving me a `openssh key type: ssh-rsa is not supported` Error while trying to instantiate the PrivateKeyAuthenticationMethod. (afaict this should be supported but is not yet in the NuGet package, is there any ETA on when this will be available?)

I converted this key to PEM, which gets me one step further but now I get a  `Permission denied (publickey).` Error. 

I can connect to the server without issues using either the PEM or the ssh-rsa key over FileZilla or ssh on console.

gregolsky:
We are experiencing this issue when using latest Ubuntu release Jammy Jellyfish. 
@wfrx Are you using Ubuntu 22.04?

Related to `ssh-rsa` deprecation?
https://discourse.ubuntu.com/t/jammy-jellyfish-release-notes/24668
https://bugs.launchpad.net/ubuntu/+source/openssh/+bug/1961833

wfrx:
> @wfrx Are you using Ubuntu 22.04?

Yes, but as I wrote I can connect via ssh on console:

```
>ssh -vvvv -i .\#keyfile# #user#@#host#
[...]
debug1: Next authentication method: publickey
debug1: Trying private key: .\\###keyfile###
debug3: sign_and_send_pubkey: RSA SHA256:###sha###
debug3: sign_and_send_pubkey: signing using rsa-sha2-512
[...]

PR #971, opened 2022-06-24
WIP: Implement support for rsa-sha2-256 key exchange.

yoyo-danielc:
There doesn't appear to be any way to select different exchange methods depending on what the server supports for the same key, so this commit allows IPrivateKeySource to expose multiple key exchange methods for the same key and will try each one.

Tested on Ubuntu 18.04 and 22.04 with default SSH server config and with only ssh-rsa and rsa-sha2-256 key exchange enabled.

Not sure if this is necessarily the best design to go with, so haven't updated the tests yet.

A simpler approach would be to just replace ssh-rsa with rsa-sha2-256, which would prevent communication with old servers (OpenSSH 7.2 added rsa-sha2-256 support in 2016).

Based on @ml054's work in https://github.com/sshnet/SSH.NET/issues/825#issuecomment-1139440419

### DIFF
--- a/src/Renci.SshNet/IPrivateKeySource.cs
+++ b/src/Renci.SshNet/IPrivateKeySource.cs
@@ -10,6 +10,6 @@ namespace Renci.SshNet
         /// <summary>
         /// Gets the host key.
         /// </summary>
-        HostAlgorithm HostKey { get; }
+        HostAlgorithm[] HostKeys { get; }
     }
 }
--- a/src/Renci.SshNet/ConnectionInfo.cs
+++ b/src/Renci.SshNet/ConnectionInfo.cs
@@ -389,6 +389,7 @@ namespace Renci.SshNet
 #endif
                     {"ssh-rsa", data => new KeyHostAlgorithm("ssh-rsa", new RsaKey(), data)},
                     {"ssh-dss", data => new KeyHostAlgorithm("ssh-dss", new DsaKey(), data)},
+                    {"rsa-sha2-256", data => new KeyHostAlgorithm("rsa-sha2-256", new RsaWithSha256SignatureKey(), data)},
                     //{"x509v3-sign-rsa", () => { ... },
                     //{"x509v3-sign-dss", () => { ... },
                     //{"spki-sign-rsa", () => { ... },
--- a/src/Renci.SshNet/PrivateKeyAuthenticationMethod.cs
+++ b/src/Renci.SshNet/PrivateKeyAuthenticationMethod.cs
@@ -4,6 +4,7 @@ using System.Collections.ObjectModel;
 using Renci.SshNet.Messages.Authentication;
 using Renci.SshNet.Messages;
 using Renci.SshNet.Common;
+using Renci.SshNet.Security;
 using System.Threading;
 
 namespace Renci.SshNet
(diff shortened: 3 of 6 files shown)


johnliaoywl:
Hi, I would like to know if there is any update for this one.

lcheng526:
when the library is going to upgrade to support sha-256 and sha-512?

lcheng526:
> There doesn't appear to be any way to select different exchange methods depending on what the server supports for the same key, so this commit allows IPrivateKeySource to expose multiple key exchange methods for the same key and will try each one.
> 
> Tested on Ubuntu 18.04 and 22.04 with default SSH server config and with only ssh-rsa and rsa-sha2-256 key exchange enabled.
> 
> Not sure if this is necessarily the best design to go with, so haven't updated the tests yet.
> 
> A simpler approach would be to just replace ssh-rsa with rsa-sha2-256, which would prevent communication with old servers (OpenSSH 7.2 added rsa-sha2-256 support in 2016).
> 
> Based on @ml054's work in [#825 (comment)](https://github.com/sshnet/SSH.NET/issues/825#issuecomment-1139440419)

Does the solution works? I am getting "Key exchanger negotiation failed" exception. The code is never hitting RsaSha256DigitalSignature class. It always hitting RsaKey class. any reason for this?

maxx8888:
Same for me, does this solution work? Do i need to change used HMAC Algorithm somehow?

darkoperator:
A lot of people are hitting this issue.

mikeblas:
What is preventing this from being merged? Looks like lots of people (me, too!) have this  issue: #825  and #956  for example.

ckdell:
@mikeblas : The build failing is probably the #1 blocker here.

Videstra:
Yeah - add me to the list of people having problems with this.  The temp workaround is to add the following two lines to the end of the **/etc/ssh/sshd_config** file
HostKeyAlgorithms +ssh-rsa
PubkeyAcceptedKeyTypes +ssh-rsa
If you don't have access to edit that file you will need a different library.   This appears to be specific to new versions of ubuntu that have deprecated the rsa keys...

et1975:
Is there an overlap with #1098?  @drieseng Any blockers other than rebase to merging this?

src/Renci.SshNet/IPrivateKeySource.cs
```diff
@@ -10,6 +10,6 @@ public interface IPrivateKeySource
         /// <summary>
         /// Gets the host key.
         /// </summary>
-        HostAlgorithm HostKey { get; }
+        HostAlgorithm[] HostKeys { get; }
```
et1975, line 13: This appears to be a pretty big change for the current codebase.

et1975: Any guidance @darinkes ?

phamvietdung:
image

It's not working for me :((

Videstra:
> It's not working for me :((
If you have access to the server you need to add the following two lines to the end of the /etc/ssh/sshd_config file
```
HostKeyAlgorithms +ssh-rsa
PubkeyAcceptedKeyTypes +ssh-rsa

```If you don't have access to edit that file you will need a different library.

**Comment by phamvietdung** [comment]
> > It's not working for me :((
> > If you have access to the server you need to add the following two lines to the end of the /etc/ssh/sshd_config file
> 
> ```
> HostKeyAlgorithms +ssh-rsa
> PubkeyAcceptedKeyTypes +ssh-rsa
> 
> ```If you don't have access to edit that file you will need a different library. 
> ```

Thanks, but i dont want to change the /etc/ssh/sshd_config  or maybe sometime i dont have permission to do this action.

**Comment by WojciechNagorski** [comment]
We will work on that in near future.

**Comment by hhklmc** [comment]
same issue here with Kernel version 5.10.50-linux4sam-2021.04+ whilst 4.9 can connect successfully.

**Comment by WojciechNagorski** [comment]
@hhklmc I've just finished the implementation of rsa-sha-256 here https://github.com/sshnet/SSH.NET/pull/1177

It would be great if you could confirm that the fix works so that we can release a new version this week.

**Comment by hhklmc** [comment]
> @hhklmc I've just finished the implementation of rsa-sha-256 here #1177
> 
> It would be great if you could confirm that the fix works so that we can release a new version this week.

How can I test it, any pre-release package published to nuget or?

**Comment by WojciechNagorski** [comment]
Currently, we don't have pre-release packages (I know we should have them). The best option is to download the repo, checkout to the branch, add the project to the solution, and reference the project instead of Nuget. This is all for testing purposes only.

```
git clone https://github.com/Rob-Hague/SSH.NET.git
cd SSH.NET
git checkout rsa-sha2
```

hhklmc:
> Currently, we don't have pre-release packages (I know we should have them). The best option is to download the repo, checkout to the branch, add the project to the solution, and reference the project instead of Nuget. This is all for testing purposes only.
> 
> ```
> git clone https://github.com/Rob-Hague/SSH.NET.git
> cd SSH.NET
> git checkout rsa-sha2
> ```

I can confirm that rsa-sha2 branch now works for me. Thank you very much!

WojciechNagorski:
@hhklmc big thanks for checking. Thanks to it we are closer to release. 😄

WojciechNagorski:
Fixed by #1177 and #1180

WojciechNagorski:

PR #975, opened 2022-06-30
Add IBaseClient for BaseClient and ISftpClient to inherit from

### LINKED ISSUE #890: Add methods from BaseClient to ISftpClient
Im still having a hard time mocking some of the methods in the SftpClient.
I came across the method Connect() which does not exist in the Interface, but comes from the BaseClient.
Would it be possible to add such methods to the ISftpClient interface? Or how would you mock them?

Owen-Krueger:
Fixes #890.

This will aid in unit testing by allowing developers to mock functionality from IBaseClient and ISftpClient. Currently, the only way to do that is to develop a wrapper around the functionality.

### DIFF
--- a/src/Renci.SshNet/BaseClient.cs
+++ b/src/Renci.SshNet/BaseClient.cs
@@ -13,7 +13,7 @@ namespace Renci.SshNet
     /// <summary>
     /// Serves as base class for client implementations, provides common client functionality.
     /// </summary>
-    public abstract class BaseClient : IDisposable, IBaseClient
+    public abstract class BaseClient : IBaseClient, IDisposable
     {
         /// <summary>
         /// Holds value indicating whether the connection info is owned by this client.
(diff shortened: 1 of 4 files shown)


drieseng (approved):

benjaminoerskov:
@drieseng Any idea when this will be released? :)

CasperWSchmidt:
@Owen-Krueger Can you please explain why you added `Dispose` to `IBaseClient` specifically instead of letting the interface extend `IDisposable`? Now the `BaseClient` class and the interface `ISftpClient` both inherits/extends `IDisposable`.

Owen-Krueger:
There isn't a good reason--it was a mistake on my part to include it in both files. I don't believe there's anything super wrong with it, just that it's some unnecessary duplication.

sga-solita:
I would also very much like to know when this is planned for release, as I need ISftpClient.Connect() for proper Dependency Injection functionality

Temporary fix which works well:

```
if (_sftpClient is SftpClient client)
{
    client.Connect();
}
// continue code which calls ISftpClient
```

Issue #454, opened 2018-08-02
ShellStream.Expect(Regex, TimeSpan) can return "too much" string


keithallenjackson:
ShellStream.Expect(Regex, TimeSpan) properly matches and dequeues data from the _incoming data queue. However, once a match is made, instead of returning all text up to the point of a match, it returns all data within the queue.

Current:
```csharp
public string Expect(Regex regex, TimeSpan timeout)
        {
           var text = string.Empty;

            while (true)
            {
                lock (_incoming)
                {
                    if (_incoming.Count > 0)
                    {
                        text = _encoding.GetString(_incoming.ToArray(), 0, _incoming.Count);
                    }

                    var match = regex.Match(text);

                    if (match.Success)
                    {
                        //  Remove processed items from the queue
                        for (var i = 0; i < match.Index + match.Length && _incoming.Count > 0; i++)
                        {
                            _incoming.Dequeue();
                        }
                        break;
                    }
                }

               if (timeout.Ticks > 0)
                    {
                        if (!_dataReceived.WaitOne(timeout))
                        {
                            return null;
                        }
                    }
                    else
                    {
                        _dataReceived.WaitOne();
                    }
           }

            return text;
        }
```

Fix:
```csharp
public string Expect(Regex regex, TimeSpan timeout)
        {
           var text = string.Empty;

            while (true)
            {
                lock (_incoming)
                {
                    if (_incoming.Count > 0)
                    {
                        text = _encoding.GetString(_incoming.ToArray(), 0, _incoming.Count);
                    }

                    var match = regex.Match(text);

                    if (match.Success)
                    {
                        //  Remove processed items from the queue
                        for (var i = 0; i < match.Index + match.Length && _incoming.Count > 0; i++)
                        {
                            _incoming.Dequeue();
                        }
                        text = text.Substring(0, match.Index + match.Length); // <-- This one line of code fixes it!
                        break;
                    }
                }

               if (timeout.Ticks > 0)
                    {
                        if (!_dataReceived.WaitOne(timeout))
                        {
                            return null;
                        }
                    }
                    else
                    {
                        _dataReceived.WaitOne();
                    }
           }

            return text;
        }

```

IgorMilavec:
I have created a discussion #926 for all ShellStream related issues.

IgorMilavec:
@keithallenjackson I'm trying to better understand usage patterns related to Expect API. Can you please explain how you use it and if possible provide a proto code for it?

[...]

Issue #383, opened 2018-02-07
ShellStream.Expect does not remove all matched data when multi-byte character encoding is used


JonathanFortier-eaton:
When we receive multi-bytes characters from a SSH server (for example, 'é' encoded in UTF-8 giving bytes [0xC3,  0xA9]), the ShellStream.Expect methods remove less bytes from the _incoming byte queue than what was really matched, so that the next read will return some of the data that should have been removed by Expect.

The bug seems to be when the functions call "_incoming.Dequeue()": they assume that the number of bytes to remove is the same as the number of character matched, which is incorrect for multi-byte encoding. 

Simple fix could be to calculate the number of byte used by the matched string (_encoding.GetByteCount(...)) like what ShellStream.ReadLine does. However, this can still return a value different of what was actually read (see https://stackoverflow.com/q/9740553).

A more complete fix could be to use "_encoding.GetDecoder().Convert(...)" (see https://msdn.microsoft.com/en-us/library/h6w985hz(v=vs.110).aspx) to calculate the number of bytes that correspond to the expected number of characters.

What do you think?

IgorMilavec:
I have created a discussion #926 for all ShellStream related issues.

jscarle:
@IgorMilavec @WojciechNagorski This was fixed by #1315.

PR #990, opened 2022-08-05
ShellStream.Expect returns string end with matched text

hhuang2013:
Modify ShellStream.Expect to trim string after matched text and process byte count of return string.

This PR may solve https://github.com/sshnet/SSH.NET/issues/383 and https://github.com/sshnet/SSH.NET/issues/454


WojciechNagorski:
I want to fix ShellStream in next version.

jscarle:
This PR can be closed as this was adressed by #1207, #1315, and #1319.

PR #1024, opened 2022-11-01
Bugfix/shell stream stopping event on chanell close

glenkleidon:
The ShellStream on channel close disposes of itself instead of leaving the task to the owner of the object.

Requires Starting and Stopping semantics same as for underlying shell. 

Full Bug fix for #44

Issue #40, opened 2016-07-11
Terminal window resizing


jcoder76:
Maybe I am looking at things improperly, but I was trying to find a way to support dynamic terminal resizing using SSH.NET as the connection.  Naturally the remote endpoint is expecting a window size in order to properly form the console output wrapping, but all the window sizing code in SSH.NET seems to be hidden as internals.

Although you can set the initial size in `ShellStream` or `Shell`, there are no methods for changing that while a connection is alive and you can't seem to relinquish the `ShellStream` and create a new one in the same connection without crashing the whole application.

I am using 2013.4.7 stable.

jcoder76:
No one is assigned to this issue, I haven't received any responses in 6 months, and no changes to this situation seem to have been implemented for 2016.0.0 which is the most current version as of the writing of this update.

To clarify this further, the component which has this feature is the `ChannelSession` in its `SendWindowChangeRequest` method.  The type and its implemented interface are both marked internal, yet that method isn't used anywhere internally which indicates the absolute uselessness of this method being internal.  Please expose this method somehow.

jcoder76:
Something like this is all that is needed:
https://github.com/Bugalex/SSH.NET/commit/f179b201a0e02f144f6dc9760ea9a7bb3931094e

jcoder76:
@drieseng or @olegkap is anyone maintaining this code anymore?

It has been 9 months without any answers.

darkoperator:
a new Beta was actually release yesterday, with tons of changes just need to take a look at the commits and issue updates. I believe it is not supported at the moment, you could do a PR based on that code and add some tests with permission of the person that did the fork.

jerix8055:
@drieseng is there any attempt to address this issue in future releases like the 2020 beta?

osagga:
@drieseng any update on this? I can work on the PR, but would this feature be interesting to add? or is there an opinion on this?

IgorMilavec:
I have created a discussion #926 for all ShellStream related issues.

PR #1062, opened 2022-12-11
Feature/change window size request for ssh client

glenkleidon:
Implemented **_ShellStream.WindowChangeRequest_**  by surfacing existing implementation in (private) __channel::IChannelSession_.

### DIFF
--- a/src/Renci.SshNet.Tests/Classes/ShellStreamTest.cs
+++ b/src/Renci.SshNet.Tests/Classes/ShellStreamTest.cs
@@ -108,7 +108,7 @@ namespace Renci.SshNet.Tests.Classes
         {
             var shellStream = CreateShellStream();
             _channelSessionMock.Setup(s => s.IsOpen).Returns(false);
-            Assert.IsFalse(shellStream.WindowChangeRequest(80, 25, 0, 0));
+            Assert.IsFalse(shellStream.ChangeWindow(80, 25, 0, 0));
         }
 
         [TestMethod]
@@ -119,7 +119,7 @@ namespace Renci.SshNet.Tests.Classes
             _channelSessionMock.Setup(s => s.SendWindowChangeRequest(
                   It.IsAny<uint>(), It.IsAny<uint>(),
                   It.IsAny<uint>(), It.IsAny<uint>())).Returns(false);
-            Assert.IsFalse(shellStream.WindowChangeRequest(80, 25, 0, 0));
+            Assert.IsFalse(shellStream.ChangeWindow(80, 25, 0, 0));
         }
 
         [TestMethod]
@@ -130,7 +130,7 @@ namespace Renci.SshNet.Tests.Classes
             _channelSessionMock.Setup(s => s.SendWindowChangeRequest(
                   It.IsAny<uint>(), It.IsAny<uint>(),
                   It.IsAny<uint>(), It.IsAny<uint>())).Returns(true);
-            Assert.IsTrue(shellStream.WindowChangeRequest(80, 25, 0, 0));
+            Assert.IsTrue(shellStream.ChangeWindow(80, 25, 0, 0));
             _channelSessionMock.Verify(v => v.SendWindowChangeRequest(80,25,0,0), Times.Once());
         }
 
--- a/src/Renci.SshNet.Tests/Classes/ShellStreamTest.cs
+++ b/src/Renci.SshNet.Tests/Classes/ShellStreamTest.cs
@@ -103,6 +103,14 @@ namespace Renci.SshNet.Tests.Classes
             _channelSessionMock.Verify(p => p.SendData(lineTerminator), Times.Once);
         }
 
+        [TestMethod]
+        public void WindowChangeRequest_ShouldReturnFalseWhenNotConnected()
+        {
+            var shellStream = CreateShellStream();
+            _channelSessionMock.Setup(s => s.IsOpen).Returns(false);
+            Assert.IsFalse(shellStream.WindowChangeRequest(80, 25, 0, 0));
+        }
+
         private ShellStream CreateShellStream()
         {
             _sessionMock.Setup(p => p.ConnectionInfo).Returns(_connectionInfoMock.Object);
--- a/src/Renci.SshNet.Tests/Classes/ShellStreamTest.cs
+++ b/src/Renci.SshNet.Tests/Classes/ShellStreamTest.cs
@@ -131,6 +131,7 @@ namespace Renci.SshNet.Tests.Classes
                   It.IsAny<uint>(), It.IsAny<uint>(),
                   It.IsAny<uint>(), It.IsAny<uint>())).Returns(true);
             Assert.IsTrue(shellStream.WindowChangeRequest(80, 25, 0, 0));
+            _channelSessionMock.Verify(v => v.SendWindowChangeRequest(80,25,0,0), Times.Once());
         }
 
         private ShellStream CreateShellStream()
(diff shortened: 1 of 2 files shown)


src/Renci.SshNet/ShellStream.cs
```diff
@@ -52,6 +54,19 @@ public bool DataAvailable
...
+        /// Sends a Window Change Request via the Channel.
+        /// </summary>
+        /// <param name="columns">New screen width in # of columns</param>
+        /// <param name="rows">New screen height in # of rows</param>
+        /// <param name="width">New screen width in Pixels</param>
+        /// <param name="height">New screen height in Pixels</param>
+        /// <returns>true when change is successful, or false when channel is NOT open or the request </returns>
+        public bool WindowChangeRequest(uint columns, uint rows, uint width, uint height )
+        {
+            if (_channel==null || !_channel.IsOpen) return false;
```
drieseng, line 67: Please have it throw an **ObjectDisposedException** when *_channel* is **null**, and an **InvalidOperationException** if the channel is closed. Please document this as such and add corresponding unit tests.

I would've prefered to have the method on **ChannelSession** also throw if the channel were closed, but that would be considered a breaking change. That method currently either throws or returns **true**, never **false**. It also returns **true** if the channel is closed 😌

src/Renci.SshNet/ShellStream.cs
```diff
@@ -52,6 +54,19 @@ public bool DataAvailable
...
         }
+        /// <summary>
+        /// Sends a Window Change Request via the Channel.
+        /// </summary>
+        /// <param name="columns">New screen width in # of columns</param>
+        /// <param name="rows">New screen height in # of rows</param>
+        /// <param name="width">New screen width in Pixels</param>
+        /// <param name="height">New screen height in Pixels</param>
+        /// <returns>true when change is successful, or false when channel is NOT open or the request </returns>
+        public bool WindowChangeRequest(uint columns, uint rows, uint width, uint height )
```
drieseng: Given the rationale below, we could consider having this method return **void**.
I prefer to discuss this with other team members before we decide how to proceed with this.

Please remove the extra whitespace before the closing parenthesis.

src/Renci.SshNet.Tests/Classes/ShellStreamTest.cs
```diff
@@ -103,6 +103,36 @@ public void WriteLine_Line_ShouldOnlyWriteLineTerminatorWhenLineIsNull()
...
+            _channelSessionMock.Setup(s => s.IsOpen).Returns(false);
+            Assert.IsFalse(shellStream.WindowChangeRequest(80, 25, 0, 0));
+        }
+
+        [TestMethod]
+        public void WindowChangeRequest_ShouldReturnFalseWhenResultFails()
+        {
+            var shellStream = CreateShellStream();
+            _channelSessionMock.Setup(s => s.IsOpen).Returns(true);
+            _channelSessionMock.Setup(s => s.SendWindowChangeRequest(
```
drieseng, line 119: Not very useful since `SendWindowChangeRequest(...)` always returns **true**, but let's discuss.

src/Renci.SshNet.Tests/Classes/ShellStreamTest.cs
```diff
@@ -103,6 +103,36 @@ public void WriteLine_Line_ShouldOnlyWriteLineTerminatorWhenLineIsNull()
...
+
+        [TestMethod]
+        public void WindowChangeRequest_ShouldReturnTrueWhenResultSucceeds()
+        {
+            var shellStream = CreateShellStream();
+            _channelSessionMock.Setup(s => s.IsOpen).Returns(true);
+            _channelSessionMock.Setup(s => s.SendWindowChangeRequest(
+                  It.IsAny<uint>(), It.IsAny<uint>(),
+                  It.IsAny<uint>(), It.IsAny<uint>())).Returns(true);
+            Assert.IsTrue(shellStream.WindowChangeRequest(80, 25, 0, 0));
```
drieseng: Please also verify that `SendWindowChangeRequest(...)` was actually invoked (once).

drieseng (changes requested):
Can you also contribute a small self-contained application that excercises this new method?
That application should signal an error if that method was not called exactly as expected (eg. if rows and columns where switched in our implementation).

I'd also like to discuss the name of this method. Are there better alternatives (eg. _ChangeWindow_)?

src/Renci.SshNet/ShellStream.cs
glenkleidon: ok removed white space. commit [ec60725]

src/Renci.SshNet/ShellStream.cs
glenkleidon: This method is essentially a wrapper for `SendWindowChangeRequest` which returns a boolean.  It makes sense that this method also returns boolean.  Nothing will break if the request fails, so therefore it is not exceptional that the method fails.

I always ask myself - as library developers, can we KNOW that it is **Exceptional** that a method is called when in the wrong state or is it simply _unexpected_.  Often the answer is that you can't know the answer because we don't know the context of how the method is being used.   The developer using the library will know if the situation is exceptional and can throw if it is.  All we need to do is say whether or not it succeeded.

If you throw an exception in a library - it has to be fundamentally wrong to do so.  In this case, I don't think it is - it is merely **_unexpected_** and of no consequence.  There wont be any data coming back to be "in the wrong shape" so it wont matter that the method failed.   There might be an argument for Throwing the exception if the channel had never been opened, but as this happens in the constructor that is not possible.

The scenario that comes to mind is, in the CanClose Event of a Desktop App, OR the Close browser event, the shape of the window may change during clean up.  If that happens, the event to change window shape MAY still be attached causing the event to fire when the channel has already closed.  It is much simpler to handle for the developer if the method does not throw an exception and the unexpected event can be ignored.   With an exception, in order to properly clean up, the developer must remember to unsubscribe to events or risk an exception being raised when it is of no consequence. 

Returning false allows to the developer to decide if failed call to change window size is exceptional.  So I think returning a boolean is the right call here.

src/Renci.SshNet/ShellStream.cs
glenkleidon: Should return false as it is not necessary to throw an exception in this case is often of no consequence that the method does not succeed.  Also it is a wrapper for the underlying method which returns boolean.

src/Renci.SshNet.Tests/Classes/ShellStreamTest.cs
glenkleidon: Hmm, I dont think SendWindowChangeRequest SHOULD always be true.  For example a change to size of 0,0,0,0 should fail and probably will on some pty implementations.

src/Renci.SshNet/ShellStream.cs
glenkleidon: Yes, `ChangeWindow` might be a better name.  Happy to do this.  Should I proceed?

glenkleidon: Ok changed in commit [61d5324]

glenkleidon: With regards to an example application - I have a fully worked example in branch `glenkleidon:feature/add_vs_2022_solution` 

Should consider doing a pull request on that one but I need some help with a failed test case.

glenkleidon: Actually the blocker is the failed test case on https://github.com/sshnet/SSH.NET/pull/1024

scott-xu:
Would you mind resolving the conflicts? @glenkleidon

glenkleidon:
I can Scott.  I am busy today but it has been quite a few months since I worked on it, so I’ll need to get myself back up to speed on it.

In the next week or so.

Glen.

jcoder76:
Tagging #40 so it is tracked there as well.

Rasmus715:
It's been a long week or so, mate

@glenkleidon can you finish this PR, please? If not, I'm happily willing to carry it further on my shoulders.

PR #1072, opened 2023-01-27
On SOCKS5 proxy: set hostname, not always IP

jjvaca:
The library always resolves the hostname to the IP. Some SOCKS5 proxies don't allow you to specify the IP, they want to resolve the hostname themselves.

The patch sends the proxy the same specified value, IP, or hostname.


Rob-Hague (approved):
The proxy connector code is not well tested in the library, but we can cross reference with:

the RFC https://datatracker.ietf.org/doc/html/rfc1928#section-5
the .NET implementation https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/SocksHelper.cs#L176
and even our own server implementation https://github.com/sshnet/SSH.NET/blob/6933e0961cd07a79b2af6a9f113f1cd38c034aef/src/Renci.SshNet/ForwardedPortDynamic.cs#L635

to see that it is obviously missing here. Thanks!

PR #1075, opened 2023-01-29
Add 'SayHello' method to 'Session' class to get Host key without authentication

omidmohajers:
Add 'SayHello' method to 'Session' class to get Host key without authentication

### DIFF
--- a/.vs/ProjectSettings.json
+++ b/.vs/ProjectSettings.json
@@ -1,0 +1,3 @@
+{
+  "CurrentProjectSetting": null
+}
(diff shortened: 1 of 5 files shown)


Rob-Hague:
Thanks. I'll close this as undesired since it is already possible outside of the library:

```c#

public static string SayHello(string host)
{
    string hostKey = null;

    using (SshClient sftpClient = new(host, "username", "password"))
    {
        sftpClient.HostKeyReceived += (sender, args) =>
        {
            hostKey = args.FingerPrintSHA256;
        };

        try
        {
            sftpClient.Connect();
        }
        catch (SshAuthenticationException)
        {
            // This is expected.
        }
    }

    return hostKey;
}

```

PR #1096, opened 2023-03-17
Bugfix Dispose - Avoid sending message when session is not connected.

shanwayne:
When connection is disconnected from the server for unexpected reasons, attempting to dispose the forwarded port will throw exception because it believes the session is still connected and will try and send a message.

### DIFF
--- a/src/Renci.SshNet.Tests/Classes/ForwardedPortRemoteTest_Dispose_PortStarted_ChannelBound_Disconnect.cs
+++ b/src/Renci.SshNet.Tests/Classes/ForwardedPortRemoteTest_Dispose_PortStarted_ChannelBound_Disconnect.cs
@@ -1,0 +1,251 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Net;
+using System.Threading;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using Renci.SshNet.Channels;
+using Renci.SshNet.Common;
+using Renci.SshNet.Messages.Connection;
+
+namespace Renci.SshNet.Tests.Classes
+{
+    [TestClass]
+    public class ForwardedPortRemoteTest_Dispose_PortStarted_ChannelBound_Disconnect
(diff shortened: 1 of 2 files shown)


shanwayne:
@drieseng Is there any change I can make to help move this forward? I've added a test case which covers the issue.

WojciechNagorski:
@shanwayne We're really close to merging #1109 PR which will provide native support for .NET 6 and 7.
After that, I'd like to improve integration tests. Thanks to them, we will find out if this PR didn't introduce any regression.

WojciechNagorski:
In `ForwardedProtDynamic` I found one breaking change for .NET 5 and above. There is a more information https://github.com/sshnet/SSH.NET/blob/89079bd23e372d6ba086cea1dd434ac99715cc0f/src/Renci.SshNet/ForwardedPortDynamic.NET.cs#L207-L219
Do you know if this can be related to this PR?

shanwayne:
@WojciechNagorski It could be similar but different. In `ForwardedPortRemote` the Session object is still not null after the connection is aborted and when disposing it attempts to send a message but the client is not connected so an unhandled exception is thrown on dispose. The other `ForwardedPort` don't have this behavior.

WojciechNagorski:
@shanwayne Can I know why did you close this issue? Did you find the problem in the library or maybe on the server side?

I'm waiting for more permissions for this project and intend to reactivate this project once I get them.

shanwayne:
@WojciechNagorski This is still an issue. I had to move on unfortunately and there was a more serious issue regarding not properly handling connections that prevented using the library. So just cleaning up.

jwfx:
@WojciechNagorski 

Any chance this pull request still can be merged? I'm hitting this exact issue on 2024.0.0 and this PR seems to fix it.

WojciechNagorski:
@shanwayne @jwfx You should ask @Rob-Hague. I don't have time for this project for the next month.

WojciechNagorski:
@Rob-Hague I reopened this PR, if you want you can close it.

PR #1109, opened 2023-04-11
Removing old target frameworks

### LINKED ISSUE #665: [RFC] Target framework support

WojciechNagorski:
I'm creating a PR to start a conversation about the topic. There is still a lot of work to do. 

We have to make some decisions:
* Choose which target framework we want to support. My proposition is: `net4.7.2`, `netstandard 2.0`, `.net7.0`. I've seen some features (for example FEATURE_HMAC_RIPEMD160) that are only implemented for the old .NET Framework. If we want to keep these features, we should keep the target platform `net472`, because `netstandard2.0` does not contain these APIs.
* I need more information about the building, preparing Nugets, and integration tests. It would be great if we could make a call?

Important information:
- On .NET 7 `ArgumentOutOfRangeException` has a different message format. There are tests that check the message and we should fix them.
- New .NET 6 and 7 contain some changes, and had to make the appropriate changes:
  - https://learn.microsoft.com/en-us/dotnet/core/compatibility/networking/7.0/socket-end-closed-sockets
  - https://github.com/dotnet/runtime/issues/41585
  - https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/5.0/thread-abort-obsolete

Fixes: #665


WojciechNagorski:
There are passed: 2529 and failed: 13 tests for .NET 7.

OranguTech:
I would suggest targeting `.net6.0`, rather than 7, as it is the current [Long Term Support version](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core#lifecycle).

src/Renci.SshNet.Tests/Classes/Common/CountdownEventTest.cs
```diff
@@ -189,14 +192,19 @@ public void Wait_ShouldReturnFalseWhenTimeoutExpiresBeforeCountdownEventIsSet()
...
 
             Assert.IsFalse(actual);
             Assert.IsFalse(countdownEvent.IsSet);
             Assert.IsFalse(countdownEvent.WaitHandle.WaitOne(0));
-            Assert.IsTrue(elapsedTime >= timeout);
+            // Use precision because it fail in CI pipeline on .NET 7 and also on .NET Framework with Stopwatch.
+            // Locally it works without precision. This is internal .NET implementation, so this is 
+            // not super important
+            var precision = TimeSpan.FromMilliseconds(10);
+            Assert.IsTrue(watch.Elapsed >= timeout.Subtract(precision));
```
zybexXL: This is not acceptable, it's faking the timer to get to a desired result.

This entire testcase is wrong and prone to race conditions. It starts up to 20 threads, and then starts measuring time. But it fails to account for the time it takes to start the threads; by the time the last thread is started, the first one might have already completed (or be almost complete), so the 30ms timeout may in fact be a sufficient delay for the event to be signaled. The 20 thread execution will also depend on whether this runs on a 4/8/16/32 core machine - the 32-core machine can start the 20 threads much faster than the 4-core one.

Also, what exactly is this test testing? I would just throw it away if it's just testing .Net intrinsics.

(I added this comment on a previous commit but it's hidden now as I didn't started a review)

zybexXL (changes requested):

src/Renci.SshNet.Tests/Classes/Common/CountdownEventTest.cs
WojciechNagorski: @zybexXL Thanks for the review. I leave this code because I wanted to show why I change this but I agree that this test is not really useful, it just tests .NET's feature. So maybe I just remove this one Assert? Or do you want to remove the entire file `CountdownEventTest`? 

Similar situation is here: https://github.com/sshnet/SSH.NET/pull/1109/files/79123fbadd40f8758b27c71b66896f51fe7bc8fc#diff-791204553618e1fe9731d44416970367c234ce1e1f6b9f5b12c883a2646d992aR338-R342

WojciechNagorski:
All tests are green and I understand all changes. Now I'm going to clean up this PR.

WojciechNagorski:
@drieseng this is ready for review. If everything will be ok I will change the build scripts https://github.com/sshnet/SSH.NET/tree/develop/build

Kim-SSi:
Should .net 4.6.2 also be included? It is supported by Microsoft until Jan 12, 2027.

WojciechNagorski:
@Kim-SSi good idea. But for now I'm waiting for @drieseng's review

OranguTech:
> Should .net 4.6.2 also be included? It is supported by Microsoft until Jan 12, 2027.

See also my comment on 6.x.

WojciechNagorski:
@OranguTech I've run tests for `.NET 6.0` and everything works. I don't mind supporting both versions. `.NET 6.0` will be supported up to [Nov 12, 2024](https://learn.microsoft.com/en-us/lifecycle/products/microsoft-net-and-net-core) and this year MS will release `.NET 8.0` that will be LTS. Then we can add support for a new one. 

But first I need a review from @drieseng

WojciechNagorski:
I've run integration tests on this branch and everything works well.

WojciechNagorski:
I've prepared a PR for integration tests https://github.com/sshnet/IntegrationTests/pull/10

src/Renci.SshNet.Tests/Classes/BaseClientTest_Connected_KeepAliveInterval_NotNegativeOne.cs
```diff
@@ -57,7 +57,11 @@ protected override void Act()
             _client.KeepAliveInterval = _keepAliveInterval;
 
             // allow keep-alive to be sent a few times
+#if NETFRAMEWORK
```
drieseng, line 61: Meaningful to have conditional compilation for this?

src/Renci.SshNet.Tests/Classes/ClientAuthenticationTest.cs
```diff
@@ -28,7 +28,7 @@ public void Ctor_PartialSuccessLimit_Zero()
             catch (ArgumentOutOfRangeException ex)
             {
                 Assert.IsNull(ex.InnerException);
-                Assert.AreEqual(string.Format("Cannot be less than one.{0}Parameter name: {1}", Environment.NewLine, ex.ParamName), ex.Message);
+                Assert.AreEqual(new ArgumentOutOfRangeException("partialSuccessLimit", "Cannot be less than one.").Message, ex.Message);
```
drieseng: I propose to add an **ArgumentExceptionAssert** class with a static `MessageEquals(string expected, ArgumentException )` that - under the hood - creates a new **ArgumentException** with the expected message and the **ParamName** of the specified **ArgumentException** and asserts whether the **Message** of that newly created **ArgumentException** matches the **Message** of the specified **ArgumentException**.

src/Renci.SshNet.Tests/Classes/ClientAuthenticationTest.cs
```diff
@@ -46,7 +46,7 @@ public void Ctor_PartialSuccessLimit_Negative()
...
```
drieseng: See earlier remark.

src/Renci.SshNet.Tests/Classes/Common/PacketDumpTest.cs
```diff
@@ -37,11 +37,9 @@ public void Create_ByteArrayAndIndentLevel_IndentLevelLessThanZero()
             catch (ArgumentOutOfRangeException ex)
             {
                 Assert.IsNull(ex.InnerException);
-#if NETFRAMEWORK
-                Assert.AreEqual(string.Format("Cannot be less than zero.{0}Parameter name: {1}", Environment.NewLine, ex.ParamName), ex.Message);
-#else
-                Assert.AreEqual(string.Format("Cannot be less than zero. (Parameter '{1}')", Environment.NewLine, ex.ParamName), ex.Message);
-#endif
+
+                Assert.AreEqual(new ArgumentOutOfRangeException("indentLevel", "Cannot be less than zero.").Message, ex.Message);
```
drieseng: See earlier remark.

src/Renci.SshNet.Tests/Classes/Common/PosixPathTest_CreateAbsoluteOrRelativeFilePath.cs
```diff
@@ -37,7 +37,7 @@ public void Path_Empty()
             catch (ArgumentException ex)
             {
                 Assert.IsNull(ex.InnerException);
-                Assert.AreEqual(string.Format("The path is a zero-length string.{0}Parameter name: {1}", Environment.NewLine, ex.ParamName), ex.Message);
+                Assert.AreEqual(new ArgumentException("The path is a zero-length string.", "path").Message, ex.Message);
```
drieseng: See earlier remark.

src/Renci.SshNet.Tests/Classes/NetConfClientTest.cs
```diff
@@ -86,7 +86,7 @@ public void OperationTimeout_LessThanLowerLimit()
             catch (ArgumentOutOfRangeException ex)
             {
                 Assert.IsNull(ex.InnerException);
-                Assert.AreEqual("The timeout must represent a value between -1 and Int32.MaxValue, inclusive." + Environment.NewLine + "Parameter name: " + ex.ParamName, ex.Message);
+                Assert.AreEqual(new ArgumentOutOfRangeException("value", "The timeout must represent a value between -1 and Int32.MaxValue, inclusive.").Message, ex.Message);
```
drieseng: See earlier remark.

src/Renci.SshNet.Tests/Classes/NetConfClientTest.cs
```diff
@@ -105,7 +105,7 @@ public void OperationTimeout_GreaterThanLowerLimit()
...
```
drieseng: See earlier remark.

src/Renci.SshNet.Tests/Classes/SftpClientTest.cs
```diff
@@ -94,11 +94,9 @@ public void OperationTimeout_LessThanLowerLimit()
             catch (ArgumentOutOfRangeException ex)
             {
                 Assert.IsNull(ex.InnerException);
-#if NETFRAMEWORK
-                Assert.AreEqual("The timeout must represent a value between -1 and Int32.MaxValue, inclusive." + Environment.NewLine + "Parameter name: " + ex.ParamName, ex.Message);
-#else
-                Assert.AreEqual("The timeout must represent a value between -1 and Int32.MaxValue, inclusive. (Parameter '" + ex.ParamName + "')", ex.Message);
-#endif
+
+                Assert.AreEqual(new ArgumentOutOfRangeException("value", "The timeout must represent a value between -1 and Int32.MaxValue, inclusive.").Message, ex.Message);
```
drieseng: See earlier remark.

src/Renci.SshNet.Tests/Classes/SftpClientTest.cs
```diff
@@ -117,11 +115,9 @@ public void OperationTimeout_GreaterThanLowerLimit()
...
```
drieseng: See earlier remark.

src/Renci.SshNet.Tests/Common/TripleATestBase.cs
```diff
@@ -24,6 +25,10 @@ protected virtual void TearDown()
         protected abstract void Arrange();
 
         protected abstract void Act();
+
+        protected virtual void AfterAct()
```
drieseng: When would or do you use this?

src/Renci.SshNet/ForwardedPortDynamic.NET.cs
```diff
@@ -204,10 +204,19 @@ private bool HandleSocks(IChannelDirectTcpip channel, Socket clientSocket, TimeS
...
                 // ignore exception thrown by interrupting the blocking receive as part of closing
                 // the forwarded port
+#if NETFRAMEWORK
                 if (ex.SocketErrorCode != SocketError.Interrupted)
                 {
                     RaiseExceptionEvent(ex);
                 }
+#else
+                // Since .NET 5 the exception has been changed. 
+                // more info https://github.com/dotnet/runtime/issues/41585
```
drieseng, line 214: Nice catch!

src/Renci.SshNet/Renci.SshNet.csproj
```diff
@@ -7,44 +7,18 @@
...
     <LangVersion>6</LangVersion>
     <SignAssembly>true</SignAssembly>
-    <TargetFrameworks>net35;net40;net472;netstandard1.3;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net472;netstandard2.0;net7.0</TargetFrameworks>
   </PropertyGroup>
 
-  <!--
-  <PropertyGroup Condition=" '$(VisualStudioVersion)' == '16.0' ">
-    <TargetFrameworks>net35;net40;netstandard1.3;netstandard2.0;netstandard2.1</TargetFrameworks>
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'net472' ">
```
drieseng: We can probably remove lots of the symbols as the corresponding API's are now available for all supported target frameworks. This can be done in a follow-up PR of course.

drieseng (changes requested):
Nice work!

src/Renci.SshNet.Tests/Classes/BaseClientTest_Connected_KeepAliveInterval_NotNegativeOne.cs
WojciechNagorski: TBH I didn't have a better idea for this. We have the following test:

https://github.com/sshnet/SSH.NET/blob/8201be44094555b1588331d339c771f7940ec6ab/src/Renci.SshNet.Tests/Classes/BaseClientTest_Connected_KeepAliveInterval_NotNegativeOne.cs#L99-L102

Which checks if the message was sent three times, but `.NET 7` has better performance and is able to send four messages in the same time. Any idea?

src/Renci.SshNet/Renci.SshNet.csproj
WojciechNagorski: Yes I've seen some places but I think it should be next PR only for this.

src/Renci.SshNet.Tests/Common/TripleATestBase.cs
WojciechNagorski: I override this method in `HttpConnectorTestBase`, `Socks4ConnectorTestBase`, `Socks5ConnectorTestBase`. Example implementation: 
https://github.com/sshnet/SSH.NET/blob/984c5c360bd115f2a60c249c5221dbf65879b024/src/Renci.SshNet.Tests/Classes/Connection/HttpConnectorTestBase.cs#L29-L33

NET7 is much more efficient so tests were performed first and then events like `Disconnected`, `BytesReceived`. For instance: https://github.com/sshnet/SSH.NET/blob/984c5c360bd115f2a60c249c5221dbf65879b024/src/Renci.SshNet.Tests/Classes/Connection/Socks4ConnectorTest_Connect_ConnectionRejectedByProxy.cs#L33-L48 

The problem occurs especially when running tests from the console.

WojciechNagorski:
@drieseng I've fixed review.

WojciechNagorski:
@OranguTech I've added support for .NET 6.0.

src/Renci.SshNet.Tests/Common/TripleATestBase.cs
drieseng: I prefer to remove this **AfterAct** method, and where necessary ntegrate the sleep in the **Act** for those tests that need it. OK for you, or want to discuss this further?

src/Renci.SshNet/Renci.SshNet.csproj
drieseng: I agree

build/sandcastle/SSH.NET.shfbproj
```diff
@@ -1,21 +1,22 @@
...
     <ProjectGuid>{f7266fb1-f50a-4a5b-b35a-5ea8ebdc1be9}</ProjectGuid>
-    <SHFBSchemaVersion>2015.6.5.0</SHFBSchemaVersion>
+    <SHFBSchemaVersion>2017.9.26.0</SHFBSchemaVersion>
     <!-- AssemblyName, Name, and RootNamespace are not used by SHFB but Visual Studio adds them anyway -->
     <AssemblyName>Documentation</AssemblyName>
     <RootNamespace>Documentation</RootNamespace>
     <Name>Documentation</Name>
     <!-- SHFB properties -->
-    <FrameworkVersion>.NET Framework 4.0</FrameworkVersion>
+    <FrameworkVersion>.NET Framework 4.7.2</FrameworkVersion>
```
drieseng: Why not 4.6.2 (which is the version of .NET Framework we're explicitly supporting).
Note, I'd be ok for moving the needle to .NET 4.8 (if there any opportunities that we're missing out by still targeting .NET 4.6.2 instead).

build/sandcastle/SSH.NET.shfbproj
```diff
@@ -1,21 +1,22 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="4.0">
   <PropertyGroup>
-    <!--
-      The configuration and platform will be used to determine which assemblies to include from solution and
-      project documentation sources
-    -->
+      <!-- A target framework version is required by Visual Studio.  It can be any version with a targeting pack installed. -->
+    <TargetFrameworkVersion>v4.7.2</TargetFrameworkVersion>
```
drieseng: See below.

src/Renci.SshNet.Tests/Classes/BaseClientTest_Connected_KeepAliveInterval_NotNegativeOne.cs
drieseng: It's ok for, perhaps just add a comment explaining why.

src/Renci.SshNet.Tests/Common/ArgumentExceptionAssert.cs
```diff
@@ -0,0 +1,16 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Renci.SshNet.Tests.Common
+{
+    public static class ArgumentExceptionAssert
+    {
+        public static void MessageEquals(string paramName, string expected, ArgumentException exception)
+        {
+            var type = exception.GetType();
+            var newMessage = (ArgumentException)Activator.CreateInstance(type, paramName, expected);
```
drieseng: I don't see why you can't just 'new' the **ArgumentException**.
Do some classes that derive from **ArgumentException** have a differerent implementation of **Message**?
**ArgumentNullException** doesn't, I'm sure of that.

src/Renci.SshNet.Tests/Common/ArgumentExceptionAssert.cs
```diff
@@ -0,0 +1,16 @@
...
+
+            Assert.AreEqual(newMessage?.Message, exception.Message);
```
drieseng: Null conditional operator can be removed if you new the **ArgumentException**.

src/Renci.SshNet.Tests/Common/TripleATestBase.cs
WojciechNagorski: It will be ok for me.

WojciechNagorski: I've removed the `AfterAct` method and added `Thread.Sleep(200)` for all tests that use `AsyncSocketListener`.

drieseng (changes requested):

src/Renci.SshNet.Tests/Common/ArgumentExceptionAssert.cs
WojciechNagorski: You're right. I overcomplicated.

build/sandcastle/SSH.NET.shfbproj
WojciechNagorski: I rolled back to 4.6.2. There can be any version of .NET Framework because it doesn't affect the documentation file.

src/Renci.SshNet.Tests/Common/ArgumentExceptionAssert.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Renci.SshNet.Tests.Common
+{
+    public static class ArgumentExceptionAssert
+    {
+        public static void MessageEquals(string paramName, string expected, ArgumentException exception)
```
drieseng: I'd remove the **paramName** argument and just use **ParamName** from **exception**.  This is about checking **Message**, the value of **ParamName** is (should be) checked separately. 

I'm ok with merging this without this, let me know if you'll change it right away.

WojciechNagorski: I've removed **paramName**.

WojciechNagorski:
I corrected all comments and added answers.

drieseng (approved):
Looks great, thanks!!

PR #1117, opened 2023-05-08
Remove old features [Part 1]

WojciechNagorski:
In this PR I've removed old features that are no longer needed. Each feature was removed in a separate commit so it should be easier to review it.

There are still three more features to be removed, but I would like to remove them in separate pull requests.


drieseng (approved):
Nice!!!

PR #1119, opened 2023-05-10
Remove FEATURE_DIRECTORYINFO_ENUMERATEFILES

WojciechNagorski:
In this PR I've removed `FEATURE_DIRECTORYINFO_ENUMERATEFILES`. 

`Dictionary.EnumerateFiles` is available for .NET 4.6.2 and .NET 6.0+. I used `IEnumerable` without calling `.ToList()` using the following pattern:
```C#
var e = enumerable.GetEnumerator();
// check first
if(!e.MoveNext()) return;
// do some stuff, then enumerate the list
do {
    actOn(e.Current);  // do stuff with the current item
} while(e.MoveNext()); // stop when we don't have anything else
```


src/Renci.SshNet/SftpClient.cs
```diff
@@ -2074,60 +2074,68 @@ private IEnumerable<FileInfo> InternalSynchronizeDirectories(string sourcePath,
...
                 {
-                    var remoteFileName = string.Format(CultureInfo.InvariantCulture, @"{0}/{1}", destinationPath, localFile.Name);
-                    try
+                    var localFile = sourceFiles.Current;
+                    if (localFile == null)
+                    {
+                        continue;
+                    }
+
+                    var isDifferent = !destDict.ContainsKey(localFile.Name);
```
drieseng: Avoid `ContainsKey(string key)` followed by indexer, and use `if (TryGetValue(string key, out var remoteFile))` instead to know whether you're dealing with an existing file.

In the if block, you can then perform an up-to-date check and use **continue** to skip the current file if it is.
That avoids nesting the whole upload block itself.

src/Renci.SshNet/SftpClient.cs
```diff
@@ -2074,60 +2074,68 @@ private IEnumerable<FileInfo> InternalSynchronizeDirectories(string sourcePath,
...
+
+                    if (!isDifferent)
                     {
-                        using (var file = File.OpenRead(localFile.FullName))
+                        var temp = destDict[localFile.Name];
+                        //  TODO:   Use md5 to detect a difference
```
drieseng, line 2116: I'd like to avoid sprinkling TODO's all over our codebase, and instead create issues (of which we have plently, I know :().

src/Renci.SshNet/SftpClient.cs
```diff
@@ -2074,60 +2074,68 @@ private IEnumerable<FileInfo> InternalSynchronizeDirectories(string sourcePath,
...
-                if (!isDifferent)
+                var destFiles = InternalListDirectory(destinationPath, null);
+                var destDict = new Dictionary<string, ISftpFile>();
+                foreach (var destFile in destFiles)
                 {
-                    var temp = destDict[localFile.Name];
-                    //  TODO:   Use md5 to detect a difference
-                    //ltang: File exists at the destination => Using filesize to detect the difference
-                    isDifferent = localFile.Length != temp.Length;
+                    if (destFile.IsDirectory)
```
drieseng, line 2092: Please always use braces, even for single line blocks.

src/Renci.SshNet/SftpClient.cs
```diff
@@ -2074,60 +2074,68 @@ private IEnumerable<FileInfo> InternalSynchronizeDirectories(string sourcePath,
...
-            var destFiles = InternalListDirectory(destinationPath, null);
-            var destDict = new Dictionary<string, ISftpFile>();
-            foreach (var destFile in destFiles)
+            using (var sourceFiles = sourceDirectory.EnumerateFiles(searchPattern).GetEnumerator())
             {
-                if (destFile.IsDirectory)
-                    continue;
-                destDict.Add(destFile.Name, destFile);
-            }
+                if(!sourceFiles.MoveNext())
```
drieseng: Please always use braces, even for single line blocks. Also add space between if and paranthese.

src/Renci.SshNet/SftpClient.cs
```diff
@@ -2074,60 +2074,68 @@ private IEnumerable<FileInfo> InternalSynchronizeDirectories(string sourcePath,
...
+
+                #region Upload the difference
+
+                const Flags uploadFlag = Flags.Write | Flags.Truncate | Flags.CreateNewOrOpen;
+                do
                 {
-                    var remoteFileName = string.Format(CultureInfo.InvariantCulture, @"{0}/{1}", destinationPath, localFile.Name);
-                    try
+                    var localFile = sourceFiles.Current;
+                    if (localFile == null)
```
drieseng, line 2108: Why would `MoveNext()`on the enumerator return **true**, but **Current** return **false**?

src/Renci.SshNet/SftpClient.cs
```diff
@@ -2074,60 +2074,68 @@ private IEnumerable<FileInfo> InternalSynchronizeDirectories(string sourcePath,
...
-                        if (asynchResult != null)
+                            if (asynchResult != null)
+                            {
+                                asynchResult.Update(uploadedFiles.Count);
+                            }
+                        }
+                        catch (Exception ex)
                         {
-                            asynchResult.Update(uploadedFiles.Count);
+                            throw new Exception(string.Format("Failed to upload {0} to {1}", localFile.FullName, remoteFileName), ex);
```
drieseng: We should not throw (have been throwing from the start) a plain **Exception**. I propose to throw an **SshException** instead (and document this), but I'm open for discussion.

src/Renci.SshNet/SftpClient.cs
```diff
@@ -2074,60 +2074,68 @@ private IEnumerable<FileInfo> InternalSynchronizeDirectories(string sourcePath,
...
                         }
                     }
-                    catch (Exception ex)
-                    {
-                        throw new Exception(string.Format("Failed to upload {0} to {1}", localFile.FullName, remoteFileName), ex);
-                    }
-                }
+                } while (sourceFiles.MoveNext());
```
drieseng: Please move the `while` to separate line.

src/Renci.SshNet/SftpClient.cs
```diff
@@ -2074,60 +2074,68 @@ private IEnumerable<FileInfo> InternalSynchronizeDirectories(string sourcePath,
...
-            var sourceFiles = FileSystemAbstraction.EnumerateFiles(sourceDirectory, searchPattern).ToList();
-            if (sourceFiles.Count == 0)
-                return uploadedFiles;
-
-            #region Existing Files at The Destination
-
-            var destFiles = InternalListDirectory(destinationPath, null);
-            var destDict = new Dictionary<string, ISftpFile>();
-            foreach (var destFile in destFiles)
+            using (var sourceFiles = sourceDirectory.EnumerateFiles(searchPattern).GetEnumerator())
```
drieseng, line 2079: What do we gain from not using a simple `foreach`?

Do we want to switch to deferred execution / lazy evaluation (using `yield return`, ...)? It could be considered a breaking change though, but I'd be ok with it (if we document it).

You'd still need to keep track of the number of files that we've uploaded to pass it to **SftpSynchronizeDirectoriesAsyncResult**.

src/Renci.SshNet/SftpClient.cs
```diff
@@ -2074,60 +2074,68 @@ private IEnumerable<FileInfo> InternalSynchronizeDirectories(string sourcePath,
...
+                    {
+                        continue;
+                    }
+
+                    var isDifferent = !destDict.ContainsKey(localFile.Name);
+
+                    if (!isDifferent)
                     {
-                        using (var file = File.OpenRead(localFile.FullName))
+                        var temp = destDict[localFile.Name];
```
drieseng: See `TryGetValue(...)` remark above.

drieseng (changes requested):

drieseng:
@WojciechNagorski: as you mentioned offline, not all of my remarks are related to changes you did.

src/Renci.SshNet/SftpClient.cs
WojciechNagorski: I agree it looks weird. I changed it to SshException and added documentation.

src/Renci.SshNet/SftpClient.cs
WojciechNagorski: In this task I'm going to remove this:
https://github.com/sshnet/SSH.NET/blob/8c932fb1e85760864393c2ed72e078a5bd8ddd9e/src/Renci.SshNet/Abstractions/FileSystemAbstraction.cs#L25-L29

The difference between those methods:
>The EnumerateFiles and GetFiles methods differ as follows: When you use EnumerateFiles, you can start enumerating the collection of names before the whole collection is returned; when you use GetFiles, you must wait for the whole array of names to be returned before you can access the array. Therefore, when you are working with many files and directories, EnumerateFiles can be more efficient.

But these lines make no sense because we always use `.ToList()` on it:
https://github.com/sshnet/SSH.NET/blob/8c932fb1e85760864393c2ed72e078a5bd8ddd9e/src/Renci.SshNet/SftpClient.cs#L2077
Both methods are available for both `.NET 4.6.2` and `.NET Standard 2.0`. I think that in our case `EnumerateFiles` is better but we should not use `.ToList()`. 

I can't use foreach because of "possible multiple enumeration"

```c#
var list = enumerable.GetEnumerator();

if (!list.any()) // <-- First use
{ 
  return;
}
foreach (var item in list) // <-- Second use
{
  // Do  some stuff
}

```
We can use the following pattern (from https://stackoverflow.com/a/39277405/4809321):
```c#
var e = enumerable.GetEnumerator();
// check first
if(!e.MoveNext())
{
  return;
}
// do some stuff, then enumerate the list
do {
    actOn(e.Current);  // do stuff with the current item
} 
while(e.MoveNext()); // stop when we don't have anything else
```
Where:
`e.MoveNext()` - returns `true` if the enumerator was successfully advanced to the next element; `false` if the enumerator has passed the end of the collection.

The method `InternalSynchronizeDirectories` collects the entire result in the list `uploadedFiles`
https://github.com/sshnet/SSH.NET/blob/6828536a7fa54a3929ccc40da908315df86b155f/src/Renci.SshNet/SftpClient.cs#L2073
so it is not a breaking change but thanks to using `EnumerateFiles` we will only have one list in memory, not two. 

Of course if you want we can leave `.ToList()` with `foreach` or use `dictionary.GetFiles()` with `foreach`

src/Renci.SshNet/SftpClient.cs
WojciechNagorski: `MoveNext()` - returns true if the enumerator was successfully advanced to the next element; false if the enumerator has passed the end of the collection.
`Current` - returns the element in the collection at the current position of the enumerator.

src/Renci.SshNet/SftpClient.cs
WojciechNagorski: I didn't add this comment so I'm going to leave it here. I've seen a PR for this TODO.

src/Renci.SshNet/SftpClient.cs
drieseng: I think I knew that already :p
The question is why do you check for **Current** being **null**? You already verified that `MoveNext()` returned **true**, so why the extra null check?

src/Renci.SshNet/SftpClient.cs
drieseng: I had a closer look at the implementation now.
We do not use a `foreach` to iterate over the result returned by `sourceDirectory.EnumerateFiles(searchPattern)` to avoid initializing the `destFiles` dictionary if there are no local files matching the pattern.

src/Renci.SshNet/SftpClient.cs
WojciechNagorski: `Current` can be null. You know the list can contain the null value. Of course, this is unlikely at this pleace, and I want to avoid hinting about the possibility of null.

https://dotnetfiddle.net/7qPNg6
```
	public static void Main()
	{
		List<string> list = new List<string>();
		list.Add("A");
		list.Add("B");
		list.Add(null);
		list.Add("D");
		
		using (var e = list.GetEnumerator())
		{
			// check first
			if(!e.MoveNext())
			{
			  return;
			}
			// do some stuff, then enumerate the list
			do {
				Console.WriteLine(e.Current ?? "This is null");
			} 
			while(e.MoveNext()); // stop when we don't have anything else
		}
	}
```

returns:
```
A
B
This is null
D
```

drieseng: Indeed, but I was talking about this specific use. `DirectoryInfo.EnumerateFiles(...)` does not return **null** entries.
I'll remove the null check in a follow-up PR.

drieseng (approved):

PR #1125, opened 2023-05-22
Fix some (lots of) issues reported by analyzers.

drieseng:
Fix some (lots of) issues reported by analyzers.
I only fixed those issues that do not require discussion.
Still lots of issues left to fix.

For now, I renamed the **.editorconfig** that I used to **.editorconfig_soon** to "disable" it until more issues are fixed.
Enabling the analyzers adds quite some time to the build, but that is expected.


Directory.Build.props
```diff
@@ -0,0 +1,48 @@
...
+        <AnalysisLevel>preview-All</AnalysisLevel>
+        <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+    </PropertyGroup>
+
+    <!--
+        Add the stylecop config to each project.
+    -->
+    <ItemGroup>
+        <AdditionalFiles Include="..\..\stylecop.json" Link="stylecop.json" />
+    </ItemGroup>
```
WojciechNagorski, line 33: I believe that the new .NET analyzer does not need the old StyleCope at all. 
`Microsoft.CodeAnalysis.NetAnalyzers` contains everything we need. I set this up in my project last year.

src/Renci.SshNet/.editorconfig
```diff
@@ -0,0 +1,25 @@
+﻿[*.cs]
+
+### StyleCop Analyzers rules ###
+
+# SA1202: Elements must be ordered by access
+# https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1202.md
+dotnet_diagnostic.SA1202.severity = none
```
WojciechNagorski, line 7: Instead of style cope rule we can use rule from `Microsoft.CodeAnalysis.NetAnalyzers`
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0036

drieseng: **IDE0036** is not related to **SA1202**. **SA1202** is about the order of "elements" (members), not about the order of modifiers.

For example, **SA1202** is reported if a class defines an internal method before a public method.

Directory.Build.props
drieseng: I'm pretty sure that's not correct.

**For example:**
As far as I'm aware none of the [documentation rules](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/DocumentationRules.md) are covered by **Microsoft.CodeAnalysis.NetAnalyzers**.
But there's much more than that.

drieseng:
@WojciechNagorski, ok to merge this?

WojciechNagorski:
@drieseng yes.

WojciechNagorski (commented):
LGTM!

WojciechNagorski (approved):

src/Renci.SshNet/.editorconfig
WojciechNagorski: @drieseng You're right. I misremembered the assumptions I made in my project. I resigned from Stylecop, but I have a reason for that. I agree with you that the rule SA1202 is useful. 
Sorry for misleading you.

Directory.Build.props
WojciechNagorski: Honestly, that's not true. There are rules for XML documentation. It's true, they're not compatible with StyleCop rules, but they're efficient, supported, and used by the .NET team. Style cope is a little bit outdated. But if you want we can use it too.

```
# CS1574: XML comment has cref attribute that could not be resolved
dotnet_diagnostic.CS1574.severity = error

# CS1573: Parameter has no matching param tag in the XML comment (but other parameters do)
dotnet_diagnostic.CS1573.severity = error

# CS1570: XML comment has badly formed XML
dotnet_diagnostic.CS1570.severity = error

# CS1723: XML comment has cref attribute that refers to a type parameter
dotnet_diagnostic.CS1723.severity = error

# CS1572: XML comment has a param tag, but there is no parameter by that name
dotnet_diagnostic.CS1572.severity = error

# CS1584: XML comment has syntactically incorrect cref attribute
dotnet_diagnostic.CS1584.severity = none

# CS1587: XML comment is not placed on a valid language element
dotnet_diagnostic.CS1587.severity = error

# CS1591: Missing XML comment for publicly visible type or member
dotnet_diagnostic.CS1591.severity = error

# CS1711: XML comment has a typeparam tag, but there is no type parameter by that name
dotnet_diagnostic.CS1711.severity = error

# CS1735: XML comment has a typeparamref tag, but there is no type parameter by that name
dotnet_diagnostic.CS1735.severity = error
```
You need to enable those rules in csproj:

```xml
    <!-- Enable .NET code quality analysis. More info: https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#enablenetanalyzers -->
    <EnableNETAnalyzers>true</EnableNETAnalyzers>

    <!-- Enable .NET code quality analysis during build. More info: https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#enforcecodestyleinbuild -->
    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
```

WojciechNagorski (approved):

PR #1126, opened 2023-05-23
ListDirectoryAsync return IAsyncEnumerable

WojciechNagorski:
Fixes discussion https://github.com/sshnet/SSH.NET/discussions/1113

I have to change:
```xml
<LangVersion>latest</LangVersion>
```
because `IAsyncEnumerable<>` requires C# 8.0, but old .NET Framework supports only C# 7.3.


src/Renci.SshNet/Renci.SshNet.csproj
```diff
@@ -5,20 +5,24 @@
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <AssemblyName>Renci.SshNet</AssemblyName>
     <AssemblyOriginatorKeyFile>../Renci.SshNet.snk</AssemblyOriginatorKeyFile>
-    <LangVersion>7.3</LangVersion>
+    <LangVersion>latest</LangVersion>
```
drieseng: Please rebase. This and other properties have been removed, and are now set [here](https://github.com/sshnet/SSH.NET/blob/develop/Directory.Build.props).

build/build.proj
```diff
@@ -25,6 +25,10 @@
 			<OutputDirectory>Renci.SshNet\bin\$(Configuration)\netstandard2.0</OutputDirectory>
 			<Moniker>netstandard2.0</Moniker>
 		</TargetFrameworkModern>
+		<TargetFrameworkModern Include=".NETStandard 2.1">
```
drieseng, line 28: Please also update the **Framework Support** section in the **README.md**.

src/Renci.SshNet/ISftpClient.cs
```diff
@@ -698,21 +698,23 @@ public interface ISftpClient : IBaseClient, IDisposable
...
         /// <summary>
-        /// Asynchronously retrieves list of files in remote directory.
+        /// Asynchronously enumerates the files in remote directory.
         /// </summary>
         /// <param name="path">The path.</param>
         /// <param name="cancellationToken">The <see cref="CancellationToken"/> to observe.</param>
         /// <returns>
-        /// A <see cref="Task{IEnumerable}"/> that represents the asynchronous list operation.
-        /// The task result contains an enumerable collection of <see cref="SftpFile"/> for the files in the directory specified by <paramref name="path" />.
+        /// An <see cref="IAsyncEnumerable{SftpFile}"/> that represents the asynchronous enumeration operation.
```
drieseng: As far as I'm aware the **IAsyncEnumerable{SftpFile}** syntax doesn't work. Here **SftpFile** (which should have been **ISftpFile** instead) will be considered as the name of the type argument, not a concrete type argument. You must write something in line of:
```c#
An <see cref="IAsyncEnumerable{T}"/> of <see cref="ISftpFile"/> ...
```

src/Renci.SshNet/ISftpClient.cs
```diff
@@ -698,21 +698,23 @@ public interface ISftpClient : IBaseClient, IDisposable
...
+        /// The enumeration contains an async stream of <see cref="SftpFile"/> for the files in the directory specified by <paramref name="path" />.
```
drieseng:
```suggestion
        /// The enumeration contains an async stream of <see cref="ISftpFile"/> for the files in the directory specified by <paramref name="path" />.
```

src/Renci.SshNet/SftpClient.cs
```diff
@@ -530,32 +533,33 @@ public IEnumerable<ISftpFile> ListDirectory(string path, Action<int> listCallbac
...
         /// <summary>
-        /// Asynchronously retrieves list of files in remote directory.
+        /// Asynchronously enumerates the files in remote directory.
         /// </summary>
         /// <param name="path">The path.</param>
         /// <param name="cancellationToken">The <see cref="CancellationToken"/> to observe.</param>
         /// <returns>
-        /// A <see cref="Task{IEnumerable}"/> that represents the asynchronous list operation.
-        /// The task result contains an enumerable collection of <see cref="SftpFile"/> for the files in the directory specified by <paramref name="path" />.
+        /// An <see cref="IAsyncEnumerable{SftpFile}"/> that represents the asynchronous enumeration operation.
```
drieseng: Same remarks as for the interface.

drieseng (changes requested):
Please also add unit tests.
I'd also like integration tests, but we should perhaps first decide if we want to integrate them into the main repo.

WojciechNagorski:
Sftp does not have working unit tests. There are integration tests in the SSH.NET repo. I've added a test for "ListDirectoryAsync" method.
image
As a server, I've used a container from the integration tests repo, of course, I had to change the value of `Resources.HOST, Resources.USERNAME, Resources.PASSWORD`

drieseng:
> Sftp does not have working unit tests. 

Indeed, and the current implementation is not mock friendly.

> There are integration tests in the SSH.NET repo.

I'd prefer to add integration tests to the https://github.com/sshnet/IntegrationTests repo, and subsequently discuss merging these into the main repo.

I'd like to cover at least all argument / state checks (eg. path is null, client is not connected, cancellation is requested, path does not exist) and of course the happy flow (eg. path contains only files that match, path contains both matching and non-matching files and sub-directories, path only contains non-matching files and sub-directories, ...).

I'm not saying YOU need to add all these tests. Let's first decide what we do with the https://github.com/sshnet/IntegrationTests repo and then continue with this one.

I once started a [discussion](https://github.com/sshnet/SSH.NET/discussions/1048) on this topic, but I'm not sure if there's useful information in that thread.

WojciechNagorski:
I prefer to merge this task and then I would like to move integration tests to the main repo and run them for each PR. During this task, I can clean up all integration tests. What do you think? We can discuss the details, but I would like to take care of it.

drieseng:
> I prefer to merge this task and then I would like to move integration tests to the main repo and run them for each PR. During this task, I can clean up all integration tests. What do you think? We can discuss the details, but I would like to take care of it.

I understand, but experience learns me that, if we merge it right now, these tests will never be implemented.

WojciechNagorski:
So it means that I can prepare a proposition of new integration tests? I can do it in separate PR.

WojciechNagorski:
I've added integration tests for `ListDirectoryAsync` with `IAsyncEnumerable` so I think this is ready.

src/Renci.SshNet/Renci.SshNet.csproj
```diff
@@ -2,18 +2,22 @@
   <PropertyGroup>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <AssemblyName>Renci.SshNet</AssemblyName>
-    <TargetFrameworks>net462;netstandard2.0;net6.0;net7.0</TargetFrameworks>
+	<TargetFrameworks>net462;netstandard2.0;netstandard2.1;net6.0;net7.0</TargetFrameworks>
```
Rob-Hague, line 5: What was the motivation to add a netstandard2.1 target, given that net6.0 is the lowest supported platform that can use it?

Should the test projects have a net5.0 (or earlier) target in order to test the netstandard2.1 target? Is it worth it for a new feature?

src/Renci.SshNet.Tests/Classes/SftpClientTest.ListDirectory.cs
```diff
@@ -89,6 +93,30 @@ public void Test_Sftp_ListDirectory_Current()
             }
         }
 
+#if NET6_0_OR_GREATER
+        [TestMethod]
+        [TestCategory("Sftp")]
+        [TestCategory("integration")]
```
Rob-Hague, line 99: Is this test unnecessary now that there is an integration tests project?

WojciechNagorski: Yes, I plan to rewrite all integration tests in the new approach, so this will be removed.

src/Renci.SshNet/Renci.SshNet.csproj
WojciechNagorski: According to https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-1#select-net-standard-version
.NET Standard 2.1 is broader than just modern .NET. For instance, Unity supports .NET Standard 2.1. 

I am open to discussing this.

Rob-Hague: I did not think of that. But given any platform supporting netstandard2.1 would also support netstandard2.0, the only advantage is the extra APIs available e.g. `IAsyncEnumerable`.

And if we are adding new features using those APIs, at some point the implementation may differ from the .NET implementation, and we then probably ought to have test coverage by using an older target.

That doesn't apply here, so while I would vote to drop netstandard2.1, I think it's fine for now.

PR #1134, opened 2023-05-29
Preview of integration tests

WojciechNagorski:
This is only an example code of integration tests and a presentation of how `testcontainer` library works.


src/Renci.SshNet.IntegrationTests/TestsFixtures/InfrastructureCollection.cs
```diff
@@ -0,0 +1,40 @@
+﻿// <copyright file="InfrastructureCollection.cs" company="Intel Corporation">
+//  ========================================================================
```
drieseng, line 2: Ahum???

WojciechNagorski:
I will close this PR and I will open a PR that using MSTest without FluentAssertion.

PR #1137, opened 2023-06-03
Replace Array<T>.Empty with Array.Empty<T>()

Rob-Hague:
Great to see some work going on. Here is a small contribution.


drieseng (approved):
Nice! Not sure how I missed that one. it's good to have an extra pair of eyes :)

PR #1140, opened 2023-06-05
Delete dead code

Rob-Hague:
Only internal stuff. That was fun.


drieseng:
@Rob-Hague, this only concerns code that we imported from other OSS projects. Updating this code may not be a good idea (or even allowed), as it makes it harder to diff new drops of this code in the future.

@WojciechNagorski, what's your opinion on this?

WojciechNagorski:
@Rob-Hague Thanks for your work but if this code is imported from an OSS project, we should not remove or modify it.

Rob-Hague:
Fair enough, perhaps it is worth considering instead taking a dependency on the [BouncyCastle nuget package](https://www.nuget.org/packages/BouncyCastle.Cryptography), now that there is one.

Rob-Hague:
@WojciechNagorski have you got a rough plan or list of priorities in mind for the repo? I would like to help where I can.

I would think something like:

1. Test infrastructure
2. Support rsa-sha2 key host algorithms
3. Publish a new release
4. ???

WojciechNagorski:
@Rob-Hague Hi, For now, I have to move all integration tests (from https://github.com/sshnet/IntegrationTests and from unit tests) to the new integration project. Then publish a new release. After that, we can think about RSA-SHA2.

WojciechNagorski:
@Rob-Hague If you want to help me, you can contact me by e-mail.

PR #1143, opened 2023-06-10
Delete performance/longrunning tests

Rob-Hague:
These do not seem very useful and are annoying to have to filter out when running tests via VS


src/Renci.SshNet.Tests/Classes/Common/ASCIIEncodingTest.cs
Rob-Hague, line 1: Deleted this whole file: the custom ASCII class no longer exists, so it is just testing `System.Text.Encoding.ASCII`

drieseng:
@Rob-Hague, would you consider transforming these tests into benchmarks (in a new project)?

Rob-Hague:
Yes that would be good to have. I can work on it (at some point). Although I would still suggest that these particular benchmarks/tests do not add much value

WojciechNagorski:
LGTM! Only one question. Did you add benchmarks for all these things in #1151

Rob-Hague:
I did not, but I will

WojciechNagorski (approved):

PR #1148, opened 2023-06-29
MSTest Integration tests

WojciechNagorski:
Just an example of new integration tests, after this PR I going to move all integration tests from IntegrationTests repo.


Rob-Hague:
Great job!

May I suggest merging with 'squash-merge'? I believe that is what has been used in the past (and IMO is cleaner)

WojciechNagorski:
Hi @Rob-Hague 
Thanks for the information. This was my fault. I will improve for the future.

Rob-Hague:
No problem 🙂

PR #1156, opened 2023-08-03
PR #971 with conflicts resolved and building

et1975:
This brings #971 up to date and hopefully nudging along the solution for #825.

### DIFF
--- a/src/Renci.SshNet.Tests/Classes/BaseClientTest_Connect_OnConnectedThrowsException.cs
+++ b/src/Renci.SshNet.Tests/Classes/BaseClientTest_Connect_OnConnectedThrowsException.cs
@@ -140,7 +140,7 @@ namespace Renci.SshNet.Tests.Classes
             using (var s = executingAssembly.GetManifestResourceStream(string.Format("Renci.SshNet.Tests.Data.{0}", "Key.RSA.txt")))
             {
                 var privateKey = new PrivateKeyFile(s);
-                return (KeyHostAlgorithm) privateKey.HostKey;
+                return privateKey.HostKeys[0];
             }
         }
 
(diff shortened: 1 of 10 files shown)


et1975:
@Rob-Hague or anyone else can comment/approve?

Rob-Hague:
> @Rob-Hague or anyone else can comment/approve?

I don't speak with any authority but I have given this problem some thought so I will leave my opinion here.

It's great that there is a (presumably) working solution already. One can pull down the code and build the library to use it.

However I do not think the implementation is good enough to be merged into the repo, mainly with regard to the duplication of the `RsaKey` and `RsaDigitalSignature` classes. I know why it was done this way: a `Key` is contracted to provide a `DigitalSignature` but all `DigitalSignature` implementations require a `Key`. So you end up with the implementations coupled together. Ideally `RsaDigitalSignature` would take a `HashAlgorithm` (or similar) in the constructor but then this needs to be passed through `RsaKey`, which needn't be concerned with the implementation of a signature.

I think the solution is to remove the `Sign` and `Verify` methods from `Key`: ultimately a key itself does not sign something. But that could be a rather big (breaking) change, including on other types like `KeyHostAlgorithm`. If I think a bit harder I will consider opening an issue with a proposal.

WojciechNagorski:
@et1975 I'm waiting for more permissions for this project and intend to reactivate this project once I get them.

WojciechNagorski:
@Rob-Hague Can you propose your solution/PR? I already have repo and Nuget permissions so we can handle this topic.

et1975:
Closing in favour of #1177

WojciechNagorski:
@et1975 Did you test #1177 ?

et1975:
I'm not using this library directly, will have to see what it means for [Posh-SSH](https://github.com/darkoperator/Posh-SSH/issues/537).

WojciechNagorski:
Ok, we need to test everything and confirm if it's ok and then we'll release it. Then PoshSSH can update.

Rob-Hague:
FYI I have tested it against several different servers (including servers which are not OpenSSH)

WojciechNagorski:

WojciechNagorski:

PR #1158, opened 2023-08-21
Create scorecards-analysis workflow

DeChrist:


WojciechNagorski:
@DeChrist Can you describe what these changes do and why?

DeChrist:
Hi @WojciechNagorski,

This PR was inadvertently pushed to the main repo rather than my fork. I'll be closing the PR.

That said, in response to your question: considering the security requirements associated with the repo, the goal is to integrate the standard security checks from [OpenSSF Scorecard](https://securityscorecards.dev/#what-is-openssf-scorecard). This way, we can transparently communicate to package consumers about the security posture of the project, aiding them in gauging the risks involved when adding it as a dependency.

PR #1160, opened 2023-08-24
Enable C# 11, System.Span, and utilize UTF-8 literals

iamcarbon:
This PR enables the use of C# 11 and support for Span<T> within the library.


iamcarbon:
@drieseng Let me know your thoughts on enabling C# 11 and Span<T> support within the library. 

CC @Rob-Hague

WojciechNagorski:
.NET Standard and .NET Framework do not support C# 11. I know that in this case, it works but it also allows to use of features that will never work on .NET Framework and .NET Standard. 

Target | Version | C# language version default
-- | -- | --
.NET | 7.x | C# 11
.NET | 6.x | C# 10
.NET | 5.x | C# 9.0
.NET Core | 3.x | C# 8.0
.NET Core | 2.x | C# 7.3
.NET Standard | 2.1 | C# 8.0
.NET Standard | 2.0 | C# 7.3
.NET Standard | 1.x | C# 7.3
.NET Framework | all | C# 7.3

[Table source](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version)

The proposed changes do not bring enough value to the project to offset the disadvantages. The Span can be also used without C# 11.

PR #1161, opened 2023-08-24
Enable getting and deleting symbolic links

fbedoyao:
The current functions in the `SftpClient `class don’t support interacting with symbolic links directly. By adding an additional parameter, `getRealPath`, to the `GetCanonicalPath `function, we can get a reference to the symbolic link instead of a reference to the file that it points to. The default value for this parameter is set to `true `so that the original behavior is maintained for existing code. 

The added functions `GetSymbolicLink`, `SymbolicLinkExists`, and `DeleteSymbolicLink` use `GetCanonicalPath `with `getRealPath `set to `false `to get and delete symbolic links.

### DIFF
--- a/src/Renci.SshNet/ISftpClient.cs
+++ b/src/Renci.SshNet/ISftpClient.cs
@@ -492,6 +492,18 @@ namespace Renci.SshNet
         void DeleteFile(string path);
 
         /// <summary>
+        /// Deletes remote symbolic link specified by path.
+        /// </summary>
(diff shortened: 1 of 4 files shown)


WojciechNagorski:
@fbedoyao It looks great. Can you add an integration test for those methods? 
Here you can find an example: https://github.com/sshnet/SSH.NET/blob/7cd0487785c82730ecfe277cb9964ab2e101a4af/src/Renci.SshNet.IntegrationTests/SftpClientTests.cs#L53

PR #1165, opened 2023-08-30
Fix the assembly resolution build warning

Rob-Hague:
I do not believe app.config bindingRedirects work on .NET Core (despite what the build output suggests to do) and so the build still gives resolution warnings:

```
warning MSB3243: No way to resolve conflict between "Renci.SshNet, Version=2023.0.0.0, Culture=neutral, PublicKeyToken=1cee9f8bde3db106" and "Renci.SshNet, Version=2020.0.2.0, Culture=neutral, PublicKeyToken=1cee9f8bde3db106". Choosing "Renci.SshNet, Version=2023.0.0.0, Culture=neutral, PublicKeyToken=1cee9f8bde3db106" arbitrarily. 
Consider app.config remapping of assembly "Renci.SshNet, Culture=neutral, PublicKeyToken=1cee9f8bde3db106" from Version "2020.0.2.0" [C:\Users\rhague\.nuget\packages\ssh.net\2020.0.2\lib\netstandard2.0\Renci.SshNet.dll] to Version "2023.0.0.0" [C:\Users\rhague\source\gitrepos\SSH.NET\src\Renci.SshNet\bin\Debug\net7.0\Renci.SshNet.dll] to solve conflict and get rid of warning.
```

This undoes the bindingRedirects and instead excludes the transitive package with an explicit PackageReference.

The project builds cleanly for me after this, and I have inspected the msbuild binlog to verify the correct assembly is chosen.


Rob-Hague:
My usual workstation is a VM on which I cannot run docker, so I have not yet tried running the tests (I will do later), but there should be no difference: the 2023.0.0.0 assembly was already being chosen, just 'arbitrarily' as the logs suggest, and now it is explicit.

WojciechNagorski (approved):
It works! I didn't know `ExcludeAssets="All"`

WojciechNagorski:
@Rob-Hague Big thanks!

PR #1166, opened 2023-08-30
Use CollectionAssert in ListDirectory tests

Rob-Hague:
These tests were failing for me because the files were coming in in a different order

Plus, `CollectionAssert.AreEquivalent` gives a nice failure message


WojciechNagorski:
Can you tell me in what order you got the results?
I don't have full knowledge about the project yet, but perhaps the problem is not checking, but a random order.

Rob-Hague:
```
 Create_directory_with_contents_and_list_it
   Source: SftpClientTests.cs line 21
   Duration: 172 ms

  Message: 
Assert.AreEqual failed. Expected:</home/sshnet/sshnet-test/. False True
/home/sshnet/sshnet-test/.. False True
/home/sshnet/sshnet-test/test-file.txt True False
>. Actual:</home/sshnet/sshnet-test/test-file.txt True False
/home/sshnet/sshnet-test/. False True
/home/sshnet/sshnet-test/.. False True
>. 
```

Same for `Create_directory_with_contents_and_list_it_async`

I have not looked at the implementation, but I would not expect it to do any sorting - the user can sort the files how they like.

WojciechNagorski:
It needs to be checked. I don't know if it shouldn't be deterministic.

Rob-Hague:
So the directory listing functionality is the sending of a `SSH_FXP_READDIR` packet. I do not see anything relating to sort order in the [RFC](https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-13#section-8.2.2) and the library implementation is reading out the filenames as they are returned (into an array, and then a list).

https://github.com/sshnet/SSH.NET/blob/9f1699b5602807e50b8ea6b4300a12ff9d00f93f/src/Renci.SshNet/SftpClient.cs#L2276

https://github.com/sshnet/SSH.NET/blob/9f1699b5602807e50b8ea6b4300a12ff9d00f93f/src/Renci.SshNet/Sftp/SftpSession.cs#L1196

https://github.com/sshnet/SSH.NET/blob/9f1699b5602807e50b8ea6b4300a12ff9d00f93f/src/Renci.SshNet/Sftp/Responses/SftpNameResponse.cs#L27-L44


It looks fine to me.

Rob-Hague:
Also seen in #1220 

```
  Failed Create_directory_with_contents_and_list_it [128 ms]
  Error Message:
   Assert.AreEqual failed. Expected:</home/sshnet/sshnet-test/. False True
/home/sshnet/sshnet-test/.. False True
/home/sshnet/sshnet-test/test-file.txt True False
>. Actual:</home/sshnet/sshnet-test/.. False True
/home/sshnet/sshnet-test/test-file.txt True False
/home/sshnet/sshnet-test/. False True
>. 
  Stack Trace:
     at Renci.SshNet.IntegrationTests.SftpClientTests.Create_directory_with_contents_and_list_it() in /home/appveyor/projects/ssh-net/test/Renci.SshNet.IntegrationTests/SftpClientTests.cs:line 48
   at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
```

test/Renci.SshNet.IntegrationTests/SftpClientTests.cs
```diff
@@ -34,21 +34,21 @@ public void Create_directory_with_contents_and_list_it()
             Assert.IsTrue(_sftpClient.Exists(testFilePath));
 
             // Check if ListDirectory works
-            var files = _sftpClient.ListDirectory(testDirectory);
+            var expectedFiles = new List<(string FullName, bool IsRegularFile, bool IsDirectory)>()
+            {
```
drieseng: I'm aware that you do not agree with this, but please indent the braces.

test/Renci.SshNet.IntegrationTests/SftpClientTests.cs
```diff
@@ -69,21 +69,24 @@ public async Task Create_directory_with_contents_and_list_it_async()
             Assert.IsTrue(_sftpClient.Exists(testFilePath));
 
             // Check if ListDirectory works
-            var files = _sftpClient.ListDirectoryAsync(testDirectory, CancellationToken.None);
+            var expectedFiles = new List<(string FullName, bool IsRegularFile, bool IsDirectory)>()
+            {
```
drieseng: Same.

drieseng (changes requested):

test/Renci.SshNet.IntegrationTests/SftpClientTests.cs
Rob-Hague: I also don't like using `var`, but I don't mind ;-)

WojciechNagorski:
It is strange, but I've just got the same error ;)
image

WojciechNagorski (approved):

WojciechNagorski:

PR #1170, opened 2023-09-08
Abstract out the hash algorithm from RsaDigitalSignature

Rob-Hague:
This is a step towards implementing the rsa-sha2 key host algorithms


src/Renci.SshNet.Tests/Classes/Security/Cryptography/RsaDigitalSignatureTest.cs
```diff
@@ -11,17 +16,152 @@ namespace Renci.SshNet.Tests.Classes.Security.Cryptography
...
+                0x01, 0xd3, 0x67, 0x0d, 0x8d, 0x85, 0x7b, 0x51, 0x14, 0xec, 0x26, 0x53, 0x00, 0x61, 0x25, 0x16,
+                0x19, 0x10, 0x3c, 0x86, 0x16, 0x59, 0x84, 0x08, 0xd1, 0xf9, 0x1e, 0x05, 0x88, 0xbd, 0x4a, 0x01,
+                0x43, 0x4e, 0xec, 0x76, 0x0b, 0xd7, 0x2c, 0xe9, 0x98, 0xb1, 0x4c, 0x0a, 0x13, 0xc6, 0x95, 0xf9,
+                0x8f, 0x95, 0x5c, 0x98, 0x4c, 0x8f, 0x97, 0x4a, 0xad, 0x0d, 0xfe, 0x84, 0xf0, 0x56, 0xc3, 0x29,
+                0x73, 0x75, 0x55, 0x3c, 0xd9, 0x5e, 0x5b, 0x6f, 0xf9, 0x81, 0xbc, 0xbc, 0x50, 0x75, 0x7d, 0xa8
+            }, signedBytes);
+
+
+            // The following fails due to the _isPrivate decision in RsaCipher.Transform. Is that really correct?
+            //Assert.IsTrue(digitalSignature.Verify(data, signedBytes));
```
Rob-Hague, line 57: This is surely a bug, but I would rather wait until the test migrations are done before looking at it.

The `Encrypt` and `Decrypt` methods on `RsaCipher` both call the private `Transform` method which changes its behaviour based (only) upon whether the `RsaKey` member has private key information:

https://github.com/sshnet/SSH.NET/blob/9f1699b5602807e50b8ea6b4300a12ff9d00f93f/src/Renci.SshNet/Security/Cryptography/Ciphers/RsaCipher.cs#L119

So `Encrypt` and `Decrypt` both perform encryption (m ^ e mod N) when given a key with only public information, and both perform decryption when given a key with private key information. They only differ by some processing of padding.

It works because the base class `CipherDigitalSignature` calls `Encrypt` when signing, when really the signing operation is 'decryption' and the verification operation is to encrypt (but `Verify` calls `Decrypt`).

https://github.com/sshnet/SSH.NET/blob/9f1699b5602807e50b8ea6b4300a12ff9d00f93f/src/Renci.SshNet/Security/Cryptography/CipherDigitalSignature.cs#L61

src/Renci.SshNet.Tests/Classes/Security/Cryptography/RsaDigitalSignatureTest.cs
```diff
@@ -11,17 +16,156 @@ namespace Renci.SshNet.Tests.Classes.Security.Cryptography
...
+                0x2e, 0x32, 0x6c, 0x72, 0x06, 0xf9, 0x7f, 0xa1, 0x6c, 0x2e, 0x13, 0x06, 0x41, 0xaa, 0x23, 0xdd,
+                0xc8, 0x1c, 0x61, 0xb6, 0x96, 0x87, 0xc4, 0x84, 0xc8, 0x61, 0xec, 0x4e, 0xdd, 0x49, 0x9e, 0x4f,
+                0x0d, 0x8c, 0xf1, 0x7f, 0xf2, 0x6c, 0x73, 0x5a, 0xa6, 0x3b, 0xbf, 0x4e, 0xba, 0x57, 0x6b, 0xb3,
+                0x1e, 0x6c, 0x57, 0x76, 0x87, 0x9f, 0xb4, 0x3b, 0xcb, 0xcd, 0xe5, 0x10, 0x7a, 0x4c, 0xeb, 0xc0,
+                0xc4, 0xc3, 0x75, 0x51, 0x5f, 0xb7, 0x7c, 0xbc, 0x55, 0x8d, 0x05, 0xc7, 0xed, 0xc7, 0x52, 0x4a
+            }, signedBytes);
+
+
+            // The following fails due to the _isPrivate decision in RsaCipher.Transform. Is that really correct?
+            //Assert.IsTrue(digitalSignature.Verify(data, signedBytes));
```
WojciechNagorski, line 101: Any idea why this isn't working?

src/Renci.SshNet.Tests/Classes/Security/Cryptography/RsaDigitalSignatureTest.cs
```diff
@@ -11,17 +16,156 @@ namespace Renci.SshNet.Tests.Classes.Security.Cryptography
...
+                0xf8, 0x2f, 0x73, 0x01, 0xc7, 0x8d, 0x37, 0xef, 0x98, 0x69, 0xc2, 0xe2, 0x7a, 0x11, 0xd5, 0xb8,
+                0xc9, 0x35, 0x45, 0xcb, 0x56, 0x4b, 0x92, 0x4a, 0xe0, 0x4c, 0xd6, 0x82, 0xae, 0xad, 0x5b, 0xe9,
+                0x40, 0x7e, 0x2a, 0x48, 0x7d, 0x57, 0xc5, 0xfd, 0xe9, 0x98, 0xe0, 0xbb, 0x09, 0xa1, 0xf5, 0x48,
+                0x45, 0xcb, 0xee, 0xb9, 0x99, 0x81, 0x44, 0x15, 0x2e, 0x50, 0x39, 0x64, 0x58, 0x4c, 0x34, 0x86,
+                0xf8, 0x81, 0x9e, 0x1d, 0xb6, 0x97, 0xe0, 0xce, 0x16, 0xca, 0x20, 0x46, 0xe9, 0x49, 0x8f, 0xe6,
+                0xa0, 0x23, 0x08, 0x80, 0xa6, 0x37, 0x70, 0x06, 0xcc, 0x8f, 0xf4, 0xa0, 0x74, 0x53, 0x26, 0x38
+            }, signedBytes);
+
+            // The following fails due to the _isPrivate decision in RsaCipher.Transform. Is that really correct?
+            //Assert.IsTrue(digitalSignature.Verify(data, signedBytes));
```
WojciechNagorski, line 144: Any idea why this isn't working?

src/Renci.SshNet.Tests/Classes/Security/Cryptography/RsaDigitalSignatureTest.cs
```diff
@@ -11,17 +16,156 @@ namespace Renci.SshNet.Tests.Classes.Security.Cryptography
...
+                0x73, 0x75, 0x55, 0x3c, 0xd9, 0x5e, 0x5b, 0x6f, 0xf9, 0x81, 0xbc, 0xbc, 0x50, 0x75, 0x7d, 0xa8
+            };
+
+            CollectionAssert.AreEqual(expectedSignedBytes, signedBytes);
+
+            // Also verify RsaKey uses SHA-1 by default
+            CollectionAssert.AreEqual(expectedSignedBytes, rsaKey.Sign(data));
+
+            // The following fails due to the _isPrivate decision in RsaCipher.Transform. Is that really correct?
+            //Assert.IsTrue(digitalSignature.Verify(data, signedBytes));
```
WojciechNagorski, line 57: Any idea why this isn't working?

WojciechNagorski (commented):
Overall it looks good, but I don't know why `digitalSignature.Verify(data, signedBytes)` doesn't work. Any idea?

Rob-Hague:
> Overall it looks good, but I don't know why `digitalSignature.Verify(data, signedBytes)` doesn't work. Any idea?

Yes, see my comment above. The encrypt/decrypt methods change based on whether the key has private key information. I think it should be changed, but it seems like long-standing behaviour, and I would rather not change it here.

WojciechNagorski:
I'm afraid to merge without further changes. I agree to these changes, but I don't know if they will fit nicely into further changes. Maybe you want to complete the entire ssh-sha256 and ssh-sha-512 implementation on this PR?

Rob-Hague:
Sure, the rest of the changes are based on this so the PR will include this either way.

WojciechNagorski:
I know. Great!

Rob-Hague:
Included in #1177

WojciechNagorski:

WojciechNagorski:

PR #1173, opened 2023-09-09
Move Integration tests

WojciechNagorski:


src/Renci.SshNet.IntegrationTests/App.config
WojciechNagorski, line 1: Does it work?

src/Renci.SshNet.IntegrationTests/Common/AsyncSocketListener.cs
```diff
@@ -0,0 +1,393 @@
+﻿using System.Net;
+using System.Net.Sockets;
+
+namespace Renci.SshNet.IntegrationTests.Common
+{
+    public class AsyncSocketListener : IDisposable
```
WojciechNagorski, line 6: This is a copy of `AsyncSocketListener` from unit tests. We can share this class

PR #1179, opened 2023-09-22
Use OS-agnostic socket error codes to allow tests run on different OSes

dimhotepus:
`SocketErrorCode` is OS agnostic, `ErrorCode` is OS specific.

On Windows `ErrorCode` = `(int) SocketErrorCode`, but on Mac and Unix it is not.

For example `ExitCode` for `HostNotFound (11001)` on Windows is `11001`, on Mac & Unix is `-131073`.
So testing for `ExitCode == (int) HostNotFound` fails on Mac & Unix.


WojciechNagorski:
I will check this soon. Big thanks for your contribution!

WojciechNagorski (approved):

PR #1182, opened 2023-09-23
Use ExceptionDispatchInfo in more places

Rob-Hague:


WojciechNagorski:
Big thanks! This kind of PR are really important for everyone!

PR #1183, opened 2023-09-23
Remove placeholder tests

Rob-Hague:
These have no value and make it harder to trace references with intellisense. Removing them also helps narrow down the tests we need to address

Before:

image

After:

image

(the 3 extra passing tests is `ExtensionsTest_Reverse.Null` for each tfm and the 2 failing tests are #1166)


WojciechNagorski:
@drieseng I very like this idea. I want to generate a code coverage raport to the next version and for each PR.

src/Renci.SshNet.Tests/Classes/Common/ExtensionsTest_Reverse.cs
```diff
@@ -61,24 +60,6 @@ public void Small()
             Assert.AreEqual(1, value[4]);
             Assert.AreEqual(0, value[5]);
         }
-
-        [TestMethod]
-        [Ignore]
-        public void Perf_Large()
```
drieseng, line 67: Move to benchmark project?

drieseng (changes requested):

drieseng:
@Rob-Hague, @WojciechNagorski: I suppose we want to merge this one before we merge the PR where we move the test projects.

WojciechNagorski:
@drieseng I think it's a good idea.

src/Renci.SshNet.Tests/Classes/Common/ExtensionsTest_Reverse.cs
Rob-Hague: Ok, done.

WojciechNagorski (approved):

PR #1184, opened 2023-09-24
Rebased AsyncResult fix

### LINKED ISSUE #329: UploadedBytes wrong value

Rob-Hague:
This is #680 by @miroslavpokorny brought up to date.

I have been seeing occasional failures of the `Test_Sftp_Multiple_Async_Upload_And_Download_10Files_5MB_Each` test.

When choosing "Run until failure" in Visual Studio, the test failed after 20, 9, 45, 32 and 10 attempts. With this change, it ran successfully over 300 times before I stopped it.

Fixes #329 

Side note: I'm not really sure why these Begin/End methods have an extra "callback" parameter as well as an `AsyncCallback` parameter, but I'm hoping we can just get rid of these methods soon anyway, or at least have them [wrap](https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/System/Threading/Tasks/TaskToAsyncResult.cs) a task based implementation

### DIFF
--- a/src/Renci.SshNet/SftpClient.cs
+++ b/src/Renci.SshNet/SftpClient.cs
@@ -583,7 +583,7 @@ namespace Renci.SshNet
         {
             CheckDisposed();
 
-            return InternalListDirectory(path, listCallback);
+            return InternalListDirectory(path, asyncResult: null, listCallback);
         }
 
         /// <summary>
@@ -666,12 +666,7 @@ namespace Renci.SshNet
             {
                 try
                 {
-                    var result = InternalListDirectory(path, count =>
-                    {
-                        asyncResult.Update(count);
-
-                        listCallback?.Invoke(count);
-                    });
+                    var result = InternalListDirectory(path, asyncResult, listCallback);
 
                     asyncResult.SetAsCompleted(result, completedSynchronously: false);
                 }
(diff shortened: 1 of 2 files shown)


Rob-Hague:
I am seeing strange failures of this test locally. The failures are not the original "uploaded/downloaded bytes do not match" assertion, rather some kind of channel closed/timeout exception such as "Renci.SshNet.Common.SshOperationTimeoutException: Session operation has timed out".

I thought I had narrowed it down and was able to reproduce the behaviour, but then it started behaving differently and now I have no idea what's going on.

WojciechNagorski:
Can you remove this in this PR https://github.com/sshnet/SSH.NET/blob/9b0262c5fb05da2f3e15c09677b3a8cadae262ca/test/Renci.SshNet.IntegrationTests/OldIntegrationTests/SftpClientTest.Upload.cs#L236

Rob-Hague:
Thanks, I had done that locally but forgot to push...

WojciechNagorski:
@Rob-Hague It looks like it works. Is it ready for review?

Rob-Hague:
Yes I think so, I think there were gremlins in my PC

WojciechNagorski (approved):

PR #1187, opened 2023-09-26
WIP: Run tests both on Windows & Linux

dimhotepus:
* Run tests both on Windows and Linux hosts.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,14 +12,35 @@ build:
   project: src\Renci.SshNet.sln
   verbosity: minimal
 
-configuration: Release
+configuration:
+  - Debug
+  - Release
 
-test_script:
-- cmd: >-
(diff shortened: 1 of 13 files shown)


dimhotepus:
Please, try to run Appveyor on this MR. Trying to add testing on Linux host.

Rob-Hague:
Nice.

Could you please keep it in Debug mode? There are a number of `Debug.Assert` calls that we would want to fire if they were hit.

Could you also try to run the integration tests? I believe the appveyor images have docker installed. I have been wondering if just `dotnet test src/Renci.SshNet.sln` is all that is needed (or if appveyor would even do that automatically without the command)

WojciechNagorski:
@dimhotepus Thanks! Great work. I will use this PR.

WojciechNagorski:
I will continue this work on #1250 
Big thenks @dimhotepus

PR #1189, opened 2023-09-27
Add/migrate hmac+cipher integration tests

Rob-Hague:
Add a couple of hmac tests, use default `ConnectionInfo`. I had added the other algorithms defined in `ConnectionInfo` but they were mostly removed in [OpenSSH 7.6](https://www.openssh.com/txt/release-7.6).


WojciechNagorski (approved):
LGTM

PR #217, opened 2017-05-04
Implement SetLastAccessTime, SetLastWriteTime

coder0xff:

PR #1194, opened 2023-09-29
Implement set last write and access time

### LINKED ISSUE #228: SetLastWriteTime and SetLastWriteTimeUtc not implemented
This is documented in the code, these methods throw a NotImplementedException. I'd rather have them do what they're supposed to do. Is this a complicated feature to implement?

Or, put it another way, can I upload a local file and keep its last write time?

WojciechNagorski:
Closes #228
Closes #217

### DIFF
--- a/src/Renci.SshNet.IntegrationTests/SftpTests.cs
+++ b/src/Renci.SshNet.IntegrationTests/SftpTests.cs
@@ -6137,6 +6137,134 @@ namespace Renci.SshNet.IntegrationTests
             }
         }
 
+        [TestMethod]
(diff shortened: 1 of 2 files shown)


Rob-Hague:
Given that `SftpFileAttributes` already handles UTC/local times, maybe we can delete the `Set*Utc` variants so we just have `SetLastAccessTime` and `SetLastWriteTime`?

https://github.com/sshnet/SSH.NET/blob/cd1151decac9f5672411eebb12d939cedc53b317/src/Renci.SshNet/Sftp/SftpFileAttributes.cs#L92-L103

PR #1203, opened 2023-10-12
Move test projects to test folder

drieseng:
Move test projects to **test** folder.
Move **global.json** to root of repo.
Update **Solution Items** in solution.


drieseng:
@WojciechNagorski, upgrade to version **7.0.400** of the .NET SDK?

WojciechNagorski:
@drieseng yes you can update. Today I will not able to review this branch but you can merge it. This is a great idea.

Rob-Hague:
Does it make more sense to just have everything under the src/ folder? The benchmarks project is a bit ambiguous.

Before merging this one, would you mind deciding on #1183 ? Would be easier to merge that one first if it's going to be.

Rob-Hague:
> Does it make more sense to just have everything under the src/ folder? The benchmarks project is a bit ambiguous.

To clarify, I am suggesting to just delete the test/ folder and merge the "shared" tests in that folder into the UnitTest project

WojciechNagorski:
TBH I don't care about it. For me we can have everything in /SRC, or we can split this.

WojciechNagorski:
#1183 I would like to merge this but I didn't have time for this.

drieseng:
I see two options:

1.  Have a **src** and **test** folder in the root of the repo.
   That is what this PR is doing.

1. Have a **main** and **test** folder below the **src** folder.
   All test related projects would be moved to the **src/test** folder.
   This closely resembles the [Standard Directory Layout](https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html) of **Maven**.

I strongly recommend to have a **test** folder that groups all test-related projects.
That way you can have an **.editorconfig** and **Directory.Build.props** in that folder which only applies to these test-related projects.

I propose to rename **Renci.SshNet.Benchmarks** to **Renci.SshNet.PerformanceTests**.

Rob-Hague:
Ok, thanks, I agree it is a good thing to have them separated in order to use different .editorconfig etc.

On that basis the changes here are good for me. And I am happy for the benchmarks project to be renamed.

test/Renci.SshNet.Tests/Data/Key.ECDSA.Encrypted.txt
Rob-Hague, line 1: The benchmarks project also references the Data/ folder

drieseng: Nice catch. I'll move it then.

drieseng: Should be ok now.

drieseng:
Still need to decide which of the two options we want to use.

WojciechNagorski:
I prefer the first option, so I think this PR is a good idea. There is a useful link https://gist.github.com/davidfowl/ed7564297c61fe9ab814

I have encountered other approaches many times and I can work with them so that's why I won't defend a specific solution.

WojciechNagorski:
@drieseng Did you see https://github.com/sshnet/SSH.NET/pull/1183? What do you think about it?

WojciechNagorski:
I've merge this PR. I wanted to revresh this PR after merging other PR. I've also had to fix integration tests, but I was unable to change this PR, even thow I've merge changes to this PR there was still conflicts.

global.json
Rob-Hague, line 1: Was there a reason for this? I don't think we have any dependency on a particular sdk version (other than 7.x.x or above).

PR #1209, opened 2023-10-14
Remove unnecessary finalizers

### LINKED ISSUE #526: Remove meaningless finalizers from library classes
Many library classes feature meaningless finalizers that do nothing at runtime but grow the finalization queue and thus slow down GC and application performance.
This is especially concerning for frequently-used classes like **SshCommand** and **Channel**.
The library classes don't need finalizers as they aren't dealing with allocation and deallocation of unmanaged resource handles.
The nuances of a correct **IDisposable** implementation are neatly summarized here:
https://github.com/sidristij/dotnetbook/blob/master/book/en/LifetimeManagement/2-Disposable.md

Rob-Hague:
All of the finalizers in the library are no-ops, but their existence means that when `Dispose` (and thus `GC.SuppressFinalize`) is not called, the objects' lifetimes are extended unnecessarily while they are waiting to be finalized.


WojciechNagorski:
I could be wrong, but I think that if we call `GC.SuppressFinalize(this)` inside the `Dispose()` method it tells GC that it does not need to run the finalizer at all. If so, nothing will change because of this PR. @Rob-Hague What do you think about it?

https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose

Rob-Hague:
You are correct. The PR is just small clean-up in the case that `Dispose` is not called.

WojciechNagorski:
@drieseng What do you think about this PR.

drieseng:
> @drieseng What do you think about this PR.

I'm ok with removing the finalizer when there's no added value, but it made sense to keep the finalizer for the few cases I reviewed (**BaseClient**, **SftpFileReader** and **PasswordConnectionInfo**).

Rob-Hague:
> > @drieseng What do you think about this PR.
> 
> I'm ok with removing the finalizer when there's no added value, but it made sense to keep the finalizer for the few cases I reviewed (**BaseClient**, **SftpFileReader** and **PasswordConnectionInfo**).

None of them make use of any unmanaged resources, so it doesn't make sense for them to have finalizers.

WojciechNagorski (approved):

Issue #1167, opened 2023-08-31
System.MissingMethodException: Cannot dynamically create an instance of type 'Renci.SshNet.Security.KeyExchangeECCurve25519'. Reason: No parameterless constructor defined.


Zero-Xiong:
Hi guys, what's the root cause?

Version:  2020.0.2

System.MissingMethodException: Cannot dynamically create an instance of type 'Renci.SshNet.Security.KeyExchangeECCurve25519'. Reason: No parameterless constructor defined.
   at Renci.SshNet.Session.WaitOnHandle(WaitHandle, TimeSpan)
   at Renci.SshNet.Session.Connect()
   at Renci.SshNet.BaseClient.CreateAndConnectSession()

Rob-Hague:
Are you trimming your application?

emetlashevskiy:
I have some error, how resolve it?
On win64 work fine, but on linux:
System.MissingMethodException: Cannot dynamically create an instance of type 'Renci.SshNet.Security.KeyExchangeECCurve25519'. Reason: No parameterless constructor defined.
   at Renci.SshNet.Session.WaitOnHandle

emetlashevskiy:
Yes, if untrimmed app when work fine!
You need set "PublishTrimmed" to false, or remove from *.csproj.

emetlashevskiy:
Another problem, how disable trimming my app if PublishAOT is true?
I tryed 	  

```
<PackageReference Include="SSH.NET" Version="2020.0.2">
    <IsTrimmable>false</IsTrimmable>
</PackageReference>
```

but no effect.

Rob-Hague:
I don't know a lot about trimming. I think you may try "rooting" the assembly as described at this link

https://learn.microsoft.com/dotnet/core/deploying/trimming/trimming-options?pivots=dotnet-7-0#root-assemblies

(untested)

```xml
<ItemGroup>
  <TrimmerRootAssembly Include="Renci.SshNet" />
</ItemGroup>
```

You could even try using "descriptors" to preserve just the specific types listed here:

https://github.com/sshnet/SSH.NET/blob/004b57ae9e7ca07f12e7192d7987a8e5d950b04c/src/Renci.SshNet/ConnectionInfo.cs#L344-L357

---

I can't see a way on the library side to mark it as incompatible with trimming, short of fixing the actual problem which is the use of reflection here:

[...]

PR #1216, opened 2023-10-16
Enable trim analysis and fix warnings

Rob-Hague:
While the library is not really so modern as to think about trimming and AOT, with the default behaviour to [trim all assemblies](https://learn.microsoft.com/en-us/dotnet/core/compatibility/deployment/7.0/trim-all-assemblies) when trimming an application we have seen issues such as #1167, and I believe https://github.com/sshnet/SSH.NET/issues/859#issuecomment-1764623253 is related.

Fortunately it does not require many changes.


src/Renci.SshNet/Security/Cryptography/RsaDigitalSignature.cs
```diff
@@ -41,13 +41,20 @@ public RsaDigitalSignature(RsaKey rsaKey, HashAlgorithmName hashAlgorithmName)
         /// </returns>
         protected override byte[] Hash(byte[] input)
         {
-            return _hash.ComputeHash(input);
+#if !NET462
+            using var hash = IncrementalHash.CreateHash(_hashAlgorithmName);
+            hash.AppendData(input);
+            return hash.GetHashAndReset();
+#else
+            using var hash = CryptoConfig.CreateFromName(_hashAlgorithmName.Name) as HashAlgorithm
```
Rob-Hague: `CryptoConfig` is a somewhat legacy API, and `CreateFromName` is annotated with `[RequiresUnreferencedCode]`. I'm not sure which of these points is the root cause for https://github.com/sshnet/SSH.NET/issues/859#issuecomment-1764623253 (needs more details), but either way we can just use `IncrementalHash` instead.

src/Renci.SshNet/Renci.SshNet.csproj
```diff
@@ -5,19 +5,23 @@
 	<TargetFrameworks>net462;netstandard2.0;netstandard2.1;net6.0;net7.0</TargetFrameworks>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" $([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0')) ">
+    <IsTrimmable>true</IsTrimmable>
+  </PropertyGroup>
```
WojciechNagorski: I don't think we need to enable this feature directly in the library, right?

src/Renci.SshNet/Renci.SshNet.csproj
```diff
@@ -5,19 +5,23 @@
...
+    
   <PropertyGroup Condition=" '$(TargetFramework)' == 'net462' ">
     <DefineConstants>FEATURE_BINARY_SERIALIZATION;FEATURE_SOCKET_EAP;FEATURE_SOCKET_APM;FEATURE_DNS_SYNC;FEATURE_HASH_RIPEMD160_CREATE;FEATURE_HMAC_RIPEMD160</DefineConstants>
   </PropertyGroup>
 
-  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' or '$(TargetFramework)' == 'netstandard2.1' or '$(TargetFramework)' == 'net6.0' or '$(TargetFramework)' == 'net7.0' ">
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' or $([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'netstandard2.1')) ">
```
WojciechNagorski, line 18: 👍 Good idea

src/Renci.SshNet/Renci.SshNet.csproj
Rob-Hague: According to https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming (if I have understood correctly) is it a library-level flag to declare that the library is compatible with trimming. It also enables the analysers.

So I think we either need `IsTrimmable` or `EnableTrimAnalyzer`, and because this PR fixes all the trimming warnings, then we can have `IsTrimmable`.

src/Renci.SshNet/Security/Cryptography/RsaDigitalSignature.cs
```diff
@@ -41,13 +41,20 @@ public RsaDigitalSignature(RsaKey rsaKey, HashAlgorithmName hashAlgorithmName)
         /// </returns>
         protected override byte[] Hash(byte[] input)
         {
-            return _hash.ComputeHash(input);
+#if !NET462
+            using var hash = IncrementalHash.CreateHash(_hashAlgorithmName);
+            hash.AppendData(input);
+            return hash.GetHashAndReset();
```
WojciechNagorski: Overall this PR looks good, but I need to understand this change before I could be able to merge this.

src/Renci.SshNet/Renci.SshNet.csproj
WojciechNagorski: This is good answer :)

src/Renci.SshNet/Security/Cryptography/RsaDigitalSignature.cs
Rob-Hague: Sure. `IncrementalHash` is a more modern and slightly lighter-weight method of hashing than `HashAlgorithm`. And it doesn't have trimming problems like `CryptoConfig.CreateFromName` does.

drieseng: Use `GetCurrentHash()` instead, unless we re-use the hash. See earlier remark.

src/Renci.SshNet/Security/Cryptography/RsaDigitalSignature.cs
```diff
@@ -41,13 +41,20 @@ public RsaDigitalSignature(RsaKey rsaKey, HashAlgorithmName hashAlgorithmName)
         /// </returns>
         protected override byte[] Hash(byte[] input)
         {
-            return _hash.ComputeHash(input);
+#if !NET462
+            using var hash = IncrementalHash.CreateHash(_hashAlgorithmName);
```
drieseng: Would we benefit (performance wise) from creating the hash in the ctor and reusing it here.
Please create a small benchmark to help us decide.

src/Renci.SshNet/Renci.SshNet.csproj
drieseng: @Rob-Hague, do you happen to know how exactly it "marks the assembly as "trimmable"? Do we also mark **SshNet.Security.Cryptography** as trimmable? I would've expect a warning because we have a dependency that is marked trimmable.

drieseng (changes requested):

src/Renci.SshNet/Security/Cryptography/RsaDigitalSignature.cs
Rob-Hague: If I read the source correctly, it actually seems like `GetCurrentHash` does more work (by duplicating the underlying hash instance)
https://github.com/dotnet/runtime/blob/main/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/HashProviderCng.cs

src/Renci.SshNet/Renci.SshNet.csproj
Rob-Hague: Good point, no, I don't know how it works with references. I have changed it just to `EnableTrimAnalyzer` until someone more knowledgeable comes along.

src/Renci.SshNet/Security/Cryptography/RsaDigitalSignature.cs
Rob-Hague: In addition to my other comment: this is a correctness fix, not a performance improvement. I would rather work on other things than to run a benchmark for such a change. Sorry for being blunt.

WojciechNagorski:
I will review this next week.

src/Renci.SshNet/Security/Cryptography/RsaDigitalSignature.cs
WojciechNagorski: So maybe let's move the hash creating to the constructor. What do you think? this is the last change and I will be able to merge it.

WojciechNagorski (changes requested):

src/Renci.SshNet/Security/Cryptography/RsaDigitalSignature.cs
Rob-Hague: Makes sense. Done

WojciechNagorski (approved):
Now LGTM!

WojciechNagorski:

PR #1232, opened 2023-11-02
Increase test coverage for AesCipher

Rob-Hague:
The tests were generated by a script which is also added for posterity. The script works by running `openssl enc [...]` (via WSL) to generate the expected encrypted values, and also verifies those values against the .NET BCL implementation as an extra validation before generating the tests.

The new tests uncovered a bug in `OfbCipherMode`, or rather, that it was an exact duplicate of `CfbCipherMode`. The second commit fixes the new failing tests (by following the nice pictures in https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation#Output_feedback_(OFB)).

I would like to help with #865, this should add confidence there.


zybexXL:
@Rob-Hague 
Thanks for this :) I've tested my code on #865 with this and all tests pass under all supported frameworks.
That PR failed one unrelated test, seems like a spurious fail.

WojciechNagorski (approved):
I've tested this PR.
LGTM

zybexXL:
image
Looks like Github doesn't handle this DST change very well :)

WojciechNagorski:

PR #1235, opened 2023-11-05
Use System.Security.Cryptography in AesCipher

Rob-Hague:
`System.Security.Cryptography.Aes` is supported on [all target platforms](https://learn.microsoft.com/en-us/dotnet/standard/security/cross-platform-cryptography#symmetric-encryption) (it wasn't in previous versions of the library). So we can use it rather than the hand-written code.

The change comes with modest performance gains (https://github.com/sshnet/SSH.NET/pull/865#issuecomment-1793748151) and from a maintenance perspective, makes it easier to achieve the design proposed in the linked comment in order to get much greater performance gains.


test/Renci.SshNet.IntegrationTests/SshTests.cs
```diff
@@ -63,7 +63,7 @@ public void Ssh_ShellStream_Exit()
                     catch (ObjectDisposedException ex)
                     {
                         Assert.IsNull(ex.InnerException);
-                        Assert.AreEqual("ShellStream", ex.ObjectName);
+                        Assert.AreEqual("Renci.SshNet.ShellStream", ex.ObjectName);
```
zybexXL, line 66: is this unrelated?

Rob-Hague: Yes, it was introduced in one of the recent analyzers fixups

WojciechNagorski:
I'm really impressed this is a great job!
Can you describe how this PRis related with #865 ?
Can you provide some Benchmark results?

Rob-Hague:
I would expect this to become the implementation of `SshNetCipherModeImpl` in the proposal in https://github.com/sshnet/SSH.NET/pull/865#issuecomment-1793748151, which would be used when using the constructor taking `Renci.SshNet.Security.Ciphers.CipherMode`. #865 would provide a much faster implementation when using a new constructor taking a `System.Security.Cryptography.CipherMode` value.

Reposting the benchmarks here (they are the same as in the linked comment)

Results on develop branch 826222f6fe33450754b1bb:

| Method      | Mean     | Error   | StdDev  | Gen0    | Allocated |
|------------ |---------:|--------:|--------:|--------:|----------:|
| Encrypt_CBC | 320.4 μs | 0.47 μs | 0.36 μs | 15.6250 |  32.41 KB |
| Decrypt_CBC | 352.3 μs | 2.37 μs | 1.85 μs | 15.6250 |  32.41 KB |
| Encrypt_CFB | 335.6 μs | 0.43 μs | 0.34 μs | 15.6250 |  32.45 KB |
| Decrypt_CFB | 367.9 μs | 4.23 μs | 3.95 μs | 15.6250 |  32.45 KB |
| Encrypt_CTR | 310.4 μs | 0.38 μs | 0.31 μs | 15.6250 |  32.45 KB |
| Decrypt_CTR | 315.3 μs | 3.96 μs | 3.70 μs | 15.6250 |  32.45 KB |


Results on these changes:

| Method      | Mean     | Error   | StdDev  | Gen0    | Allocated |
|------------ |---------:|--------:|--------:|--------:|----------:|
| Encrypt_CBC | 221.2 μs | 2.01 μs | 1.88 μs | 15.8691 |   32.7 KB |
| Decrypt_CBC | 220.0 μs | 2.73 μs | 2.56 μs | 15.8691 |  32.71 KB |
| Encrypt_CFB | 251.3 μs | 0.33 μs | 0.27 μs | 15.6250 |  32.75 KB |
| Decrypt_CFB | 255.3 μs | 0.67 μs | 0.60 μs | 15.6250 |  32.75 KB |
| Encrypt_CTR | 201.3 μs | 0.41 μs | 0.38 μs | 15.8691 |  32.75 KB |
| Decrypt_CTR | 209.3 μs | 0.22 μs | 0.19 μs | 15.8691 |  32.75 KB |


Results on #865 https://github.com/zybexXL/SSH.NET/commit/a9f68fbed70b94728869202c44f69869f2eb7693:

| Method      | Mean       | Error     | StdDev    | Gen0    | Allocated |
|------------ |-----------:|----------:|----------:|--------:|----------:|
| Encrypt_CBC |  29.637 μs | 0.1278 μs | 0.1067 μs | 16.1133 |  33.14 KB |
| Decrypt_CBC |   6.579 μs | 0.0160 μs | 0.0142 μs | 16.1285 |  33.14 KB |
| Encrypt_CFB | 333.761 μs | 0.1710 μs | 0.1335 μs | 15.6250 |  32.52 KB |
| Decrypt_CFB | 333.899 μs | 0.2090 μs | 0.1955 μs | 15.6250 |  32.52 KB |
| Encrypt_CTR |  24.630 μs | 0.0327 μs | 0.0273 μs | 78.1250 |  161.2 KB |
| Decrypt_CTR |  24.544 μs | 0.0817 μs | 0.0683 μs | 78.1250 |  161.2 KB |

The allocations in CTR mode can easily be removed, and I would expect it to extend to CFB as well. So given how much faster it is, it will be worth having as an implementation. The proposal is to keep the implementations separate so that we keep using a `Renci.SshNet.Security.Ciphers.CipherMode` if it is provided.

zybexXL:
@WojciechNagorski I've restructured #865 to incorporate suggestions from Rob-Hague in the above linked comment. This makes both PRs functionally similar, but <span>#</span>865 is much faster as this one only accelerates blockwise encryption/decryption, while 865 also accelerates one-shot buffer operations.

<span>#</span>865 already used acceleration (2 years ago) but the code was injected into the CipherMode classes instead of AesCipher.cs. I've now refactored and improved it.

WojciechNagorski:
@Rob-Hague, @zybexXL Which of these two PRs should be checked first? #1235 #865 They are related and I don't really know where to start.

src/Renci.SshNet/Security/Cryptography/Ciphers/AesCipher.cs
```diff
@@ -1,559 +1,16 @@
...
-using Renci.SshNet.Common;
+using System.Security.Cryptography;
 
 namespace Renci.SshNet.Security.Cryptography.Ciphers
 {
     /// <summary>
     /// AES cipher implementation.
     /// </summary>
-    public sealed class AesCipher : BlockCipher
+    public sealed class AesCipher : BlockCipher, IDisposable
```
WojciechNagorski, line 9: This `IDisposable` is problematic. `AesCipher` is created in some places, like here:

https://github.com/sshnet/SSH.NET/blob/2eec748b184e6ef3dfec93649c1ca1bdc27a627a/src/Renci.SshNet/ConnectionInfo.cs#L363C43-L378

but there is no Dispose call anywhere, so in the current state of this PR I cannot merge it.

WojciechNagorski (changes requested):

src/Renci.SshNet/Security/Cryptography/Ciphers/AesCipher.cs
Rob-Hague: All of the `Key`, `DigitalSignature`, `HashAlgorithm` classes are `IDisposable` but we do not dispose them in `ConnectionInfo`.

I don't think this introduces any new problems. I can remove the `IDisposable` if you prefer. Either way the `IDisposable` members would not be disposed.

WojciechNagorski: They should be disposed in `Session`. The `ConnectionInfo` has only a function that create `Key`, `DigitalSignature`, `HashAlgorithm`

But it is not true that they are not used:
https://github.com/sshnet/SSH.NET/blob/2eec748b184e6ef3dfec93649c1ca1bdc27a627a/src/Renci.SshNet/Session.cs#L1427-L1438

I don't want to say that the current implementation is perfect, but we just introduced a new problem in this PR. I didn't check how important it is.

Rob-Hague: You're right, I will add a check for `IDisposable` in `Session`

WojciechNagorski:
I like this kind of PR 
image

WojciechNagorski (approved):

WojciechNagorski:

PR #1236, opened 2023-11-07
For now, by default disable analyzers during build.

drieseng:
For now, by default disable analyzers during build.
Document how to locally enable the analyzers.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -21,6 +21,17 @@
         <EnableNETAnalyzers>true</EnableNETAnalyzers>
         <AnalysisLevel>preview-All</AnalysisLevel>
         <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+        <!--
+            To locally enable analyzers during build, create a 'Directory.Build.props.user'
+            file with the following content next to the current file:
+
+            <Project>
+                <PropertyGroup>
+                    <RunAnalyzersDuringBuild>true</RunAnalyzersDuringBuild>
+                </PropertyGroup>
+            </Project>
+        -->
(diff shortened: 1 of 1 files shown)


drieseng:
@WojciechNagorski, please review.

WojciechNagorski:
To be honest I don't agree with this change, I think we should run analyzers during build in CD process.

zybexXL:
Why not disable in DEBUG only ? It is somewhat annoying when coding.

WojciechNagorski:
Yes I agree that this is annoying. 
1. If we disable only for debug then when the contributor makes the new PR the build fails. 
2. If we disable analyzers at all - we will merge PRs that do not meet our rules.

Only a developer who makes a lot of changes and often should turn off the analyzers locally. This is only one setting and for an advanced developer it is not a problem. Junior may not be aware of how to turn it off, but if he changes a few lines, he learns more thanks to the analyzers. And there is a greater chance that PR will be built in the pipeline.

Rob-Hague:
I just don't know if we need 4 sets of analyzers (.NET, StyleCop, Meziantou, Sonar) in order to have a coherent code style.

Following this [guide](https://www.meziantou.net/understanding-the-impact-of-roslyn-analyzers-on-the-build-time.htm) a clean build of the solution takes 1min30 on my laptop, with 7min20 of CPU time burnt in Sonar. Both of these numbers are far too high.

image

image

Rob-Hague:
I think ideally we would not enable all analyzers by default with opt-out, instead disabling all by default and opting-in to those that are useful. But practically it would be a good start by disabling those which are not useful and take the most time.

BTCRob1993 (approved):

Rob-Hague:
I think the dust has settled on this now.

PR #1240, opened 2023-11-15
Update global.json

Rob-Hague:


WojciechNagorski (approved):
Today I've done the same changes in my project 😄

PR #1248, opened 2023-11-20
Fix build break in Release

Rob-Hague:


WojciechNagorski (approved):

WojciechNagorski:

PR #1255, opened 2023-12-01
Add support for .NET 8.0

scott-xu:


scott-xu:
Was going to add source link to SSH.NET and just found that it is [included in the .NET 8 SDK and enabled by default]( https://github.com/dotnet/sourcelink#using-source-link-in-net-projects).
This PR is only to add support for .NET 8.0. I can create another PR afterwards, to generate the [symbol package](https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg).
Would you please kindly have a review? Thanks! @WojciechNagorski @drieseng

appveyor.yml
```diff
@@ -12,20 +12,20 @@ for:
       - image: Ubuntu2204
 
   install:
-    - sh: sudo apt-get update && sudo apt-get install -y dotnet-sdk-7.0=7.0.403-1
+    - sh: sudo apt-get update && sudo apt-get install -y dotnet-sdk-8.0
```
WojciechNagorski: I would like to use the explicit version `dotnet-sdk-8.0=8.0.100` (or `8.0.100-1` you have to check it). When Microsoft releases a new version our tests will no longer work.

WojciechNagorski (changes requested):

appveyor.yml
scott-xu: From reading the console output, I know it is `8.0.100-1`.

Rob-Hague: The tests will work fine with a later version. Having an explicit version is just extra maintenance

test/Renci.SshNet.Tests/Classes/BaseClientTest_Connected_KeepAliveInterval_NotNegativeOne.cs
```diff
@@ -56,12 +56,12 @@ protected override void Act()
...
             _client.KeepAliveInterval = _keepAliveInterval;
 
-            // allow keep-alive to be sent a few times. .NET 7 is faster and
+            // allow keep-alive to be sent a few times. .NET 8 is faster and
             // we need to wait less because we want exactly three messages in a session.
 #if NETFRAMEWORK
             Thread.Sleep(195);
 #else
-            Thread.Sleep(180);
+            Thread.Sleep(170);
```
Rob-Hague, line 64: We should really get rid of timing in tests...

scott-xu: Agree. Better to create another PR...

appveyor.yml
scott-xu: If the unit test won't work for newer SDK, it means either our unit test can improve or it is a real problem found.

WojciechNagorski (approved):
LGTM! 
@drieseng @Rob-Hague do you mind?

Rob-Hague:
It's fine for me. I'm not convinced we need a net8.0 target without using any .NET 8 features (other than throw helpers), but using the .NET 8 SDK for source-link is good 👍

WojciechNagorski:

PR #1256, opened 2023-12-02
Merge nuspec and CommonAssemblyInfo into csproj

scott-xu:
Suggest using [Nerdbank.GitVersioning](https://github.com/dotnet/Nerdbank.GitVersioning) after this PR


src/Renci.SshNet/Renci.SshNet.csproj
```diff
@@ -1,23 +1,53 @@
...
-    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <AssemblyName>Renci.SshNet</AssemblyName>
-	<TargetFrameworks>net462;netstandard2.0;netstandard2.1;net6.0;net7.0;net8.0</TargetFrameworks>
+    <Product>SSH.NET</Product>
+    <AssemblyTitle>SSH.NET</AssemblyTitle>
+    <TargetFrameworks>net462;netstandard2.0;netstandard2.1;net6.0;net7.0;net8.0</TargetFrameworks>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
```
Rob-Hague: This should not be in the csproj (except for testing). We should use `dotnet pack` when we need to generate a package

```suggestion
```

Rob-Hague:
Will you also delete https://github.com/sshnet/SSH.NET/blob/develop/build/nuget/SSH.NET.nuspec?

(assuming this works well with sandcastle etc.)

scott-xu:
> Will you also delete https://github.com/sshnet/SSH.NET/blob/develop/build/nuget/SSH.NET.nuspec?
> 
> (assuming this works well with sandcastle etc.)

I didn't delete the nuspec because I need confirmation from @WojciechNagorski.

scott-xu:
Regarding sandcastle, it would be good to migrate to [docfx](https://dotnet.github.io/docfx/index.html)

src/Renci.SshNet/Renci.SshNet.csproj
```diff
@@ -1,23 +1,52 @@
...
+    <AssemblyTitle>SSH.NET</AssemblyTitle>
+    <TargetFrameworks>net462;netstandard2.0;netstandard2.1;net6.0;net7.0;net8.0</TargetFrameworks>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <PackageId>SSH.NET</PackageId>
+    <Title>SSH.NET</Title>
+    <Version>2023.0.1</Version>
+    <Description>SSH.NET is a Secure Shell (SSH) library for .NET, optimized for parallelism.</Description>
+    <Copyright>Copyright © Renci 2010-2023</Copyright>
```
drieseng: Can we replace **2023** with a dynamic value?

scott-xu: This PR only focus on the merging. Versioning could be another PR.

Rob-Hague:
```suggestion
    <Copyright>Copyright © Renci 2010-$([System.DateTime]::UtcNow.Year)</Copyright>
```

scott-xu:
> > Will you also delete https://github.com/sshnet/SSH.NET/blob/develop/build/nuget/SSH.NET.nuspec?
> > (assuming this works well with sandcastle etc.)
> 
> I didn't delete the nuspec because I need confirmation from @WojciechNagorski.

@WojciechNagorski @drieseng , what do you think? Should we delete SSH.NET.nuspec?

WojciechNagorski:
@scott-xu I'm sorry that the review is delayed, but I need more time to think and review it thoroughly.

Rob-Hague:
It looks good. Before (using build/build.proj per the [wiki](https://github.com/sshnet/SSH.NET/wiki/Build-a-new-release)):

<details>

image

</details>

After (using `dotnet pack`):

<details>

image

</details>

I think the next steps are:

1. Delete the nuspec
2. Modify build.proj to use `dotnet pack` instead of nuget.exe

Would it be possible to do these in this PR? That way we don't end up with two different ways to generate the release artifacts, and the new nuget method should integrate seamlessly into the current release process (which will also help with validation).

After that (in another PR) maybe we can call build.proj in CI to generate the artifacts.

WojciechNagorski:
I will try to find time next week for this PR

WojciechNagorski:
Hi I'm busy but I still remember about this PR.

Rob-Hague (approved):
Nice. Now we have source-link 🙂 

`msbuild /p:ReleaseVersion=2023.0.1 /t:Package`:

image

WojciechNagorski (approved):

WojciechNagorski:

PR #1257, opened 2023-12-02
Integrate with codecov

scott-xu:


scott-xu:
Need [codecov_token](https://docs.codecov.com/docs/codecov-uploader#per-repository-token), can you please help? @WojciechNagorski

PR #1258, opened 2023-12-02
Build with Release configuration

scott-xu:


WojciechNagorski:
We want to run tests in debug mode.

scott-xu:
Just want to see if it reduces the CI time.
The other thing is, if we want to use CI to generate the nuget package, it has to be Release mode.
Or we can build twice: one for unit test (debug), the other for nuget package (release).

WojciechNagorski:
We need to optimize our tests. For example many tests upload 5mb file, many multiple times. In unit tests we have many thread sleep.

WojciechNagorski:
We want to run our test in the debug mode because we want to have more logs

scott-xu:
Since the unit tests are much more stable now, can we switch to release mode? I ask because we can leverage Appvoyer to build, test and finally [deploy the package to nuget.org](https://www.appveyor.com/docs/deployment/nuget/). @WojciechNagorski @Rob-Hague

Rob-Hague:
I think we should still test in Debug mode, in order to cover various `Debug.Assert` calls. I don't think that stops us from creating a release package in CI(?)

WojciechNagorski:
I agree with Rob. If we want to release test nugget we have to build two times.

PR #1262, opened 2023-12-07
Always use factories for HmacAlgorithms

Rob-Hague:
Add a wrapping delegate on all the hmac factories in `ConnectionInfo`. This removes the need to load SshNet.Security.Cryptography.dll unless HMACRIPEMD160 is actually used (helpful when referencing the project directly in a test app).

Also re-order the algorithms. Note: a `Dictionary` does not guarantee order, and we should change it, but it does tend to enumerate in order as an implementation detail when items have not been removed).


WojciechNagorski (approved):
LGTM!

WojciechNagorski:

PR #1263, opened 2023-12-07
Enable diagnostics in release

Rob-Hague:
This should make troubleshooting apps easier


.editorconfig
```diff
@@ -571,6 +571,17 @@ MA0053.exceptions_should_be_sealed = true
 # Also report diagnostic for types that define (new) virtual members.
 MA0053.class_with_virtual_member_shoud_be_sealed = true
 
+# MA0076: Do not use implicit culture-sensitive ToString in interpolated strings
+#
+# Interpolated strings have performance benefits.
+dotnet_diagnostic.MA0076.severity = none
```
drieseng: Please revert this. This diagnostic does not question use of interpolated string.
It merely points out that - by default - interpolated strings are culture-sensitive.
For more information, see [Culture-specific formatting](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated#culture-specific-formatting). 

I asked MS to consider a prefix to fix an interpolated string to use invariant culture (eg. $i"....."), but alas.

Conclusion: it's a pain unless we drop support for anything < .NET 6.0.

src/Renci.SshNet/Diagnostics.cs
```diff
@@ -1,28 +1,24 @@
...
-        /// Currently, the library only traces events when compiled in Debug mode.
-        /// </para>
-        /// <para>
         /// Configuration on .NET Core must be done programmatically, e.g.
         /// <code>
-        /// DiagnosticAbstraction.Source.Switch = new SourceSwitch("sourceSwitch", "Verbose");
-        /// DiagnosticAbstraction.Source.Listeners.Remove("Default");
-        /// DiagnosticAbstraction.Source.Listeners.Add(new ConsoleTraceListener());
-        /// DiagnosticAbstraction.Source.Listeners.Add(new TextWriterTraceListener("trace.log"));
+        /// Diagnostics.Source.Switch = new SourceSwitch("sourceSwitch", nameof(SourceLevels.Verbose));
```
drieseng: We should update [Submitting an issue](https://github.com/sshnet/SSH.NET/wiki/Submitting-an-issue) on our wiki, and have a template for GitHub issues that asks users to provide diagnostic logging where applicable.

.editorconfig
```diff
@@ -571,6 +571,17 @@ MA0053.exceptions_should_be_sealed = true
...
+
+# MA0089: Optimize string method usage
+#
+# Duplicate of CA1865-CA1867 /
```
drieseng: Can you give me a few examples?

.editorconfig
```diff
@@ -571,6 +571,17 @@ MA0053.exceptions_should_be_sealed = true
...
+# For string.Join, results in verbose code for very little gain.
```
drieseng: Same.

src/Renci.SshNet/Security/KeyExchange.cs
```diff
@@ -183,10 +254,6 @@ public Cipher CreateServerCipher()
 
             serverKey = GenerateSessionKey(SharedKey, ExchangeHash, serverKey, _serverCipherInfo.KeySize / 8);
 
-            DiagnosticAbstraction.Log(string.Format("[{0}] Creating {1} server cipher.",
```
drieseng, line 185: I'd at least like to discuss why you prefer to remove these. The logging you added in this class does not help if you have multiple key exchanges in progress. The previous logging used a session id prefix for this purpose.

I'd prefer to use that prefix whether it's accessible.

src/Renci.SshNet/Session.cs
```diff
@@ -1157,15 +1172,19 @@ private bool TrySendMessage(Message message)
...
-                DiagnosticAbstraction.Log(string.Format("Failure sending message '{0}' to server: '{1}' => {2}", message.GetType().Name, message, ex));
-                return false;
-            }
-            catch (SocketException ex)
-            {
-                DiagnosticAbstraction.Log(string.Format("Failure sending message '{0}' to server: '{1}' => {2}", message.GetType().Name, message, ex));
-                return false;
+                if (Diagnostics.IsEnabled(TraceEventType.Error))
+                {
+                    Diagnostics.Log($"Failure sending message '{message.GetType().Name}' to server: '{message}' => {ex}", TraceEventType.Error);
```
drieseng: SessionId prefix? May not always be initialized though.

src/Renci.SshNet/Security/KeyExchange.cs
```diff
@@ -208,10 +275,6 @@ public Cipher CreateClientCipher()
 
             clientKey = GenerateSessionKey(SharedKey, ExchangeHash, clientKey, _clientCipherInfo.KeySize / 8);
 
-            DiagnosticAbstraction.Log(string.Format("[{0}] Creating {1} client cipher.",
```
drieseng, line 210: I'd at least like to discuss why you prefer to remove these. The logging you added in this class does not help if you have multiple key exchanges in progress. The previous logging used a session id prefix for this purpose.

I'd prefer to use that prefix whether it's accessible.

src/Renci.SshNet/Security/KeyExchange.cs
```diff
@@ -232,10 +295,6 @@ public HashAlgorithm CreateServerHash()
                                                Hash(GenerateSessionKey(SharedKey, ExchangeHash, 'F', sessionId)),
                                                _serverHashInfo.KeySize / 8);
 
-            DiagnosticAbstraction.Log(string.Format("[{0}] Creating {1} server hmac algorithm.",
```
drieseng, line 234: I'd at least like to discuss why you prefer to remove these. The logging you added in this class does not help if you have multiple key exchanges in progress. The previous logging used a session id prefix for this purpose.

I'd prefer to use that prefix whether it's accessible.

src/Renci.SshNet/Security/KeyExchange.cs
```diff
@@ -255,10 +314,6 @@ public HashAlgorithm CreateClientHash()
                                                Hash(GenerateSessionKey(SharedKey, ExchangeHash, 'E', sessionId)),
                                                _clientHashInfo.KeySize / 8);
 
-            DiagnosticAbstraction.Log(string.Format("[{0}] Creating {1} client hmac algorithm.",
```
drieseng, line 257: I'd at least like to discuss why you prefer to remove these. The logging you added in this class does not help if you have multiple key exchanges in progress. The previous logging used a session id prefix for this purpose.

I'd prefer to use that prefix whether it's accessible.

src/Renci.SshNet/Security/KeyExchange.cs
```diff
@@ -275,10 +330,6 @@ public Compressor CreateCompressor()
                 return null;
             }
 
-            DiagnosticAbstraction.Log(string.Format("[{0}] Creating {1} client compressor.",
```
drieseng, line 277: I'd at least like to discuss why you prefer to remove these. The logging you added in this class does not help if you have multiple key exchanges in progress. The previous logging used a session id prefix for this purpose.

I'd prefer to use that prefix whether it's accessible.

src/Renci.SshNet/Security/KeyExchange.cs
```diff
@@ -299,10 +350,6 @@ public Compressor CreateDecompressor()
                 return null;
             }
 
-            DiagnosticAbstraction.Log(string.Format("[{0}] Creating {1} server decompressor.",
```
drieseng, line 301: I'd at least like to discuss why you prefer to remove these. The logging you added in this class does not help if you have multiple key exchanges in progress. The previous logging used a session id prefix for this purpose.

I'd prefer to use that prefix whether it's accessible.

drieseng (changes requested):
What rationele did you apply for deciding when to "protect" logging with a corresponding `IsEnabled(...)` check?

I see you protecting (cheap) warning-level logging, but not verbose-level logging. While you could expect more people to have warning and error level logging enabled by default, but almost no one enables verbose logging unless you're troubleshooting an issue.

Can you perform (and commit) a (a simple connect) benchmark that shows the impact on performance with all tracing disabled (with any without this PR applied)?

I don't think we have benchmarks where we make use of our docker infrastructure.
Would be nice to have this. @WojciechNagorski, am I right?

That said, I'd love to get this PR in! Thanks!

.editorconfig
Rob-Hague: Right, but what's wrong with using current culture here? Doesn't that make the most sense?

src/Renci.SshNet/Diagnostics.cs
Rob-Hague: I'll update the wiki when this goes in (also https://github.com/sshnet/SSH.NET/wiki/Troubleshooting-SSH.NET#net-tracesource-and-tracelisteners)

.editorconfig
Rob-Hague: I think these are new analyzers in net8.0 sdk, but I couldn't get them to fire (maybe they are aware of the other tfms which don't have the replacement methods?).

I could only get CA1847

image

and some other similar looking warnings

image

I can revert, pending discussion on string.Join

.editorconfig
Rob-Hague: To comply with this, every "offer" message in `KeyExchange` needs to be duplicated or otherwise #if-def-ed to use `string.Join(',', ..` or `string.Join(",", ..`. But the methods perform the same:

```c#
        private static readonly string[] s_ciphers = ["aes128-ctr", "aes192-ctr", "aes256-ctr", "aes128-cbc", "aes192-cbc", "aes256-cbc", "3des-cbc", "blowfish-cbc", "twofish-cbc", "twofish192-cbc", "twofish128-cbc", "twofish256-cbc", "arcfour", "arcfour128", "arcfour256", "cast128-cbc"];

        [Benchmark]
        public string JoinString() => string.Join(",", s_ciphers);

        [Benchmark]
        public string JoinChar() => string.Join(',', s_ciphers);
```

| Method     | Mean     | Error    | StdDev   | Gen0   | Allocated |
|----------- |---------:|---------:|---------:|-------:|----------:|
| JoinString | 79.95 ns | 0.686 ns | 0.642 ns | 0.1912 |     400 B |
| JoinChar   | 80.61 ns | 0.901 ns | 0.843 ns | 0.1912 |     400 B |

src/Renci.SshNet/Security/KeyExchange.cs
Rob-Hague: I wanted the "offer" messages and figured I could put these ones in the same place. I'll put them back

Rob-Hague:
> What rationele did you apply for deciding when to "protect" logging with a corresponding `IsEnabled(...)` check?

If the message is a literal string, I did not guard it.

> Can you perform (and commit) a (a simple connect) benchmark that shows the impact on performance with all tracing disabled (with any without this PR applied)?
> 
> I don't think we have benchmarks where we make use of our docker infrastructure.

I don't think a benchmark would be suited for such an end-to-end networking scenario (too much noise). The module shows as ~1% of CPU of a basic connect->listdirectory->downloadfiles scenario at verbose level, and does not show in a memory trace.

image

Where we could use the benchmarks project is to define a number of scenarios from which it is easy to generate profiles for identifying where we should place the focus on performance. I found a [nice blog](https://wojciechnagorski.com/2020/04/cross-platform-profiling-.net-code-with-benchmarkdotnet/) which could be used as a reference 😉

src/Renci.SshNet/Channels/Channel.cs
```diff
@@ -554,9 +554,9 @@ protected virtual void Close()
                         // SSH_MSG_CHANNEL_CLOSE as response to a SSH_MSG_CHANNEL_CLOSE sent by the
                         // server
                         var closeWaitResult = _session.TryWait(_channelClosedWaitHandle, ConnectionInfo.ChannelCloseTimeout);
-                        if (closeWaitResult != WaitResult.Success)
+                        if (closeWaitResult != WaitResult.Success && Diagnostics.IsEnabled(TraceEventType.Warning))
```
WojciechNagorski: I don't get why we need the condition `Diagnostics.IsEnabled(TraceEventType.Warning)` outside the log method. 

what if someone forgot to check it?

WojciechNagorski (changes requested):

src/Renci.SshNet/Channels/Channel.cs
Rob-Hague: We don't need it, it just prevents doing any extra work like allocating string instances unnecessarily in the case that the log level is not enabled. But TBH I don't think it would make a material impact either way.

Rob-Hague:
Superseded by #1509

PR #1265, opened 2023-12-08
Replace SemaphoreLight with SemaphoreSlim

Rob-Hague:
This paves the way for asynchronous synchronisation via WaitAsync (and eliminates a timing test which sometimes fails in CI)


WojciechNagorski:
How do you know there is a problem with SemaphoreLight?

WojciechNagorski (approved):
LGTM! It is stable and I think it's better to use official semaphore.

Rob-Hague:
> How do you know there is a problem with SemaphoreLight?

I looked through the recent builds and saw this one which had failed https://ci.appveyor.com/project/drieseng/ssh-net/builds/48640576/tests. I don't think it will fix any other tests

WojciechNagorski:

PR #1266, opened 2023-12-09
Standardizes additional file types in .editorconfig

jacobslusser:
This is my first PR for this project and I didn't find any contributing guidelines, so please let me know if there is anything else I should do for submitting PRs. 

- Fixes the path to `TestResults/` in the `.gitignore`. When the `.sln` file was moved 2 months ago, this started putting `TestResults/` folder in the root instead of `src/`.
- Adds formatting rules (2 space tabs) for `.xml`, `.csproj`, and `.props` files to `.editorconfig`. I also formatted those files according to these rules.
- Adds `visible=false` to the `stylecop.json` reference in `Directory.Build.props`. Previously the `stylecop.json` file would show in every project, but now will only show in the Solution Items.

These are just simple things I found as I started getting acquainted with the code and running tests. I have bigger and better submissions planned but wanted to test the waters a little with the group and understand the process.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,6 +9,11 @@ indent_size = 4
 tab_width = 4
 end_of_line = crlf
 
+[*.{xml,csproj,props}]
+
+indent_size = 2
+tab_width = 2
+
 [*.cs]
 
 #### Sonar rules ####
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@ test/**/bin/**
 test/**/obj/**
 
 # MSTest test Results
-src/TestResults/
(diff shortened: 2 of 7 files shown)


WojciechNagorski (approved):
Amazing contribution! Thanks.
Welcome on board and we look forward to more!

Rob-Hague:
Thanks @jacobslusser. Please do let it be known what your plans are - the library is gaining a bit of momentum at the moment, and we probably ought to sync up a bit or start assigning ourselves to issues etc.

WojciechNagorski:

PR #1138, opened 2023-06-04
Add a few spans on net6.0 or greater

Rob-Hague:

PR #1272, opened 2023-12-11
Reduces heap allocations for the some byte[] uses

jacobslusser:
Profiling reveals that creation of `byte[4]` arrays are our most common heap objects. Many of these come from reading integers from the byte stream where we allocate a small `byte[4]` or `byte[8]` array and then convert it to a 32 or 64-bit integer. 

Results from a simple test of using the `SftpClient` to upload and download some files:

**Before: when byte[] arrays were created for each integer read**
Screenshot 2023-12-10 181955

**After: when byte[] are allocated and read from the stack**
Screenshot 2023-12-10 182136

**Changes**
- Creates a new `SshDataStream.ReadBytes` overload that supports `Span<byte>` for .NET Standard 2.1+ and .NET 6+ targets.
- Updates `SshDataStream.ReadUInt32` and `SshDataStream.ReadUInt64` to read onto the stack and use newer `BinaryPrimitives` class for integer conversion for .NET Standard 2.1+ and .NET 6+ targets.
- Updates `SshData.ReadUInt32` and `SshData.ReadUInt64` to use the existing stream methods to avoid another temporary buffer.

### DIFF
--- a/src/Renci.SshNet/Common/SshDataStream.cs
+++ b/src/Renci.SshNet/Common/SshDataStream.cs
@@ -188,8 +188,14 @@ namespace Renci.SshNet.Common
         /// </returns>
         public uint ReadUInt32()
         {
+#if NETSTANDARD2_1_OR_GREATER || NET6_0_OR_GREATER
+            Span<byte> span = stackalloc byte[4];
+            ReadBytes(span);
+            return System.Buffers.Binary.BinaryPrimitives.ReadUInt32BigEndian(span);
+#else
             var data = ReadBytes(4);
             return Pack.BigEndianToUInt32(data);
+#endif // NETSTANDARD2_1_OR_GREATER || NET6_0_OR_GREATER
         }
 
         /// <summary>
@@ -200,8 +206,14 @@ namespace Renci.SshNet.Common
         /// </returns>
         public ulong ReadUInt64()
         {
+#if NETSTANDARD2_1_OR_GREATER || NET6_0_OR_GREATER
+            Span<byte> span = stackalloc byte[8];
+            ReadBytes(span);
+            return System.Buffers.Binary.BinaryPrimitives.ReadUInt64BigEndian(span);
+#else
             var data = ReadBytes(8);
             return Pack.BigEndianToUInt64(data);
+#endif // NETSTANDARD2_1_OR_GREATER || NET6_0_OR_GREATER
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Rob-Hague:
Funny, I had very similar changes in #1138 🙂 Feel free to cherry-pick some of those, or otherwise this looks good to me

src/Renci.SshNet/Common/SshDataStream.cs
```diff
@@ -264,5 +276,21 @@ private byte[] ReadBytes(int length)
...
+        /// Reads data into the specified <paramref name="buffer" />.
+        /// </summary>
+        /// <param name="buffer">The buffer to read into.</param>
+        /// <exception cref="ArgumentOutOfRangeException"><paramref name="buffer"/> is larger than the total of bytes available.</exception>
+        private void ReadBytes(Span<byte> buffer)
+        {
+            var bytesRead = Read(buffer);
+            if (bytesRead < buffer.Length)
+            {
+                throw new ArgumentOutOfRangeException(nameof(buffer), string.Format(CultureInfo.InvariantCulture, "The requested length ({0}) is greater than the actual number of bytes read ({1}).", buffer.Length, bytesRead));
```
WojciechNagorski, line 291: Wouldn't it be better to throw exceptions from separate methods? I don't remember exactly, but it was probably better for performance reasons. Throwing exceptions probably blocked the method inlining.

This doesn't block this PR, it's more of a question.

Rob-Hague: The benefit of using "throw helpers" is that the code size of the method is reduced by shifting the "cold" throwing code into another method (which is not inlined).

See [this example](https://sharplab.io/#v2:EYLgtghglgdgNAFxBAzmAPgAQEwEYCwAUJgAwAEmuAdACJQQDmMA9iglAMYpUDCzAJgFMAgjAgAbAJ4ooKANxFSFagHFxzYBKgAvCO2YwFxcpSoBJAPJHMAZgrYyPMiDIBZQWGYAnSQGUEXoIQYGQA3kRkkRR2mAAsZABKQfwAQpIIgii4ABS+AA4QMAA8wOmCAHxkwACuAGa1gl4AlGERUe0AbhBeVWUoSRD8ZAC8icnZNfWNTUbt7VC1ZBN9A0NFVXUNXlQAMoIwDAgAFk1tc62E5+fHXswA7mQwgg/CXgzVYPsIFtXftQmFBiCACiAA8OII8voYNkxJ9mLUJptpnBlCQqAAxbyQBDZHjVcQIaqBMwwWrMcwwLpeegwBD4wnEwSogBEABUjoIyIEAI7VTIZIbifaHI5LUIkAC+LVkZAYgT0jTIx0Kys5ZAgHCJEkeH2ASoRvQyKG5yXFuGlVBZqMmW12IuONpWySaMzOc0l7s9lyi7tsFHiqzSxuwuQKxVKGUqtumFyuZGpRsyqxGY0GSKmzVmVwWS0jybN6xj2z2B2Opx953ClfjZA5tzuGa2TuNqzdNfa3vj7vaAG0aMxMgA5ZgIJJErwwAC6PailAAbAG60cG2HCiUytHkV5UbAEEn+i7Z5Fq7Wojd7o9nmRXu9PnSfn8AQcQeDIdDYcFBAimyi0ZjsT0PECQnQRSXJSlqVpekQKZVkOS5Xl+TYQQhQdMVsglaUyFleUggyHoVRgNUuU1bVxF1MB9R6Q18xNBUhkwi0mitG1t3tMsjhbAtBldbNzklbsa29SUgA). The generated code for `ReadBytes2` is much smaller than `ReadBytes1`.

In practice, I don't think we need to be thinking at this level of optimisation just yet

WojciechNagorski (approved):

jacobslusser:
> Funny, I had very similar changes in #1138 🙂 Feel free to cherry-pick some of those, or otherwise this looks good to me

@Rob-Hague wow, your changes are more than a little similar. At one point I had even ported the `ReadExactly` just as you had but decided instead to follow our existing pattern. I will definitely look at merging the others. 

Why didn't you ever merge this?

Rob-Hague:
I think I wanted to justify it with some profiling like you did, but got stuck into some other work

WojciechNagorski:
I will merge any optimization if it has justify as goof as this PR 😃. https://github.com/sshnet/SSH.NET/pull/1138 was a draft.

WojciechNagorski:

PR #1274, opened 2023-12-15
Send the client key exchange init in Connect

Rob-Hague:
After discussion in #972, I have been running this in production against several different servers for a week or so.

An example key exchange with this change:

image

And an example when the server init comes in first:

image

cc @geoffstewart

### DIFF
--- a/src/Renci.SshNet/Security/IKeyExchange.cs
+++ b/src/Renci.SshNet/Security/IKeyExchange.cs
@@ -38,8 +38,9 @@ namespace Renci.SshNet.Security
         /// Starts the key exchange algorithm.
         /// </summary>
         /// <param name="session">The session.</param>
-        /// <param name="message">Key exchange init message.</param>
-        void Start(Session session, KeyExchangeInitMessage message);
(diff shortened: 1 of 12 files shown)


WojciechNagorski:
I will check this next week.

Rob-Hague:
@BoronBGP @expyram perhaps you could test this on a cisco device? Bonus points for pasting the packet trace

WojciechNagorski:
Can we prepare tests to cover two cases? When does server initialization occur and when does it not?

Rob-Hague:
Added a unit test for when the server does not send kexinit

edit: 

> When does server initialization occur and when does it not?

The [RFC](https://datatracker.ietf.org/doc/html/rfc4253#section-7) says "Key exchange (kex) begins by each side sending name-lists of supported algorithms." I.e. there is no indication that the client or server should wait for the other.

SSH.NET currently waits for the server to send its supported algorithms. This works OK against most server implementations, but if the server is waiting for SSH.NET, then we have a problem.

WojciechNagorski (approved):
LGTM!

WojciechNagorski:

PR #1278, opened 2023-12-20
Try to stabilise a few tests

Rob-Hague:


WojciechNagorski:
It's ready for review?

Rob-Hague:
Yes I think it should capture most recent failures that we've seen

WojciechNagorski (approved):

WojciechNagorski:

PR #1279, opened 2023-12-22
Upgrade packages

Rob-Hague:
- Remove reference to [Microsoft.CodeAnalysis.NetAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.NetAnalyzers) (use the stable version by default)
- Upgrade [StyleCop.Analyzers](https://www.nuget.org/packages/StyleCop.Analyzers#versions-body-tab) to 1.2.0-beta.556 (fixes a collection expression whitespace issue - the motivation for this change)
- Upgrade [SonarAnalyzer.CSharp](https://www.nuget.org/packages/SonarAnalyzer.CSharp/#versions-body-tab) to 9.16.0.82469 (the release notes have some performance improvements 👍)
- Upgrade [Microsoft.Bcl.AsyncInterfaces](https://www.nuget.org/packages/Microsoft.Bcl.AsyncInterfaces) to 8.0.0 while we are here.

I did not upgrade [Meziantou.Analyzer](https://www.nuget.org/packages/Meziantou.Analyzer) because the release notes show some new analyzers so it probably needs more attention there.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -34,9 +34,8 @@
         Use fixed version of analyzers.
     -->
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="8.0.0-preview1.23165.1" PrivateAssets="all" />
-    <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.435" PrivateAssets="all" />
+    <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.556" PrivateAssets="all" />
     <PackageReference Include="Meziantou.Analyzer" Version="2.0.103" PrivateAssets="all" />
-    <PackageReference Include="SonarAnalyzer.CSharp" Version="9.12.0.78982" PrivateAssets="all" />
+    <PackageReference Include="SonarAnalyzer.CSharp" Version="9.16.0.82469" PrivateAssets="all" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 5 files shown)


Directory.Build.props
```diff
@@ -34,9 +34,8 @@
         Use fixed version of analyzers.
     -->
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="8.0.0-preview1.23165.1" PrivateAssets="all" />
```
WojciechNagorski, line 37: > Remove reference to Microsoft.CodeAnalysis.NetAnalyzers (use the stable version by default)

You should not remove this package because of .NET Framework do not have analyzers by default, should you?

WojciechNagorski (changes requested):

Directory.Build.props
Rob-Hague: According to https://www.nuget.org/packages/Microsoft.CodeAnalysis.NetAnalyzers, it is included by virtue of `<EnableNETAnalyzers>true</EnableNETAnalyzers>` on line 21

WojciechNagorski: But I think you still need to install Nuget. Can you check it?

Rob-Hague: The packages README says

> You do not need to manually install this NuGet package to your project if you are using .NET5 SDK or later

We are using .NET8 SDK. An example of a CA analyzer on net462 target:

image

WojciechNagorski (approved):

PR #1177, opened 2023-09-16
Add support for RSA SHA-2 public key algorithms

Rob-Hague:
See #1174 and [RFC 8332](https://datatracker.ietf.org/doc/html/rfc8332) for details.

closes #1174 
closes #1145 
closes #1116
closes #1105
closes #956
closes #825

It may also close the following:
#1141
#1025
#1003
#858
#845

PR #1283, opened 2023-12-28
Authenticate with ssh-rsa by default

### LINKED ISSUE #1233: SshConnectionException after update to 2023.0.0 with private key file
I'm getting a Renci.SshNet.Common.SshConnectionException: An established connection was aborted by the server. after upgrading to SSH.NET 2023.0.0 
my code still works fine with SSH.NET 2020.0.2

I'm using a private key file to connect (connecting with another server where I have just user/pw still works fine with 2023.0.0) 

I have tried this with .NET Framework 4.8 and .NET 7.0, neither works

could you please look into this?

Rob-Hague:
... to appease servers which disconnect rather than send SSH_MSG_USERAUTH_FAILURE when they do not support sha2 signatures for client authentication (and who may or may not have otherwise sent the server-sig-algs extension which we do not currently implement).

It will still use a sha2 signature if the server does not allow ssh-rsa and sends SSH_MSG_USERAUTH_FAILURE. I would think a server which disables ssh-rsa would be more friendly/compliant than those which disconnect rather than sending SSH_MSG_USERAUTH_FAILURE (i.e. the risk of the opposite now happening, where a server disconnects on receiving ssh-rsa, is low - it is going back to the behaviour as before #1177).

See https://github.com/sshnet/SSH.NET/issues/1233#issuecomment-1871196405 for further discussion

Fixes #1233

### DIFF
--- a/src/Renci.SshNet/PrivateKeyFile.cs
+++ b/src/Renci.SshNet/PrivateKeyFile.cs
@@ -250,11 +250,11 @@ namespace Renci.SshNet
                 case "RSA":
                     var rsaKey = new RsaKey(decryptedData);
                     _key = rsaKey;
+                    _hostAlgorithms.Add(new KeyHostAlgorithm("ssh-rsa", _key));
 #pragma warning disable CA2000 // Dispose objects before losing scope
                     _hostAlgorithms.Add(new KeyHostAlgorithm("rsa-sha2-512", _key, new RsaDigitalSignature(rsaKey, HashAlgorithmName.SHA512)));
                     _hostAlgorithms.Add(new KeyHostAlgorithm("rsa-sha2-256", _key, new RsaDigitalSignature(rsaKey, HashAlgorithmName.SHA256)));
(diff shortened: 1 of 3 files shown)


jkillingsworth:
I have created a local build that includes this commit and tested it against three different servers.

* :heavy_check_mark: `SSH-2.0-Axway.Gateway`
* :heavy_check_mark: `SSH-2.0-OpenSSH_for_Windows_9.4`
* :heavy_check_mark: `SSH-2.0-AWS_SFTP_1.1`

All three connect and authenticate successfully when using private key authentication. Only two out of the three were working before. This commit solves the problem I was having as described in https://github.com/sshnet/SSH.NET/issues/1233#issuecomment-1869896774.

WojciechNagorski (commented):
LGTM!

WojciechNagorski (approved):

Rob-Hague:
Thanks!

WojciechNagorski:

Issue #1287, opened 2024-01-03
2023.0.1 requires exact Microsoft.Bcl.AsyncInterfaces version


mus65:
I noticed that #1268 introduced a dependency to an exact version of Microsoft.Bcl.AsyncInterfaces (8.0.0) . I assume this was a copy&paste mistake from the SshNet.Security.Cryptography dependency. Please changes this to a minimum version (>= 8.0.0), otherwise projects using SSH.NET will run into issues when trying to upgrade to a newer Microsoft.Bcl.AsyncInterfaces release.

darkoperator:
seems I hit an issue with this in Posh-SSH

mus65:
@darkoperator doesn't look like the same issue. This issue is about the too strict version constraint which would cause nuget restore failures if other projects depend on another Microsoft.Bcl.AsyncInterfaces version. Your issue seems to be about the dependency itself missing.

darkoperator:
ok, sadly my module is broken because of this dependency on PS 7, it is one weird bug I can not figure out so sticking with the previous nugget for now

mus65:
Fixed by #1288

WojciechNagorski:
This issue has been fixed in the 2024.0.0 version.

ThatBrianK:
Hi, I just wanted to chime in and say that this fix actually broke the library I had built. I now get an error:

Could not load file or assembly 'Microsoft.Bcl.AsyncInterfaces, Version=1.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or 
one of its dependencies. The system cannot find the file specified.

I do not get this error with 2023.0.1 but with any higher version I do. I tried 2 tricks I could find online but neither of them worked.
1. check "auto-generate binding redirects
2. adding an App.config file containing the following:
`		<assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
			<dependentAssembly>
				<assemblyIdentity name="Microsoft.Bcl.AsyncInterfaces" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
				<bindingRedirect oldVersion="0.0.0.0-8.0.0.0" newVersion="8.0.0.0" />
			</dependentAssembly>
		</assemblyBinding>`

I'm not really sure where to go for help. Thanks!

mus65:
This does indeed look like missing binding redirects. There is no text on 2. , did you forget to paste it?

[...]

PR #1288, opened 2024-01-04
Require Microsoft.Bcl.AsyncInterfaces >=1.0.0 instead of =8.0.0

mus65:
This technically fixes 2 different issues:
    - previously SSH.NET would require exactly(!) version 8.0.0, which would
      cause issues in the future if e.g. a project using SSH.NET would
      try to update to version 9.0.0. In this case the constraint on the
      exact version would cause NuGet to refuse to upgrade to 9.0.0 (#1287)
    - PowerShell Core seems to have problems with Version 8.0.0, therefore
      downgrade the requirement to 1.0.0 (see https://github.com/darkoperator/Posh-SSH/issues/558)

I had to re-enable AutoGenerateBindingRedirects, otherwise there is a
FileLoadException with net462 in one of the tests. I don't know why this
was disabled in 3ecbd1071d653bbf04254c08d14d7ce75d60688b .

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -11,7 +11,6 @@
     <LangVersion>latest</LangVersion>
     <WarningLevel>9999</WarningLevel>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <AutoGenerateBindingRedirects>false</AutoGenerateBindingRedirects>
   </PropertyGroup>
 
   <!--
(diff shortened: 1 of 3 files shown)


WojciechNagorski:
@Rob-Hague can you make a review?

Rob-Hague (approved):
Thanks!

WojciechNagorski (approved):

WojciechNagorski:

PR #1290, opened 2024-01-06
Remove unneeded feature flags and pass cancellationToken to DNS request

mus65:
This removes feature flags that were unused. FEATURE_SOCKET_EAP and FEATURE_DNS_SYNC were set for all frameworks, so unneeded. The other feature flags were only used in elif of FEATURE_SOCKET_EAP and FEATURE_DNS_SYNC, so also unneeded.

After removing these, DnsAbstractions.cs seemed to be an unneeded abstraction at first, but  then I noticed that .NET 6+ supports a cancellationToken on GetHostAddressesAsync, so I kept DnsAbstractions.cs to pass this through  (I could do a separate PR for this if you want).


Rob-Hague (approved):

src/Renci.SshNet/Abstractions/DnsAbstraction.cs
```diff
@@ -36,72 +22,27 @@ public static IPAddress[] GetHostAddresses(string hostNameOrAddress)
...
         /// <param name="hostNameOrAddress">The host name or IP address to resolve.</param>
+        /// <param name="cancellationToken">A cancellation token that can be used to signal the asynchronous operation should be canceled.</param>
         /// <returns>
         /// A task with result of an array of type <see cref="IPAddress"/> that holds the IP addresses for the host that
         /// is specified by the <paramref name="hostNameOrAddress"/> parameter.
         /// </returns>
         /// <exception cref="ArgumentNullException"><paramref name="hostNameOrAddress"/> is <see langword="null"/>.</exception>
         /// <exception cref="SocketException">An error is encountered when resolving <paramref name="hostNameOrAddress"/>.</exception>
-        public static Task<IPAddress[]> GetHostAddressesAsync(string hostNameOrAddress)
+        public static Task<IPAddress[]> GetHostAddressesAsync(string hostNameOrAddress, CancellationToken cancellationToken)
```
WojciechNagorski: I have a problem with this line. The Cancellation token is used only for .NET 6 +

mus65: I don't see an alternative here, there is no API to make it cancellable on old frameworks.

Rob-Hague: Can we just delete the whole file? It is not providing any abstraction now that the rest of the code is gone. `GetHostAddressesAsync` is only used in one place so we can move the #ifdef on the cancellation token to the callsite

mus65: I thought about this too, but if/when the sync `GetHostAddresses()` calls are changed to async, you will have do ifdef on multiple places, so the abstraction may be useful after all. But I'm fine with either...

Rob-Hague: I don't think that TODO is worthwhile, and the library already makes liberal use of #ifdefs. Deleting the file sounds good to me

mus65: ok, I pushed the changes

WojciechNagorski (approved):

WojciechNagorski:

PR #1293, opened 2024-01-23
Added SshCommand.InputStream to allow writing to stdin of SshCommand

realvizu:
Changes:
- Added SshCommand.InputStream (type: ChannelInputStream), which sends the data written to the stream through a command's channel.
- Added an integration test that echos back the sent data using the "cat" command.
- All unit tests pass.

I you have recommendations for further unit tests, please advise.

---

I started the work by trying to run all unit tests on my dev machine to see if they all pass. Unfortunately a lot of them failed, so I had to make the following modifications (see the changes in my first commit in this PR):
- Excluded MD5 tests for the net462 target, because I got System.InvalidOperationException: 'This implementation is not part of the Windows Platform FIPS validated cryptographic algorithms.'
- Modified SshdConfig.cs: do not throw for "Include", just do nothing.
- Modified failing dos2unix parameters in Dockerfile.TestServer.
- Added a line to .gitattributes to force LF line ending for key files used by integration tests, otherwise using them causes error.
- Modified SftpClientTest.Test_Sftp_Multiple_Async_Upload_And_Download_10Files_5MB_Each to upload only 2 files instead of 10 because for more than 2 files I got timeout errors.

If you know a better fix for any of the above modifications, please advise.


test/Renci.SshNet.IntegrationTests/OldIntegrationTests/SftpClientTest.Upload.cs
```diff
@@ -78,7 +78,9 @@ public void Test_Sftp_Upload_Forbidden()
         [TestCategory("Sftp")]
         public void Test_Sftp_Multiple_Async_Upload_And_Download_10Files_5MB_Each()
         {
-            var maxFiles = 10;
+            // Works for up to 2 files, but fails for more files (on my machine).
+            // I get either "Renci.SshNet.Common.SshException: Channel was closed" or timeout exception.
+            var maxFiles = 2;
```
WojciechNagorski: I'll have to check this on my machine because it seems to be working now. But I know that this test is unstable. so don't worry about it.

test/Renci.SshNet.Tests/Classes/Common/HostKeyEventArgsTest.cs
```diff
@@ -45,6 +45,10 @@ public void HostKeyEventArgsConstructorTest()
             Assert.AreEqual(2048, target.KeyLength);
         }
 
+// Excluding on net462 platform, because using MD5 hash throws: System.InvalidOperationException:
+// 'This implementation is not part of the Windows Platform FIPS validated cryptographic algorithms.'
+#if !NET462_TARGET_FRAMEWORK
```
WojciechNagorski: I'll have to check this on my machine because it seems to be working now.

test/Renci.SshNet.Tests/Classes/PrivateKeyFileTest.cs
```diff
@@ -1,4 +1,8 @@
-﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+﻿// Excluding these tests on net462 platform, beacuse using MD5 hash throws: System.InvalidOperationException:
+// 'This implementation is not part of the Windows Platform FIPS validated cryptographic algorithms.'
+#if !NET462_TARGET_FRAMEWORK
```
WojciechNagorski: I'll have to check this on my machine because it seems to be working now.

test/Renci.SshNet.IntegrationTests/Dockerfile.TestServer
```diff
@@ -10,7 +10,8 @@ RUN apk update && apk upgrade --no-cache && \
     apk add --no-cache openssh && \
     # install openssh-server-pam to allow for keyboard-interactive authentication
     apk add --no-cache openssh-server-pam && \
-    dos2unix /etc/ssh/* && \
+    # must not use * for dos2unix parameter otherwise it tries to process folders too and fails
+    dos2unix /etc/ssh/ssh*key && \
```
WojciechNagorski, line 14: Maybe there is a new alpine image and it has different behavior, but I don't have a problem with this change as long as it works.

WojciechNagorski:
It looks good to me! But I need more time to test it!

Do you have a specific function in mind that you want to implement using this functionality? 
Did you see that it can be useful in https://github.com/dotnet/Docker.DotNet/issues/540

WojciechNagorski:
image
Those tests work on my machine.

src/Renci.SshNet/Common/ChannelInputStream.cs
```diff
@@ -0,0 +1,221 @@
...
+            if (count == 0)
+            {
+                return;
+            }
+
+            _channel.SendData(buffer, offset, count);
+            _totalPosition += count;
+
+            // Must send EOF, otherwise SshCommand.EndExecute never gets called.
+            _channel.SendEof();
```
Rob-Hague: This does not look right. What if `Write` is called multiple times? Can another test be added which writes one byte at a time to validate that? It is a valid scenario when calling `stream.CopyTo` which calls `Write` under the hood

test/Renci.SshNet.IntegrationTests/OldIntegrationTests/SftpClientTest.Upload.cs
Rob-Hague: It also fails on my machine, but I've not gotten to the bottom of why (I think sshd is crashing but maybe SSH.NET is doing something wrong). I think we should revert this change for this PR

test/Renci.SshNet.TestTools.OpenSSH/SshdConfig.cs
```diff
@@ -385,6 +385,8 @@ private static void ProcessGlobalOption(SshdConfig sshdConfig, string line)
                 case "AuthorizedKeysFile":
                 case "PasswordAuthentication":
                 case "GatewayPorts":
+                // Had to add this otherwise docker container setup in integration test fails.
+                case "Include":
```
Rob-Hague: No need for the comment

```suggestion
                case "Include":
```

test/Renci.SshNet.Tests/Classes/Common/HostKeyEventArgsTest.cs
Rob-Hague: Seems like @realvizu has FIPS enabled. We should probably condition these tests for when [CryptoConfig.AllowOnlyFipsAlgorithms](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.cryptoconfig.allowonlyfipsalgorithms) is false. In the meantime, can we revert these changes for this PR?

test/Renci.SshNet.Tests/Renci.SshNet.Tests.csproj
```diff
@@ -3,6 +3,10 @@
     <TargetFrameworks>net462;net6.0;net7.0;net8.0</TargetFrameworks>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'net462' ">
+    <DefineConstants>$(DefineConstants);NET462_TARGET_FRAMEWORK</DefineConstants>
```
Rob-Hague: If we revert the FIPS test changes, then this is already unnecessary, but FYI the constant `NET462` is defined [by default](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives#conditional-compilation) so one can do `#if NET462` without any additional configuration

Rob-Hague (commented):
I don't think we should send channel EOF in `Write`. How about calling it only in `Dispose` and initialising `InputStream` lazily, something like:

```c#
var cmd = ...
using (cmd.InputStream) // This initialises InputStream
{
    myStream.CopyTo(cmd.InputStream);
} // This disposes InputStream and sends EOF

```

?

test/Renci.SshNet.IntegrationTests/OldIntegrationTests/SftpClientTest.Upload.cs
WojciechNagorski: Yes, I agree because this is not a part of this PR. @realvizu Can you revert this?

test/Renci.SshNet.Tests/Classes/Common/HostKeyEventArgsTest.cs
WojciechNagorski: Yes, I agree because this is not a part of this PR. @realvizu Can you revert this?

test/Renci.SshNet.Tests/Classes/PrivateKeyFileTest.cs
WojciechNagorski: @realvizu Can you revert this? It is not a part of this PR.

test/Renci.SshNet.IntegrationTests/OldIntegrationTests/SftpClientTest.Upload.cs
realvizu: Sure, reverted it.

test/Renci.SshNet.Tests/Classes/Common/HostKeyEventArgsTest.cs
realvizu: Sure, reverted it.

test/Renci.SshNet.Tests/Classes/PrivateKeyFileTest.cs
realvizu: Sure, reverted it.

realvizu:
> It looks good to me! But I need more time to test it!
> 
> Do you have a specific function in mind that you want to implement using this functionality? Did you see that it can be useful in [dotnet/Docker.DotNet#540](https://github.com/dotnet/Docker.DotNet/issues/540)

I had a very specific use case: I had to call some legacy system that is implemented as a set of SSH commands and one of the commands requires sending a file via input stream. These changes made it work.

src/Renci.SshNet/Common/ChannelInputStream.cs
realvizu: You're absoloutely right, I moved the EOF sending to Dispose, and added a unit test that writes the input one byte at a time.

test/Renci.SshNet.TestTools.OpenSSH/SshdConfig.cs
realvizu: OK, removed the comment.

test/Renci.SshNet.Tests/Renci.SshNet.Tests.csproj
realvizu: OK, removed it.

realvizu:
Thanks a lot @WojciechNagorski and @Rob-Hague for the comments and suggestions!

- I reverted all the FIPS-related changes and the change made to SftpClientTest.Upload.cs.
- I moved the EOF sending to Dispose and added a unit test that validates that now multiple calls to Write works as expected.
- I've adapted @Rob-Hague 's suggestion about making InputStream lazily initialized with some modification. I wanted to make it more apparent that a resource is created here (that should be disposed after using), so I changed the InputStream from a property to a CreateInputStream method.
- Also, because the input stream uses the channel created in BeginExecute and closed in EndExecute, I added a check that throws invalid operation exception if CreateInputStream is called when there's no channel to use. Also added some unit tests for this.

src/Renci.SshNet/Common/ChannelInputStream.cs
```diff
@@ -0,0 +1,220 @@
...
+using System.IO;
+
+using Renci.SshNet.Channels;
+
+namespace Renci.SshNet.Common
+{
+    /// <summary>
+    /// ChannelInputStream is a one direction stream intended for channel data.
+    /// </summary>
+    public class ChannelInputStream : Stream
```
Rob-Hague: Since the constructor is internal, perhaps the type could just be internal unless there is a reason to make it public

src/Renci.SshNet/Common/ChannelInputStream.cs
```diff
@@ -0,0 +1,220 @@
...
+            _channel.SendData(buffer, offset, count);
+            _totalPosition += count;
+        }
+
+        /// <summary>
+        /// Releases the unmanaged resources used by the Stream and optionally releases the managed resources.
+        /// </summary>
+        /// <param name="disposing"><c>true</c> to release both managed and unmanaged resources; <c>false</c> to release only unmanaged resources.</param>
+        /// <remarks>
+        /// Disposing a <see cref="PipeStream"/> will interrupt blocking read and write operations.
```
Rob-Hague: Maybe remove this remark :-)

src/Renci.SshNet/Common/ChannelInputStream.cs
```diff
@@ -0,0 +1,220 @@
...
+        /// </remarks>
+        protected override void Dispose(bool disposing)
+        {
+            if (!_isDisposed)
+            {
+                _isDisposed = true;
+
+                // Closing the InputStream requires sending EOF.
+                if (_totalPosition > 0 && _channel?.IsOpen == true)
```
Rob-Hague: This should guard on `disposing`

```suggestion
                if (disposing && _totalPosition > 0 && _channel?.IsOpen == true)
```

src/Renci.SshNet/SshCommand.cs
```diff
@@ -64,6 +65,25 @@ public class SshCommand : IDisposable
...
+ #pragma warning disable CA1859 // Use concrete types when possible for improved performance
+        public Stream CreateInputStream()
+#pragma warning restore CA1859 // Use concrete types when possible for improved performance
+        {
+            if (_channel == null)
+            {
+                throw new InvalidOperationException($"The input stream can be used only after calling BeginExecute and before calling EndExecute.");
+            }
+
+            _inputStream ??= new ChannelInputStream(_channel);
```
Rob-Hague: Perhaps it should throw an error if it's already not-null, instead of silently doing nothing?

Rob-Hague (approved):
Nice

src/Renci.SshNet/Common/ChannelInputStream.cs
realvizu: Sure.

src/Renci.SshNet/Common/ChannelInputStream.cs
realvizu: Sure.

src/Renci.SshNet/Common/ChannelInputStream.cs
realvizu: Right.

src/Renci.SshNet/SshCommand.cs
realvizu: Right.

realvizu:
Shall I do a squash+rebase onto develop before merging the PR?

Rob-Hague (approved):

Rob-Hague:
It will be merged with "Squash and merge" so it is not necessary

(it requires a CODEOWNERS approval before it can merge)

WojciechNagorski (approved):

WojciechNagorski:
There is a build problem:
> C:\projects\ssh-net\test\Renci.SshNet.IntegrationTests\SshClientTests.cs(37,33): error CS7036: There is no argument given that corresponds to the required parameter 'offset' of 'Stream.Write(byte[], int, int)' [C:\projects\ssh-net\test\Renci.SshNet.IntegrationTests\Renci.SshNet.IntegrationTests.csproj::TargetFramework=net48]

test/Renci.SshNet.IntegrationTests/SshClientTests.cs
```diff
@@ -17,12 +17,99 @@ public SshClientTests()
...
+            var inputByteArray = Encoding.UTF8.GetBytes("Hello world!");
+
+            // Make the server echo back the input file with "cat"
+            using (var command = _sshClient.CreateCommand("cat"))
+            {
+                var asyncResult = command.BeginExecute();
+
+                using (var inputStream = command.CreateInputStream())
+                {
+                    inputStream.Write(inputByteArray);
```
Rob-Hague:
```suggestion
                    inputStream.Write(inputByteArray, 0, inputByteArray.Length);
```

WojciechNagorski: Thanks

WojciechNagorski:

PR #1305, opened 2024-02-03
Use BinaryPrimitives in Pack

Rob-Hague:
These are more efficient than bit-shifting, so we should use them where possible.


WojciechNagorski (approved):

WojciechNagorski:

PR #1313, opened 2024-02-10
Add tests and benchmarks for ShellStream.Read and Expect

Rob-Hague:
Most of them are ignored because they fail.


WojciechNagorski (approved):
Good direction

WojciechNagorski:

PR #1315, opened 2024-02-11
Improve ShellStream Expect

WojciechNagorski:
There are three implementation of Expect method:
```c#
public string Expect(Regex regex, TimeSpan timeout);
public void Expect(TimeSpan timeout, params ExpectAction[] expectActions);
public IAsyncResult BeginExpect(TimeSpan timeout, AsyncCallback callback, object state, params ExpectAction[] expectActions);
```

Two of them return string ends with the expected expression.
Only the first one works differently. In this PR I fixed this problem.

Moreover, the _incoming array contains bytes and regex works on UTF8 encoding, so we should count how many bytes we should remove from the incoming array.

Continuation of #1313
Added test failures for #1207 (does not fix performance issue)

### DIFF
--- a/src/Renci.SshNet/ShellStream.cs
+++ b/src/Renci.SshNet/ShellStream.cs
@@ -268,8 +268,9 @@ namespace Renci.SshNet
                             if (match.Success)
                             {
                                 var result = text.Substring(0, match.Index + match.Length);
+                                var charCount = _encoding.GetByteCount(result);
 
-                                for (var i = 0; i < match.Index + match.Length && _incoming.Count > 0; i++)
+                                for (var i = 0; i < charCount && _incoming.Count > 0; i++)
                                 {
                                     // Remove processed items from the queue
                                     _ = _incoming.Dequeue();
@@ -348,7 +349,7 @@ namespace Renci.SshNet
         /// </returns>
         public string Expect(Regex regex, TimeSpan timeout)
         {
-            var text = string.Empty;
+            var result = string.Empty;
 
             while (true)
             {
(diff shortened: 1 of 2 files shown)


WojciechNagorski:
@Rob-Hague Can I ask for a review?

Rob-Hague (approved):
Thanks

I have nearly fixed the remaining tests

WojciechNagorski:
@Rob-Hague Thanks! I haven't touched the other tests and I won't.

src/Renci.SshNet/ShellStream.cs
```diff
@@ -471,6 +475,7 @@ public IAsyncResult BeginExpect(TimeSpan timeout, AsyncCallback callback, object
                                     if (match.Success)
                                     {
                                         var result = text.Substring(0, match.Index + match.Length);
+                                        var charCount = _encoding.GetByteCount(result);
 
                                         for (var i = 0; i < match.Index + match.Length && _incoming.Count > 0; i++)
```
jscarle, line 480: Shouldn't this line use charCount?

Rob-Hague: Yes

Rob-Hague:
Btw, working on ShellStream makes me wish it derived from StreamReader rather than Stream... it would save a lot of repeated encoding work for the Expect methods (and just make more sense in general). But perhaps not worth the break.

src/Renci.SshNet/ShellStream.cs
jscarle: I fixed it in my [PR](https://github.com/sshnet/SSH.NET/pull/1207).

jscarle:
I believe that ShellStream still has it's place, however perhaps a new ShellStreamReader could be added that would consume the a ShellStream and we could slowly migrate functionality between the two classes.

It could even be done in two steps. First by adding the new methods into ShellStreamReader, and then later deprecating them on ShellStream.

WojciechNagorski:

PR #1316, opened 2024-02-12
Support ETM (Encrypt-then-MAC) variants for HMAC

### LINKED ISSUE #837: Server HMAC algorithm not found
Hi ,

I used the shClient.Connect(), and catch an error "Server HMAC algorithm not found";I used the newest version 2020.0.1, and the server only support the "hmac-sha2-512-etm@openssh.com,hmac-sha2-256-etm@openssh.com,umac-128-etm@openssh.com"  MAC algorithms, How can I do? Is there a solution to add MAC algorithms or fix this error?

Thanks

### LINKED ISSUE #875: Can release/add a new algorithm
Hi Team,

The latest library does not implement the hmac-sha2-256-etm@openssh.com or hmac-sha2-512-etm@openssh.com algorithms.

Can someone advice how to add or use? Thanks!!

### LINKED ISSUE #1050: Server HMAC algorithm not found
I have a problem connecting to my linux server.

When I try to connect, I get a "Server HMAC algorithm not found" message.
I check the log and I find that lib cannot use the hmac-sha2-256-etm algorithm.

My server only accepts these 3 algorithms:
- hmac-sha2-256-etm@openssh.com
- hmac-sha2-512-etm@openssh.com
- umac-128-etm@openssh.com

[#825](https://github.com/sshnet/SSH.NET/issues/825) does not work

### LINKED ISSUE #1092: Unable to connect to AWS SFTP Server
I was hoping someone might be able to give some guidance on configuration of server or this library as we cannot connect to a customers SFTP Server they setup on AWS.

I have tried various formats of private key from PuttyGen; nothing really help.

When attempting a connection using a private key I get error:
`An established connection was aborted by the server.`

Server: `SSH-2.0-AWS_SFTP_1.1`

Connection Info:
image

scott-xu:
Should fix #875 and fix #1050 and fix #837 and fix #1092

References:
1. https://github.com/openssh/openssh-portable/blob/cbbdf868bce431a59e2fa36ca244d5739429408d/PROTOCOL
2. https://github.com/openssh/openssh-portable/blob/cbbdf868bce431a59e2fa36ca244d5739429408d/packet.c

Written by hand **without testing** by now.


scott-xu:
Would you mind giving a try for this PR?
Note: it is written without testing at this moment. Feel free to improve the PR if you want.
@PashCracken @SkymaX86 @robert-scheck @alef75 @Jedrzej94 @Ph43n0m @AliceTJ

Rob-Hague:
Nice. Testing should be straightforward:

```diff
diff --git a/test/Renci.SshNet.IntegrationTests/HmacTests.cs b/test/Renci.SshNet.IntegrationTests/HmacTests.cs
index 993e5ec98..e5057248c 100644
--- a/test/Renci.SshNet.IntegrationTests/HmacTests.cs
+++ b/test/Renci.SshNet.IntegrationTests/HmacTests.cs
@@ -58,6 +58,24 @@ public void HmacSha2_512()
             DoTest(MessageAuthenticationCodeAlgorithm.HmacSha2_512);
         }
 
+        [TestMethod]
+        public void HmacSha1Etm()
+        {
+            DoTest(MessageAuthenticationCodeAlgorithm.HmacSha1Etm);
+        }
+
+        [TestMethod]
+        public void HmacSha2_256_Etm()
+        {
+            DoTest(MessageAuthenticationCodeAlgorithm.HmacSha2_256_Etm);
+        }
+
+        [TestMethod]
+        public void HmacSha2_512_Etm()
+        {
+            DoTest(MessageAuthenticationCodeAlgorithm.HmacSha2_512_Etm);
+        }
+
         private void DoTest(MessageAuthenticationCodeAlgorithm macAlgorithm)
         {
             _remoteSshdConfig.ClearMessageAuthenticationCodeAlgorithms()
```

src/Renci.SshNet/Session.cs
```diff
@@ -1257,6 +1279,20 @@ private Message ReceiveMessage(Socket socket)
                 }
             }
 
+            // validate encrypted message against MAC
+            if (_serverMac != null && _serverMac.ETM)
+            {
+                var clientHash = _serverMac.HashAlgorithm.ComputeHash(data, blockSize, data.Length - blockSize - serverMacLength);
```
Rob-Hague:
```suggestion
                var clientHash = _serverMac.HashAlgorithm.ComputeHash(data, 0, data.Length - serverMacLength);
```

scott-xu:
Do we want to support `HmacSha1Etm`? IMHO, `HmacSha2_256_Etm` and `HmacSha2_512_Etm` is sufficient (and secure). @Rob-Hague

Rob-Hague:
I think we might as well, while we still have `hmac-sha1`

src/Renci.SshNet/Security/Cryptography/HMAC.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System;
+using System.Security.Cryptography;
+
+namespace Renci.SshNet.Security.Cryptography
+{
+    /// <summary>
+    /// Represents the info for Message Authentication Code (MAC).
+    /// </summary>
+    public sealed class HMAC : IDisposable
```
Rob-Hague: The new type seems unnecessary. Could we just change `IKeyExchange.Create[..]Hash` to `Create[..]Hash(out bool isEncryptThenMac)` or to return a tuple e.g. `(HashAlgorithm, bool) Create[..]Hash()`?

src/Renci.SshNet/Messages/Message.cs
```diff
@@ -37,7 +37,7 @@ protected override void WriteBytes(SshDataStream stream)
             base.WriteBytes(stream);
         }
 
-        internal byte[] GetPacket(byte paddingMultiplier, Compressor compressor)
```
Rob-Hague, line 40: fyi @jacobslusser (or anyone else interested), the Message->Socket logic (in `Session.SendMessage` and `Message.GetPacket`) is somewhat cumbersome. There is probably some efficiency to be gained in untangling it.

Rob-Hague:
Looks good to me otherwise

src/Renci.SshNet/Security/Cryptography/HMAC.cs
scott-xu: My first revision (not committed) is `Create[...]Hash(out bool etm)`. However, compared with other argumentless `Create[...]()` methods in `IKeyExchange` interface, it looks a little bit different.

Then I took a look at other implementations, like [OpenSSH](https://github.com/openssh/openssh-portable) and [sshj](https://github.com/hierynomus/sshj). They both have dedicated data structure for MAC. Their implementation defines the detailed hash related methods plus `etm` or `isEtm` property.

https://github.com/openssh/openssh-portable/blob/master/mac.c
https://github.com/hierynomus/sshj/blob/master/src/main/java/net/schmizz/sshj/transport/mac/MAC.java

Rob-Hague: Right, but we already have HashInfo, which is practically the same thing. We don't need another type representing the same information.

scott-xu: From my understanding, the `HashInfo` is kind of the factory. The proposed `HMAC` is the _product_. I couldn't find a way to merge them. Should we merge the two?

src/Renci.SshNet/HashInfo.cs
```diff
@@ -20,17 +20,22 @@ public class HashInfo
         /// <summary>
         /// Gets the cipher.
         /// </summary>
-        public Func<byte[], HashAlgorithm> HashAlgorithm { get; private set; }
```
WojciechNagorski: Breaking change

src/Renci.SshNet/Security/Cryptography/HMAC.cs
WojciechNagorski: How about removing the HMAC class and move ETM information to HashInfo?
You've already added the property here:
https://github.com/sshnet/SSH.NET/pull/1316/files#diff-8ad50217d0fd4634eedc20ae503cfa6584eb26b59c5922c807b581c2455f4e25R25-R28

This would protect us from the breaking change from my second comment.

scott-xu: That's a good spot! The `ETM` property was added previously and then forgot to delete.
At the very beginning, I had the very similar thoughts with you which is to leverage `HashInfo` to store `ETM` Boolean value. However, later on I found the `HashInfo` is kind of factory. The `HashAlgorithm` is the _product_.

See https://github.com/sshnet/SSH.NET/blob/develop/src/Renci.SshNet/Security/IKeyExchange.cs#L72 and https://github.com/sshnet/SSH.NET/blob/develop/src/Renci.SshNet/Security/IKeyExchange.cs#L80

The `HashAlgorithm` is a .NET build-in class which cannot add new properties. The only way is to change `IKeyExchange.Create[..]Hash` to `Create[..]Hash(out bool isEncryptThenMac)` (which is my first revision) or to return a tuple e.g. `(HashAlgorithm, bool) Create[..]Hash()` just as @Rob-Hague pointed. I've already explained the reason why I abandoned my first revision and created dedicated `HMAC` class. See my previous comments.

I know it is a breaking change, but I don't think it will bring critical impact. There was similar successful breaking change previously. See https://github.com/sshnet/SSH.NET/commit/24838e6173e2a7d4cbb498ce5edc3e32b254ca11

Rob-Hague: I do think we should either merge the classes or go with the `out bool` parameter in `IKeyExchange`, and I would suggest doing the latter. It is a much simpler change and creates one small break in the internals of the library where it matters a lot less.

scott-xu: The `Cipher` returned from `CreateServerCipher` and `CreateClientCipher` is defined in SshNet. The `Compressor` returned from `CreateCompressor` and `CreateDecompressor` is also defined in SshNet.
The `HashAlgorithm` returned from `CreateServerHash` and `CreateClientHash` is a BCL class which prevents us from extending it.
There's already a bunch of `HMAC[...]` classes in SshNet, which has the same name with BCL `HMAC[...]`, specifically used for SSH purpose.

WojciechNagorski:
If it weren't a breaking change, I would approve it. I need some time to think it over and check it out. 

@Rob-Hague any idea?

Rob-Hague:
Yes I think we should go with the bool property on HashInfo and change `IKeyExchange.CreateServerHash()` from `HashAlgorithm CreateServerHash()` to `HashAlgorithm CreateServerHash(out bool isEncryptThenMac)`. I don't see the problem with that.

scott-xu:
Sorry I don't understand why you against `HMAC` so much.

The `HashAlgorithm CreateServerHash(out bool isEncryptThenMac)` is a break change anyway.

Shouldn't we focus on the implementation details of ETM? That's the main purpose of this PR.

scott-xu:
> I know everyone else's code always smells, but enough is enough - this is a collaboration, not a competition.

-- @zybexXL

Rob-Hague:
I think we've made it pretty clear that it's an unnecessary breaking change

`IKeyExchange` takes a break either way. That's fine: it's in the internals of the library and less likely to affect anyone.

Breaking HashInfo does not need to happen. It is much closer to the common API that people use.

You said you had already gone in this direction in an earlier iteration. I've not seen any reason to stray from that.

> Shouldn't we focus on the implementation details of ETM? That's the main purpose of this PR.

I've checked the details. They look good.

scott-xu:
I've already explained why I abandoned the first iteration. I don't think `HashInfo` is widely used externally just like you believe `IKeyExchange` is not widely used externally.

scott-xu:
To get the ball roll, someone has to compromise. Let me update the PR.

Rob-Hague (approved):
Thanks for making the change. It looks great to me. Another big pain point of the library bites the dust 🙂

Perhaps you could update the description to close a bunch of issues

scott-xu:
> Thanks for making the change. It looks great to me. Another big pain point of the library bites the dust 🙂
> 
> Perhaps you could update the description to close a bunch of issues

The description has had the format to close the 2 related issues. Are there more?

Rob-Hague:
Only one more as it turned out, I thought there were a lot more. I've updated the description

WojciechNagorski (approved):

WojciechNagorski:
@scott-xu thanks!

scott-xu:
Cheers!

PR #1317, opened 2024-02-13
No need to collect nuget packages in appveyor.yml as the packages are not generated when build

scott-xu:
No nupkg nor snupkg is generated at this moment. We can add these 2 lines back later if we decide to generate nuget packages when build.


Rob-Hague:
Could you also delete the artifacts/tcpdump.pcap comments while you are here?

scott-xu:
> Could you also delete the artifacts/tcpdump.pcap comments while you are here?

TBH, I'm not familiar with this .pcap file in artifacts folder. I know it has already been commented out, but this PR is focused on not collecting nuget packages.
And I know this PR's title is not intuitive. But if we want to have a better name, it would be easy when concentrate on one thing.

Rob-Hague:
Thanks. I've removed them (I added them)

Rob-Hague (approved):

WojciechNagorski (approved):

PR #1319, opened 2024-02-13
Fixed dequeuing of incoming queue

jscarle:
Fixed the dequeueing of the _incoming queue following #1207 and added the related test to protect against future regressions.

### DIFF
--- a/src/Renci.SshNet/ShellStream.cs
+++ b/src/Renci.SshNet/ShellStream.cs
@@ -282,11 +282,14 @@ namespace Renci.SshNet
 
                             if (match.Success)
                             {
-                                var returnText = matchText.Substring(0, match.Index + match.Length);
-                                var returnLength = _encoding.GetByteCount(returnText);
+#if NETSTANDARD2_1_OR_GREATER || NET6_0_OR_GREATER
+                                var returnLength = _encoding.GetByteCount(matchText.AsSpan(0, match.Index + match.Length));
+#else
+                                var returnLength = _encoding.GetByteCount(matchText.Substring(0, match.Index + match.Length));
+#endif
 
                                 // Remove processed items from the queue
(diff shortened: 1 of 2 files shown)


jscarle:
@WojciechNagorski Following @Rob-Hague 's observation in #1207, there is indeed an issue where the incoming queue may not dequeue at the same rate as the expect queue if the data that has not been processed exceeds the size of the expect buffer. An improbable case, but nonetheless, this PR fixes that.

jscarle:
Additionally, I made sure that the full incoming queue is returned when Expect matches and not just the expected queue.

src/Renci.SshNet/ShellStream.cs
```diff
@@ -869,5 +846,37 @@ private void OnDataReceived(byte[] data)
...
+        private string SyncQueuesAndReturn(int bytesToDequeue)
+        {
+            string incomingText;
+
+            lock (_incoming)
+            {
+                var incomingLength = _incoming.Count - _expect.Count + bytesToDequeue;
+                incomingText = _encoding.GetString(_incoming.ToArray(), 0, incomingLength);
+
+                SyncQueuesAndDequeue(bytesToDequeue);
```
WojciechNagorski, line 859: SyncQueuesAndDequeue method has `lock (_incoming)` and calling is also in the same `lock`. How does it work? I would have thought it would be blocked.

jscarle: Locks can be nested. See [this conversation](https://stackoverflow.com/questions/5035126/why-do-nested-locks-not-cause-a-deadlock).

jscarle:
@WojciechNagorski @Rob-Hague This is ready to be merged.

WojciechNagorski (approved):

PR #1321, opened 2024-02-13
Limit TimeSpan timeouts to Int32 MaxValue

### LINKED ISSUE #14: ConnectionInfo Timeout doesn't allow for Int64.MaxValue as designed
The ConnectionInfo class makes use of a Timeout property of type TimeSpan that allows for an Int64 constructor parameters upon instantiation. A Developer can either use the default Timeout value of 30 seconds or set their own value up to Int64.MaxValue. However the underlying Socket construct's "SendTimeout" and "RecieveTimeout" only allow for Int32 values. This issue throws an OutOfRangeExeception at runtime when attempting to set the TimeSpan value to a tick value greater than Int32.MaxValue. The maximum value of TimeSpan is designed to allow for 29K years using Int64 values of the 100ns scale, but the underlying Socket's maximum value uses Int32 on the millisecond scale, which is roughly 24 days.

jscarle:
Added guard clauses to various timeouts to ensure they don't exceed an Int32 in milliseconds.

Resolves #14.


jscarle:
@WojciechNagorski @Rob-Hague Ready for review and merge.

src/Renci.SshNet/Common/TimeSpanExtensions.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
+#if NET21_OR_GREATER || NET6_0_OR_GREATER
```
Rob-Hague: Do you mean `NETSTANDARD2_1_OR_GREATER`?

src/Renci.SshNet/Common/TimeSpanExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// Thrown when <paramref name="timeSpan"/> does not represent a value between -1 and <see cref="int.MaxValue"/>, inclusive.
+        /// </exception>
+        public static int AsTimeout(this TimeSpan timeSpan,
+#if NET21_OR_GREATER || NET6_0_OR_GREATER
+                                    [CallerArgumentExpression(nameof(timeSpan))]
+#endif
+                                    string callerMemberName = "")
+        {
+            var timeoutInMilliseconds = timeSpan.TotalMilliseconds;
+            return timeoutInMilliseconds is < -1d or > int.MaxValue
```
Rob-Hague, line 24: What if it's -0.5?

Rob-Hague (commented):
I'm not seeing what problem this solves. If an underlying API doesn't accept a certain range of TimeSpan, aren't we going to get an error either way? Isn't it better to let the API decide?

src/Renci.SshNet/Common/TimeSpanExtensions.cs
jscarle: Yes, you're right.

src/Renci.SshNet/Common/TimeSpanExtensions.cs
jscarle: I'm not sure that timespan allows fractional milliseconds, I'll check.

jscarle:
> I'm not seeing what problem this solves. If an underlying API doesn't accept a certain range of TimeSpan, aren't we going to get an error either way? Isn't it better to let the API decide?

I'm not sure what you mean by "the API".

But basically, the issue boils down to the fact that the timeouts are injested by SSH.NET's public methods as TimeSpans but then used internally for timeouts that int based. The goal is to bring validation to the outer edge of the public methods of SSH.NET to avoid OutOfRange exceptions that could come from the internals of SSH.NET.

src/Renci.SshNet/Common/TimeSpanExtensions.cs
jscarle: (int) of a double truncates the decimal. Do we want to round or is truncating ok?

src/Renci.SshNet/SshCommand.cs
```diff
@@ -44,7 +45,19 @@ public class SshCommand : IDisposable
...
+        {
+            get
+            {
+                return _commandTimeout;
+            }
+            set
+            {
+                _commandTimeout.EnsureValidTimeout();
+
+                _commandTimeout = value;
```
WojciechNagorski: `_commandTimeout.EnsureValidTimeout();` -> `value.EnsureValidTimeout();`?

WojciechNagorski:
How to reproduce the error or how does it manifest itself? I don't know what's improved here either?

Can you add a test that didn't work before but now works?

jscarle:
@Rob-Hague @WojciechNagorski The latest version that's in develop gives me a ton of IDE0005 build errors:

image

src/Renci.SshNet/Common/TimeSpanExtensions.cs
jscarle: According to [the docs](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.callerargumentexpressionattribute?view=net-8.0), the attribute became available in .NET Core 3.0, so I've updated it to `#if NETCOREAPP3_0_OR_GREATER`.

src/Renci.SshNet/SshCommand.cs
jscarle: Fixed, good catch!

jscarle:
> How to reproduce the error or how does it manifest itself? I don't know what's improved here either?
> 
> Can you add a test that didn't work before but now works?

I'll take care of it.

Rob-Hague:
> I'm not sure what you mean by "the API".

I meant, if we're forwarding the TimeSpan to a method on e.g. `Socket` or `Monitor`, then those methods will perform their own validation and throw if they don't accept a certain value, so there should be no need to try and presume what values are acceptable or not.

I guess I felt like I was missing some context or motivation for this change. But I think it's OK in principal

> The latest version that's in develop gives me a ton of IDE0005 build errors

That sounds like VS weirdness. It's OK for me (although sometimes it feels like the IDE doesn't know what to do with so many analyzers)

src/Renci.SshNet/Common/TimeSpanExtensions.cs
Rob-Hague: Casting to int is fine. Despite my comment, no-one should reasonably expect a timeout to conform to a fractional number of milliseconds

src/Renci.SshNet/Common/TimeSpanExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// </exception>
+        public static int AsTimeout(this TimeSpan timeSpan,
+#if NETCOREAPP3_0_OR_GREATER
+                                    [CallerArgumentExpression(nameof(timeSpan))]
+#endif
+                                    string callerMemberName = "")
+        {
+            var timeoutInMilliseconds = timeSpan.TotalMilliseconds;
+            return timeoutInMilliseconds is < -1d or > int.MaxValue
+                       ? throw new ArgumentOutOfRangeException(callerMemberName, "The timeout must represent a value between -1 and Int32.MaxValue, inclusive.")
```
Rob-Hague: I know this message was copied from some existing code, but it doesn't mention anything about milliseconds, and it probably should

```suggestion
                       ? throw new ArgumentOutOfRangeException(callerMemberName, "The timeout must represent a value between -1 and Int32.MaxValue milliseconds, inclusive.")
```

test/Renci.SshNet.Tests/Classes/NetConfClientTest.cs
```diff
@@ -105,7 +107,9 @@ public void OperationTimeout_GreaterThanLowerLimit()
             {
                 Assert.IsNull(ex.InnerException);
                 ArgumentExceptionAssert.MessageEquals("The timeout must represent a value between -1 and Int32.MaxValue, inclusive.", ex);
+#if NETCOREAPP3_0_OR_GREATER
                 Assert.AreEqual("value", ex.ParamName);
+#endif
```
Rob-Hague: On lower targets, `ex.ParamName` is going to be `""`. It would probably be better to specify `nameof` at the callsite and remove the `[CallerArgumentExpression]` attribute?

jscarle:
> > I'm not sure what you mean by "the API".
> 
> I meant, if we're forwarding the TimeSpan to a method on e.g. `Socket` or `Monitor`, then those methods will perform their own validation and throw if they don't accept a certain value, so there should be no need to try and presume what values are acceptable or not.
> 
> I guess I felt like I was missing some context or motivation for this change. But I think it's OK in principal

Generally speaking, it is considered good design that libraries validate and throw early in public API methods. As an example of this, you can take analyzer rule [CA1062](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1062?view=vs-2019) for null checks. The original reason I looked into this was because of #14, which although at the surface may seem superfluous, is actually valid.

jscarle:
> > The latest version that's in develop gives me a ton of IDE0005 build errors
> 
> That sounds like VS weirdness. It's OK for me (although sometimes it feels like the IDE doesn't know what to do with so many analyzers)

I'm using Rider and dotnet build with SDK 8.0.200. So I don't think this has anything to do with VS.

src/Renci.SshNet/Common/TimeSpanExtensions.cs
jscarle: Fixed the error message.

test/Renci.SshNet.Tests/Classes/NetConfClientTest.cs
jscarle: I removed `[CallerArgumentExpression]` and replaced it with the relevant `nameof`. Turns out the result is better as the argument will now be against the property name and not just `value`.

jscarle:
> > > The latest version that's in develop gives me a ton of IDE0005 build errors
> > 
> > 
> > That sounds like VS weirdness. It's OK for me (although sometimes it feels like the IDE doesn't know what to do with so many analyzers)
> 
> I'm using Rider and dotnet build with SDK 8.0.200. So I don't think this has anything to do with VS.

I added `IDE0005` to the `.editorconfig`.

jscarle:
@WojciechNagorski @Rob-Hague Ready for final review and merge.

src/Renci.SshNet/Abstractions/SocketAbstraction.cs
```diff
@@ -129,7 +129,7 @@ public static void ClearReadBuffer(Socket socket)
 
         public static int ReadPartial(Socket socket, byte[] buffer, int offset, int size, TimeSpan timeout)
         {
-            socket.ReceiveTimeout = (int) timeout.TotalMilliseconds;
+            socket.ReceiveTimeout = timeout.AsTimeout(nameof(socket));
```
WojciechNagorski: Why `nameof(socket)`? not `nameof(timeout)`?

src/Renci.SshNet/Abstractions/SocketAbstraction.cs
```diff
@@ -274,7 +274,7 @@ public static int Read(Socket socket, byte[] buffer, int offset, int size, TimeS
             var totalBytesRead = 0;
             var totalBytesToRead = size;
 
-            socket.ReceiveTimeout = (int) readTimeout.TotalMilliseconds;
+            socket.ReceiveTimeout = readTimeout.AsTimeout(nameof(Timeout));
```
WojciechNagorski: Why not `nameof(readTimeout)`?

src/Renci.SshNet/Abstractions/SocketAbstraction.cs
jscarle: Fixed.

src/Renci.SshNet/Abstractions/SocketAbstraction.cs
jscarle: Fixed.

jscarle:
@WojciechNagorski I fixed the reference you mentioned. I must have been really tired, that slipped right past me.

src/Renci.SshNet/.editorconfig
```diff
@@ -159,3 +159,7 @@ dotnet_diagnostic.IDE0048.severity = none
 # IDE0305: Collection initialization can be simplified
 # https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0305
 dotnet_diagnostic.IDE0305.severity = none
+
+# IDE0005: Remove unnecessary using directives
+# https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0005
+dotnet_diagnostic.IDE0005.severity = suggestion
```
WojciechNagorski, line 165: Can you revert this? I think after restarting VS (or restarting your computer) everything should work.

This is the last message and then I will be able to merge this PR.

Rob-Hague: I am seeing this locally as well, the errors look real. I think it could be related to upgrading the SDK but I have no idea why we haven't seen it before.

I am trying to fix it with `dotnet format` but not having much luck :-(

WojciechNagorski: I was wrong, I don't know what's going on.

WojciechNagorski (approved):
LGTM!

PR #1322, opened 2024-02-13
Fix a few issues with ShellStream

### LINKED ISSUE #63: ShellStream.Read not blocking when no data available

### LINKED ISSUE #180: Wrong commands sequence. ShellStream.WriteLine didn't flush buffer before send its data.
```
using (var sshStream = sshClient.CreateShellStream("C#", 80, 25, 100, 100, 1000))
{
    new MemoryStream(sshClient.ConnectionInfo.Encoding.GetBytes("echo 'command 1'\r")).CopyTo(sshStream);
    sshStream.WriteLine("echo 'command 2'");
    sshStream.Flush();
}

```
Expected result:
   command 1
   command 2

Current result:
  command 2
  command 1

My use case:
 I open embedded resource (somescript.sh) as stream and want to send all content of the somescript.sh to sshStream...

### LINKED ISSUE #301: ShellStream: Write(String) and WriteLine(String) do not use write buffer
The `Wrìte(String)` and `WriteLine(String)` methods in **ShellStream** currently do not write to the buffer, and do not flush the buffer before data is send to the server.

Consider the following example:

```cs
var abcBytes = Encoding.UTF8.GetBytes("abc");
var ghiBytes = Encoding.UTF8.GetBytes("ghi");
shellStream.Write(abcBytes, 0, abcBytes.Length); // writes to buffer, unless buffer is full
shellStream.Write("def");
shellStream.Write(ghiBytes, 0, ghiBytes.Length); // writes to buffer, unless buffer is full
shellStream.WriteLine("jkl");
shellStream.Flush();
```
**Expected result:**

All bytes are buffered, and - upon invoking `Flush()` - the following text is sent to the server in a single **SSH_MSG_CHANNEL_DATA** message:

- abcdefghijkl\r

**Actual result:**

Tollowing fragments are sent in the order list, and each fragment is sent in a separate **SSH_MSG_CHANNEL_DATA** message:

- def
- jkl\r
- abcghi

### LINKED ISSUE #302: ShellStream: Buffer is not flushed before read
When one of the `Read(Line)` or `Expect` overloads is invoked, the write buffer should first be flushed.

### LINKED ISSUE #365: [Feature Request]  ShellStream.Expect - Abort if No Match Found Within a Specified Timeout Period
The ShellStream Expect method currently supports a timeout as follows:
```csharp
public string Expect(Regex regex, TimeSpan timeout)
```

I believe that when standard output is detected, examining it is a synchronous operation. Consequently, if the standard output consists of thousands of lines then you have to wait for as long as it takes and there is no way to abort:
```csharp
text = _encoding.GetString(_incoming.ToArray(), 0, _incoming.Count);
```

I am currently using `Expect` against a process which can potentially emit thousands of lines to standard output. What I would like is a way to abort `Expect` if no match is found within a specified period of time.

I am not a C# programmer but am more than happy to conduct some testing and build the project from source if anyone can suggest any changes to make this feature request possible.

### LINKED ISSUE #453: Race Condition in ShellStream causing NullReferenceException
Issue:
Race condition in ShellStream occurs when underlying connection abruptly disconnects. The disconnect causes the stream to dispose. The Expect(Regex, TimeSpan) method utilizes an AutoResetEvent used to synchronize when new data is received, but the AutoResetEvent is disposed of and set to null.

Fix:
Do not dispose of an object that could be in use by another thread. In this particular context, the Producer (underlying connection) has abruptly finished, so the ShellStream would need have a state that reflects this. Any methods that utilize the existing (and only remaining) data, should not wait for more data to arrive and either return null (such as Expect) immediately, or throw SshConnectionException. The Consumer (user of the API) should be the one that disposes.

### LINKED ISSUE #672: Close of channel should not dispose ShellStream

### LINKED ISSUE #714: ShellStream.Expect TimeSpan timeout, not working.
Microsoft Visual Studio Professional 2019 Version 16.6.5
.NET Framework 4.7.2
SSN.NET v2016.1.0
==================
I'm using the latest SSH.NET with ShellStream and when calling the Expect(Regex, TimeSpan) the timeout doesn't seem to work at all.    I've also tried using TimeSpan.FromSeconds(30) with the same results.   If the expected information comes back, it works fine.  

```
this.shellStream.WriteLine(command);
var expectRegex = new Regex(@"\[(.*?)\][$#]|[$]");
var output = this.shellStream.Expect(expectRegex, new TimeSpan(0, 0, 30));
```

### LINKED ISSUE #748: ShellStream.ReadLine only recognises CrLf as a line ending
I am connecting to a server that ends its lines with a single Cr. Because ShellStream.ReadLine is looking for CrLf it is not finding the end of the line.

I propose that this method be changed to look for CrLf or Cr or Lf that should cover most cases.

### LINKED ISSUE #917: Cannot access a disposed object (ShellStream)

### LINKED ISSUE #1320: Discussion: Class-level expectSize or method-level expectSize.
@WojciechNagorski @Rob-Hague

As it stands, there are two public APIs for setting the expectSize buffer.

https://github.com/sshnet/SSH.NET/blob/d07827b6c801a91eb37d1519d56a621ed6e2b0ce/src/Renci.SshNet/SshClient.cs#L446
https://github.com/sshnet/SSH.NET/blob/d07827b6c801a91eb37d1519d56a621ed6e2b0ce/src/Renci.SshNet/SshClient.cs#L509

By default, unless specified otherwise, the expectSize will default to 2 * bufferSize. Since bufferSize defaults to 1024, that means expectSize defaults to 2048. As a default, that should cover most uses cases considering a full 80 x 25 screen is 2000 characters.

Should we remove the two public APIs that I originally added in #1207 before the next release in order to prevent any dependencies on it and allow us to rework it as a method-level parameter?

Rob-Hague:
~~I have not yet rebased this on the expectSize changes merged today, pending further discussion. So there are merge conflicts from the start. Just wanted to get it out there.~~

---

The main change is to replace the Queue<byte> with a byte[] and a couple of variables which index the start and end of the data. The remainder is mainly slightly more careful locking semantics.

One possibly contentious point: in fixing the Write behaviour (#301) I chose to remove the outgoing buffer and immediately send the data across the channel. Write(string) and WriteLine(string) were already doing this, and I felt it was better to change Write(byte[]) to match rather than changing the string methods. This is the reason for the number of deleted test classes for Write.

It also makes the definition of infinite timeout to consistently be -1 milliseconds instead of 0, as is standard in other .NET methods with a wait period. This was not documented previously.

The rest of the changes are hopefully fairly agreeable. Happy to clarify anything.

Here are the (allocation) numbers on 3bfac50a, which this is based on:

| Method              | Mean     | Error   | StdDev  | Gen0      | Gen1      | Allocated   |
|-------------------- |---------:|--------:|--------:|----------:|----------:|------------:|
| ShellStreamReadLine | 865.1 ms | 7.89 ms | 6.99 ms |         - |         - |   507.98 KB |
| ShellStreamExpect   | 864.8 ms | 5.95 ms | 5.57 ms | 6000.0000 | 6000.0000 | 12565.52 KB |

And here is after:

| Method              | Mean     | Error   | StdDev  | Allocated |
|-------------------- |---------:|--------:|--------:|----------:|
| ShellStreamReadLine | 866.6 ms | 9.37 ms | 8.77 ms | 389.62 KB |
| ShellStreamExpect   | 851.1 ms | 8.27 ms | 7.74 ms | 408.65 KB |

(for reference, current develop d07827b6)

| Method              | Mean     | Error    | StdDev  | Gen0      | Gen1      | Allocated  |
|-------------------- |---------:|---------:|--------:|----------:|----------:|-----------:|
| ShellStreamReadLine | 866.3 ms | 10.52 ms | 9.32 ms |         - |         - |  510.59 KB |
| ShellStreamExpect   | 864.7 ms |  6.10 ms | 5.70 ms | 1000.0000 | 1000.0000 | 3321.11 KB |

closes #63
closes #453 
closes #672 
closes #917 
closes #180 
closes #301 
closes #365 (the previous implementation would read for the full timeout on each update, rather than just how long we have left)
closes #714
closes #748
closes #302
closes #1320 
closes #923 
closes #1024


src/Renci.SshNet/ShellStream.cs
```diff
@@ -237,160 +231,188 @@ public override void SetLength(long value)
...
+                while (true)
                 {
-                    if (_incoming.Count > 0)
-                    {
-                        result = _encoding.GetString(_incoming.ToArray(), 0, _incoming.Count);
-                    }
+                    AssertValid();
 
-                    var match = regex.Match(result);
+                    var bufferText = _encoding.GetString(_buffer, _head, _tail - _head);
```
Rob-Hague, line 418: To integrate the expectSize, it would probably just be a case of restricting the search space here, e.g.

```c#
var expectHead = Math.Max(_head, _tail - _expectSize);
var bufferText = _encoding.GetString(_buffer, expectHead, _tail - expectHead);
```

The question is whether to put it as a parameter on the method or the constructor (I would probably prefer method)

WojciechNagorski: I also prefer a method parameter with the default value it will work for everyone without any breaking change

WojciechNagorski:
It looks great! I'm patiently waiting for the merge with the develop branch

src/Renci.SshNet/ShellStream.cs
Rob-Hague: I went with the parameter and named it `windowSize`, which felt to me a little more intuitive in meaning.

@jscarle, regretfully these changes (as they stand) reduce most of your recent contributions to be only in spirit rather than in code. I hope you consider the end result to be a net positive nonetheless. These recent changes wouldn't have happened without your motivating persistence.

Rob-Hague: also on this line: unfortunately I think we still need to materialise the entire buffer as a string for `Expect(Regex)` in order to correctly deal with multi-byte encoding. (`Expect(string)` does not suffer from this because it works directly on the bytes). We can at least still use a reduced window for the regex match itself.

jscarle: In the end, what's important is that collaboratively we'll have managed to drastically improve Expect and fix bugs with ShellStream. These are issues that date back years now and made the previous implementation difficult to use, and in some cases impossible to use, in long running scenarios.

As for windowSize, I'd recommend against using the word window in this context. As we've been putting our hands in the plumbing, it may be obvious to us, however to someone consuming the API, it may not. I feel that it is too confusing and may lead people to misinterpret that as having some sort of relationship with the terminal window size.

In terms of performance, I'm glad that although my contribution has literally been obliterated by this PR, at least I was on par with your version in terms of execution time. My allocation improved, but your version is obviously much more efficient.

It would be a good idea to performance test Expect with a very large incoming buffer (in the several MB range) to ensure that there's no regression of the original performance issues that brought me to run a parallel queue to begin with.

Rob-Hague: Right, `windowSize` is probably not a great name here. And I'd already been a bit confused by `expectSize`. I'll change it later. So far my best attempt is `rollingWindowSize`

jscarle: We could use one of these:
- match
- search
- inspect / inspection
- observation 
- peek
- frame
- analysis
- lookahead
- lookbehind

And/or a combination of above with Max, Size, Length, Range, or Bytes. ie: matchLength, peekSize.

Personally, I like matchLength, searchSize, inspect, and inspectSize. Though my most preferred is simply **inspect**. It's short, consice, and unambiguous.

Rob-Hague: Thanks for the suggestions. I went with `lookback`

WojciechNagorski:
Is it ready for review?

Rob-Hague:
Yes, thanks

WojciechNagorski:
Changing the behavior of the Write method is controversial, but we can try.

Rob-Hague:
Yeah, I think either `WriteLine` and `Write(string)` needed to be buffered (which would require users to add a call to `Flush`), or `Write(byte[])` needed to not buffer. I think changing `Write(byte[])` to not buffer is the safest thing there.

edit: it could be flagged in the release notes as a behavioural breaking change

jscarle:
I propose to do it the other way around. WriteLine could call Flush internally. And Write could buffer.

jscarle:
Basically, my reasoning is that if I'm consuming ShellStream in my application, when I call WriteLine, I'm done. That's what I wanted to send. While using Write, maybe I have several different Writes to do before I'm done.

Rob-Hague:
Yep that makes sense, but note that `Write(string)` will also have to flush internally to meet the same behaviour as before. It's only `Write(byte[])` that wouldn't flush.

jscarle:
> Yep that makes sense, but note that `Write(string)` will also have to flush internally to meet the same behaviour as before. It's only `Write(byte[])` that wouldn't flush.

I think that continues to make sense. If you're sending a full string, you probably are done as well. I suspect that people who are calling Write repeatedly will do so with byte[] and not strings.

WojciechNagorski (approved):

Rob-Hague:
Thanks. I will restore the write buffer shortly

WojciechNagorski:
Okej so then we will release the new version

jscarle:
@WojciechNagorski I'd like to submit my PR before you release if you don't mind.

WojciechNagorski:
@jscarle Ok, no problem, just let me know which one.

jscarle:
> @jscarle Ok, no problem, just let me know which one.

#1321

PR #1324, opened 2024-02-16
Drop RIPEMD160

scott-xu:
hmac-ripemd160 was removed in [OpenSSH 7.6](https://www.openssh.com/txt/release-7.6) at 2017-10-03 (both client and server).
With this PR, the [SshNet.Security.Cryptography](https://www.nuget.org/packages/SshNet.Security.Cryptography) nuget dependency can be removed.
Also relates to https://github.com/sshnet/SSH.NET/pull/1109


Rob-Hague (approved):
Thanks

@WojciechNagorski we might want to archive/deprecate the Cryptography repo/package once this makes a release.

WojciechNagorski (approved):
LGTM

PR #1325, opened 2024-02-16
Remove feature binary serialization

scott-xu:
The constant **FEATURE_BINARY_SERIALIZATION** was to help build with .NET Standard 1.3.
Since current all target frameworks support [Serialization](https://learn.microsoft.com/en-us/dotnet/api/system.serializableattribute?view=net-8.0#applies-to), we don't need this flag anymore.
See https://github.com/sshnet/SSH.NET/commit/cf8510013db0e8a1d54086d67694e041f5ef4b98?diff=split&w=0#diff-e4b38dd57f5fc1250fafcccb6ba71f914244bb03755abeea94cbfc0ea4bbbc90L44

branch **scott-xu:remove-FEATURE_BINARY_SERIALIZATION** is based on **scott-xu:drop-RIPEMD160**


scott-xu:
Close this PR based on https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md
Open question: can we remove `[Serialization]` attribute for net462? Code will be more consistency but probably no value added in practice.

Rob-Hague:
I think we should keep it as it is. You can replace `#if FEATURE_BINARY_SERIALIZATION` with `#if NETFRAMEWORK` if you like.

scott-xu:
Good suggestion.

Rob-Hague (approved):

WojciechNagorski (approved):
LGTM

PR #1326, opened 2024-02-17
Add support for Zlib compression (.NET 6.0 onward only)

### LINKED ISSUE #119: Enable ZLib compression with Ionic
What are the lines I need to un-comment to enable Zlib support using the Ionic library?

### LINKED ISSUE #336: Enable compression?
Hello,

How i can enable the compression?

Zlib is supported i guess?

Thanks for advance

### LINKED ISSUE #559: Compression with port forwarding - client drops connection
Hello,
Our aim is to enable compression and use port forwarding.  
We managed to get compression working without any port forwarding. 
However, with port forwarding, the ssh client keeps getting disconnected.

We believe that this is a client side related, since we were able to connect with compression using other clients (while troubleshooting)

Can you please provide any guidance or help ?

This is our code sample on the client side
```
var con = new ConnectionInfo(sshServer, 22, sshUsername, new[] { new PasswordAuthenticationMethod(sshUsername, sshPassword) });
con.CompressionAlgorithms.Add("zlib@openssh.com", typeof(Renci.SshNet.Compression.ZlibOpenSsh));
                client = new SshClient(con);
 client.Connect();
var remotePort = !string.IsNullOrWhiteSpace(dataConnector.PortNumber) ? Convert.ToUInt32(dataConnector.PortNumber) : Common.Constants.DEFAULT_SSH_REMOTE_PORT;
 ForwardedPortLocal forPort = new ForwardedPortLocal(Common.Constants.LOCAL_HOST_IP_ADDRESS, Common.Constants.DEFAULT_SSH_FORWARD_PORT, dataConnector.Server, remotePort);
client.AddForwardedPort(forPort);
forPort.Start();
```
Kindest Regards,

### LINKED ISSUE #1130: Compression support
Even though **SSH.NET** supports compression, we do not ship with compression enabled.
We have a (stub-like) implementation of **zlib** compression in place which consists of the following classes:
* Renci.SshNet.Compression.Compressor
* Renci.SshNet.Compression.Zlib
* Renci.SshNet.Compression.ZlibOpenSsh
* Renci.SshNet.Compression.ZlibStream

The implementation of **ZlibStream** is commented out, and without this compression is of course broken.
I wasn't actively invoked in **SSH.NET** when this was decided, but I suppose this was done to avoid an extra dependency on a **zlib** library.

I see the following options for completing support for **zlib** (de)compression:
|Option|Advantages|Disadvantages|
|:-|:-|-|
|[ZipLibStream](https://learn.microsoft.com/en-us/dotnet/api/system.io.compression.zlibstream)|Supports RFC1950.|Only available in .NET 6+.|
|[DeflateStream](https://learn.microsoft.com/en-us/dotnet/api/system.io.compression.deflatestream)<sup>1</sup>|Supports wide range of TFMs.|Doesn't support RFC1950 header and trailer bytes.|
|Third-party **zlib** library<sup>2</sup>|Possible support wide range of TFMs.|Licensing.|

<sup>1</sup> We'll have to verify whether the lack of support for TFC1950 is a showstopper for [DeflateStream](https://learn.microsoft.com/en-us/dotnet/api/system.io.compression.deflatestream).
<sup>2</sup> If we decide to use a third-party library, we should move the **zlib** compression support to a separate library/assembly/nuget.

**Note:**
The difference between **zlib** and **zlib@openssh.org** is that the latter [delays](https://datatracker.ietf.org/doc/html/draft-miller-secsh-compression-delayed) compression until after user authentication.

scott-xu:
This PR adds support for **zlib<nolink>@openssh.com** described in https://www.openssh.com/txt/draft-miller-secsh-compression-delayed-00.txt

Resolves https://github.com/sshnet/SSH.NET/issues/1130
Resolves https://github.com/sshnet/SSH.NET/issues/559
Resolves https://github.com/sshnet/SSH.NET/issues/336
Resolves https://github.com/sshnet/SSH.NET/issues/119

**Notes:**
1. This PR only adds compression support for .NET 6.0+.
2. "none" is the first and preferred compression method which means no compression by default, just like [ssh_config](https://www.man7.org/linux/man-pages/man5/ssh_config.5.html#:~:text=be%20yes%20or-,no%20(the%20default,-).%0A%0A%20%20%20%20%20%20%20ConnectionAttempts%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20Specifies):
      > Compression
            Specifies whether to use compression.  The argument must
            be yes or no (the default).

      The description of [ssh -C option](https://www.man7.org/linux/man-pages/man1/ssh.1.html#:~:text=Compression%20is%20desirable%20on%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20modem%20lines%20and%20other%20slow%20connections%2C%20but%20will%20only%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20slow%20down%20things%20on%20fast%20networks) explains the reason:
      > Compression is desirable on
         modem lines and other slow connections, but will only
         slow down things on fast networks.
3. **zlib** (pre-auth) is not supported in OpenSSH server (sshd) but is listed in OpenSSH client (ssh).
    Since it is not easy to do integration-test for **zlib** (pre-auth), this PR does **not** add **zlib** (pre-auth).


README.md
```diff
@@ -120,6 +120,14 @@ Private keys can be encrypted using one of the following cipher methods:
 * hmac-sha2-256-etm<span></span>@openssh.com
 * hmac-sha2-512-etm<span></span>@openssh.com
 
+
+## Compression
+
+**SSH.NET** supports the following compression algorithms:
+* none (default)
+* zlib<span></span>@openssh.com (.NET 6 and higher)
```
Rob-Hague, line 128: Is there another way to escape the URL? It doesn't work in the nuget README: https://www.nuget.org/packages/SSH.NET/#readme-body-tab

src/Renci.SshNet/Compression/ZlibStream.cs
Rob-Hague, line 1: Can we just delete this file?

test/Renci.SshNet.IntegrationTests/CompressionTests.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        [TestMethod]
+        public void None()
+        {
+            DoTest(new KeyValuePair<string, Func<Compressor>>("none", null));
+        }
+
+#if NET6_0_OR_GREATER
+        [TestMethod]
+        public void ZlibOpenSsh()
```
Rob-Hague, line 24: Because the compression initialises after user authentication, are we actually testing any compressed packets? Should these tests upload/download a file, like e.g. SftpClientTest.Test_Sftp_Upload_And_Download_1MB_File?

src/Renci.SshNet/Compression/ZlibStream.cs
scott-xu: If we want to support Zlib compression for .NET lower than 6, then we need this facade class. Inside this class, if it is .NET lower than 6, we use 3rd party implementation (Ionic), otherwise use BCL `ZLibStream`.

README.md
scott-xu: Let me try `[zlib@openssh.com]()` -> [zlib@openssh.com]()

scott-xu: try `zlib&#64;openssh.com` -> zlib&#64;openssh.com

scott-xu: try `zlib<nolink>@openssh.com` -> zlib<nolink>@openssh.com

scott-xu: > try `zlib<nolink>@openssh.com` -> zlib<nolink>@openssh.com

probably not work in nuget...

scott-xu:
With this PR, SSH.NET targeted to frameworks below .NET 6.0 can easily use 3rd party library to do the compression. For example: https://github.com/scott-xu/SSH.NET.Ionic.Zlib/blob/main/src/SshNet.IonicZlib/ZLib.cs

scott-xu:
The decompression works when receive SSH_MSG_GLOBAL_REQUEST.
The decompression stops working when receive SSH_MSG_CHANNEL_OPEN_CONFIRMATION (compared with non-compression mode). It tries to decompress byte array: `[20 13 74 63 12 112 208 192 0 16]`. No idea what's going on here. Thoughts? @Rob-Hague

scott-xu:
Ready to review @Rob-Hague @WojciechNagorski

test/Renci.SshNet.IntegrationTests/CompressionTests.cs
```diff
@@ -0,0 +1,64 @@
...
+            DoTest(new KeyValuePair<string, Func<Compressor>>("zlib@openssh.com", () => new ZlibOpenSsh()));
+        }
+#endif
+
+        private void DoTest(KeyValuePair<string, Func<Compressor>> compressor)
+        {
+            using (var scpClient = new ScpClient(_connectionInfoFactory.Create()))
+            {
+                scpClient.ConnectionInfo.CompressionAlgorithms.Clear();
+                scpClient.ConnectionInfo.CompressionAlgorithms.Add(compressor);
```
Rob-Hague, line 35: nit: in e.g. CipherTests, HmacTests, we specify the algorithm on the server and test the default SSH.NET configuration against it, rather than specifying the algorithm on the client.

Rob-Hague (approved):
Nice. Is there a reason for not testing "zlib" as well?

scott-xu:
> Nice. Is there a reason for not testing "zlib" as well?

Because sshd doesn't support "zlib"(pre-auth) anymore.

test/Renci.SshNet.IntegrationTests/CompressionTests.cs
scott-xu: The server can only be configured to compression yes/no. https://www.man7.org/linux/man-pages/man5/sshd_config.5.html

src/Renci.SshNet/ConnectionInfo.cs
```diff
@@ -439,6 +439,10 @@ public ConnectionInfo(string host, int port, string username, ProxyTypes proxyTy
             CompressionAlgorithms = new Dictionary<string, Func<Compressor>>
                 {
                     { "none", null },
+#if NET6_0_OR_GREATER
+                    { "zlib@openssh.com", () => new ZlibOpenSsh() },
+                    { "zlib", () => new Zlib() },
```
Rob-Hague:
```suggestion
```

Rob-Hague (commented):
Ok, in that case I suggest we do not support the "zlib" algorithm since we can't test it easily. What do you think?

This is good to go otherwise.

Rob-Hague:
Thanks, this is good to merge.

scott-xu:
@WojciechNagorski Is there any other thing I can do to get this merged? Thanks!

WojciechNagorski:
Yes. You can ask @Rob-Hague for review. I do not have time now.

Rob-Hague:
I approved it but it needs CODEOWNERS approval:

image

No rush

WojciechNagorski:
I've added @Rob-Hague to code owners.

scott-xu:
Perfect!

PR #1329, opened 2024-02-20
Moved IDE0005 warning configuration.

jscarle:
Moved the IDE0005 warning into the .editorconfig for BouncyCastle and Chaos.NaCl.


jscarle:
@WojciechNagorski @Rob-Hague I moved the IDE0005 in the directories for BouncyCastle and Chaos.NaCl which allows us to not mute it for the entire solution and eliminates the source of the issue, which are these two packages which have been integrated.

WojciechNagorski (approved):
Great idea!

Rob-Hague:
Thanks

PR #1332, opened 2024-02-21
Add `Closed` event to `ShellStream`.

### LINKED ISSUE #44: Channel_Closed in ShellStream
Are you planning to add an event "Closed" in the class "ShellStream"?

Something like this:

public event EventHandler<EventArgs> Closed;

public override void Close()
{
    if (_channel != null)
        _channel.Close();
    base.Close();
}

private void Channel_Closed(object sender, ChannelEventArgs e)
{
    if (Closed != null)
    {
        //  Handle event on different thread
        ExecuteThread(() => Closed(this, new EventArgs()));
    }
}

instead of this:

private void Channel_Closed(object sender, ChannelEventArgs e)
{
    //  TODO:   Do we need to call dispose here ??
    Dispose();
}

scott-xu:
Lib consumer could hook to this event to detect if channel is closed by server **in time**.
Closes https://github.com/sshnet/SSH.NET/issues/44

### DIFF
--- a/src/Renci.SshNet/ShellStream.cs
+++ b/src/Renci.SshNet/ShellStream.cs
@@ -45,6 +45,11 @@ namespace Renci.SshNet
         public event EventHandler<ExceptionEventArgs>? ErrorOccurred;
 
         /// <summary>
+        /// Occurs when the channel was closed.
+        /// </summary>
+        public event EventHandler<EventArgs>? Closed;
+
+        /// <summary>
         /// Gets a value indicating whether data is available on the <see cref="ShellStream"/> to be read.
         /// </summary>
         /// <value>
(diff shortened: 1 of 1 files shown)


jscarle:
Can you please link https://github.com/sshnet/SSH.NET/issues/44 in your description to close it with this PR?

scott-xu:
In our case, for some reason, the server may send channel close message to the client. As the client, we want to know this event **in time** and do some follow up work.
I checked the existing unit tests but no related cases found.

scott-xu:
> Can you please link #44 in your description to close it with this PR?

I should search first. Anyway I'm not alone. Thanks!

Rob-Hague:
What's the use case here?

Now that Read blocks until the channel is closed, can you just wait until Read returns 0?

scott-xu:
> What's the use case here?
> 
> 
> 
> Now that Read blocks until the channel is closed, can you just wait until Read returns 0?

We don't call `Read`. We use `Write` method and `DataRecieved` event.
You can check my previous comment for use case.

WojciechNagorski:
I think we can merge this PR. Fits the current SSH.NET architecture.

WojciechNagorski (approved):

PR #1339, opened 2024-02-24
enable nullable on BaseClient and SftpClient

mus65:
This is just a start to get some feedback, I could do more of the public APIs.

### DIFF
--- a/src/Renci.SshNet/ISftpClient.cs
+++ b/src/Renci.SshNet/ISftpClient.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿#nullable enable
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Text;
@@ -192,7 +193,7 @@ namespace Renci.SshNet
         /// <remarks>
         /// Method calls made by this method to <paramref name="output" />, may under certain conditions result in exceptions thrown by the stream.
         /// </remarks>
-        IAsyncResult BeginDownloadFile(string path, Stream output, AsyncCallback asyncCallback);
+        IAsyncResult BeginDownloadFile(string path, Stream output, AsyncCallback? asyncCallback);
 
         /// <summary>
         /// Begins an asynchronous file downloading into the stream.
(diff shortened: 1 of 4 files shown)


src/Renci.SshNet/ISftpClient.cs
```diff
@@ -211,7 +212,7 @@ public interface ISftpClient : IBaseClient
         /// <remarks>
         /// Method calls made by this method to <paramref name="output" />, may under certain conditions result in exceptions thrown by the stream.
         /// </remarks>
-        IAsyncResult BeginDownloadFile(string path, Stream output, AsyncCallback asyncCallback, object state, Action<ulong> downloadCallback = null);
+        IAsyncResult BeginDownloadFile(string path, Stream output, AsyncCallback asyncCallback, object? state, Action<ulong>? downloadCallback = null);
```
Rob-Hague: Every `AsyncCallback` in this file and in SftpClient.cs can be nullable

```suggestion
        IAsyncResult BeginDownloadFile(string path, Stream output, AsyncCallback? asyncCallback, object? state, Action<ulong>? downloadCallback = null);
```

src/Renci.SshNet/SftpClient.cs
```diff
@@ -2121,6 +2122,11 @@ public IAsyncResult BeginSynchronizeDirectories(string sourcePath, string destin
                 throw new ArgumentException("destDir");
```
Rob-Hague:
```suggestion
                throw new ArgumentException(nameof(destinationPath));
```

If it's not too much trouble while we are here :-)

src/Renci.SshNet/SftpClient.cs
```diff
@@ -1704,7 +1705,7 @@ public string[] ReadAllLines(string path, Encoding encoding)
             {
                 while (!stream.EndOfStream)
                 {
-                    lines.Add(stream.ReadLine());
+                    lines.Add(stream.ReadLine()!);
```
Rob-Hague: This block could be more canonically written as:

```c#
string? line;
while ((line = stream.ReadLine()) != null)
{
    lines.Add(line);
}
```

(github won't let me "add a suggestion")

src/Renci.SshNet/SftpClient.cs
mus65: Strictly speaking this is (already) wrong since ArgumentException expects a message, not a parameter name. I changed it anyway.

ArgumentException.ThrowIfNullOrEmpty would be nice here, but it's only available in .NET 7+.

Rob-Hague (approved):
Thanks

For future reference, it is easier if you can avoid force-pushing once a review has started (merge `develop` into the PR branch to fix conflicts if possible). Then the new changes can more easily be seen

PR #1342, opened 2024-03-04
Handle timeout correctly on Socks5 Proxy

BoronBGP:
Currently when using the Socks5 proxy there are issues with timeouts when a socks5 proxy server takes a long time to respond to a connection request (e.g. when a host is unreachable). This PR adds a timeout to all socket read calls as it standard with the Socks4Connector and a test for this.


WojciechNagorski:
Looks great! I need to test this.

Rob-Hague (approved):
Thanks

Issue #1147, opened 2023-06-26
CancelAsync does not cancel the command started by BeginExecute and blocks execution


togarha:
When calling `CancelAsync` for a process started by `BeginExecute`, the process is not canceled and the result of the command cannot be retrieved.

With enough time for executing the command and without `CancelAsync` everything works fine:
```
            using (SshClient client = new SshClient(host, username, password))
            {
                client.Connect();
                using (SshCommand command = client.CreateCommand("sleep 10; echo run"))
                {
                    IAsyncResult task = command.BeginExecute();
                    Thread.Sleep(15000);
                    //command.CancelAsync();
                    var result = command.EndExecute(task);
                    System.Console.WriteLine(result);
                }
                client.Disconnect();
            }
```
After 15s, the following correctly appears in the console and the execution finishes:
```
run
```

If I reduce the sleep to 5s, and uncomment the `CancelAsync`:
```
            using (SshClient client = new SshClient(host, username, password))
            {
                client.Connect();
                using (SshCommand command = client.CreateCommand("sleep 10; echo run"))
                {
                    IAsyncResult task = command.BeginExecute();
                    Thread.Sleep(5000);
                    command.CancelAsync();
                    var result = command.EndExecute(task);
                    System.Console.WriteLine(result);
                }
                client.Disconnect();
            }
```
The process gets blocked at `EndExecute` instruction, it doesn't finish even though the command hasn't been canceled and it finishes after 10s from the beginning.

Issue #1023, opened 2022-10-28
Is it possible to cancel SshCommand?


MistyKuu:
There is CancelAsync method but it seems to only cancel connection between client and server. Is there a way to stop the command from running? Similarly to CTRL+C while running in BASH shell.

lifeincha0s:
No, using RunCommand, the Execute method blocks until all data is returned from the command, so nothing can be sent to stop the command. You need to use CreateShellStream to kill a running process, which allows writing data while reading data. You can then send Ctrl^C (\x3) to the output stream and kill the process.

bsmithcompsci:
I solved this after a long day trying to force this to cancel, by hijacking the SSHCommand via reflection. This is nasty, please contributors/maintainers make a proper fix for this, this hurts production code. 

# Answer
```cs
token.Register(() => {
  FieldInfo? _sessionErrorOccuredWaitHandleField = sshCommand.GetType().GetField("_sessionErrorOccuredWaitHandle", BindingFlags.NonPublic | BindingFlags.Instance);
  if (_sessionErrorOccuredWaitHandleField != null)
  {
      object? _sessionErrorOccuredWaitHandle = _sessionErrorOccuredWaitHandleField.GetValue(sshCommand);
      MethodInfo[] methods = _sessionErrorOccuredWaitHandleField.FieldType.GetMethods();
      MethodInfo? Set = _sessionErrorOccuredWaitHandleField.FieldType.GetMethod("Set");
      if (Set != null)
          Set.Invoke(_sessionErrorOccuredWaitHandle, new object[] { });
  }
   sshCommand.CancelAsync();
   client.Disconnect();
});
```

# Diagnoses

To start, what's the issue. `EndExecute()` blocks the thread. This as expected to do, but we should be able to have a `CancellationToken` to cancel a asynchronous task. 

Using `CancelAsync()` just calls a dispose.... not greatly helpful. 
![image](https://github.com/sshnet/SSH.NET/assets/87924822/b2fef2fc-e07c-4f35-a886-57c7136c9310)

Okay, let's kill the client `Disconnect()`... nothing.... Well, let's understand how the Client communicates to the SSHCommand; we can see that the ctor of SSHCommand will append delegates for client callbacks.
![image](https://github.com/sshnet/SSH.NET/assets/87924822/3a0c48e2-8f17-4cac-b968-edc7d93346b6)
[...]

PR #1343, opened 2024-03-05
Add ExecuteAsync, Fix CancelAsync Deadlock

zeotuan:
- Implemented TAP over APM ExecuteAsync(CancellationToken) mentiond in https://github.com/sshnet/SSH.NET/issues/931
- Fix CancelAsync causing deadlocking and it now send SIGTERM signal to try to abort the request which will probably also solve https://github.com/sshnet/SSH.NET/issues/1147 and https://github.com/sshnet/SSH.NET/issues/1023


src/Renci.SshNet/SshCommand.cs
```diff
@@ -348,21 +367,109 @@ public string EndExecute(IAsyncResult asyncResult)
...
+        {
+#if NET || NETSTANDARD2_1_OR_GREATER
+            await using var ctr = cancellationToken.Register(() => CancelAsync(forceKill), useSynchronizationContext: false).ConfigureAwait(continueOnCapturedContext: false);
+#else
+            using var ctr = cancellationToken.Register(() => CancelAsync(forceKill), useSynchronizationContext: false);
+#endif // NET || NETSTANDARD2_1_OR_GREATER
+
+            try
+            {
+                var status = await Task<int>.Factory.FromAsync(BeginExecute(), EndExecuteWithStatus).ConfigureAwait(false);
```
WojciechNagorski, line 453: I want to add `Async` overloaded methods but I think this is not a real `async` code. I've prepared a POC here: https://github.com/sshnet/SSH.NET/pull/1318 

I know that your approach is easier to implement, but this is similar to nuget: https://www.nuget.org/packages/Renci.SshNet.Async#versions-body-tab

You can find code here: https://github.com/JohnTheGr8/Renci.SshNet.Async/blob/master/Renci.SshNet.Async/SshNetExtensions.cs

src/Renci.SshNet/SshCommand.cs
```diff
@@ -105,56 +107,72 @@ public Stream CreateInputStream()
         /// <summary>
         /// Gets the command execution result.
         /// </summary>
+#pragma warning disable S1133 // Deprecated code should be removed
+        [Obsolete("Please read the result from the OutputStream. I.e. new StreamReader(shell.OutputStream).ReadToEnd().")]
```
WojciechNagorski, line 111: Why should we make it obsolete?

src/Renci.SshNet/SshCommand.cs
```diff
@@ -105,56 +107,72 @@ public Stream CreateInputStream()
...
             }
+
+            return _result.ToString();
         }
 
         /// <summary>
         /// Gets the command execution error.
         /// </summary>
+#pragma warning disable S1133 // Deprecated code should be removed
+        [Obsolete("Please read the error result from the ExtendedOutputStream. I.e. new StreamReader(shell.ExtendedOutputStream).ReadToEnd().")]
```
WojciechNagorski, line 144: Why should we make it obsolete?

Rob-Hague:
@WojciechNagorski commented just as I was commenting some similar feedback, so sorry for the duplication 🙂 :

I think you are still working on this so just some high-level thoughts:

There are a few semi-related changes here. The cancellation via signal is good. I would suggest splitting that to a standalone PR to start with.

I understand the reason for obsoleting the Result/Error properties (because they consume the stream properties) but I do not think the change is justified. Especially given how many changes it causes just in our own code (forwarding to a now internal version at that). I would suggest reverting that.

I'm a little uneasy about using `TaskFactory.FromAsync` in this library because many of the Begin/End (APM) methods are not implemented so well and we end up spawning new threads/blocking on more threads than we need. I basically share the same sentiment as https://github.com/sshnet/SSH.NET/issues/931#issuecomment-1066160452. But at first glance it is not *so* bad here.

zeotuan:
@Rob-Hague @WojciechNagorski Very Interesting. I would love to help with the project.

https://www.nuget.org/packages/Renci.SshNet.Async#versions-body-tab unfortunately does not support ```CancellationToken```. 
So I was thinking of adding a simple TAP over APM wrapper here until we get a proper TAP rewrite. 

I also expect the execution of command to leave the ```OutputStream``` and ```ExtendedOutputStream``` intact and provide my own way to consuming them. for example: I want to forward the stdout and stderr to another program while the command is being executed not after.

Therefore, I would suggest Future ExecuteAsync Implementation to at least just return an exit status code instead of consuming those stream.  Also Providing a public helper method to let people get Result/Error from those stream. It would require refactoring for any downstream project that will migrate to Async anyway

I have submitted a standalone PR for CancelAsync: https://github.com/sshnet/SSH.NET/pull/1345

WojciechNagorski:
@zeotuan Sory for the delay. If you want you can continue to help us add true asynchronous methods. let us know.

zeotuan:
@WojciechNagorski Hi! Yes, I would like to help with adding true async methods for sshnet.

PR #1345, opened 2024-03-07
Fix CancelAsync Cause Deadlock

### LINKED ISSUE #1023: Is it possible to cancel SshCommand?

### LINKED ISSUE #1147: CancelAsync does not cancel the command started by BeginExecute and blocks execution

zeotuan:
Fix CancelAsync causing deadlocking:

it now signal WaitHandle of Cancelled event after sending SIGTERM/SIGKILL signal to try to abort the request.

close https://github.com/sshnet/SSH.NET/issues/1147 and close https://github.com/sshnet/SSH.NET/issues/1023


test/Renci.SshNet.IntegrationTests/OldIntegrationTests/SshCommandTest.cs
```diff
@@ -51,6 +51,30 @@ public void Test_Execute_SingleCommand()
...
+            using var cmd = client.CreateCommand(command);
+            try
+            {
+                var asyncResult = cmd.BeginExecute();
+                cmd.CancelAsync();
+                cmd.EndExecute(asyncResult);
+            }
+            catch (OperationCanceledException)
+            {
+            }
```
Rob-Hague: It cannot currently throw `OperationCanceledException` but perhaps it should?

```suggestion
            var asyncResult = cmd.BeginExecute();
            cmd.CancelAsync();
            Assert.ThrowsException<OperationCanceledException>(() => cmd.EndExecute(asyncResult));
```

test/Renci.SshNet.IntegrationTests/OldIntegrationTests/SshCommandTest.cs
```diff
@@ -51,6 +51,30 @@ public void Test_Execute_SingleCommand()
...
+            client.Disconnect();
+            Assert.AreNotEqual(cmd.Result.Trim(), testValue);
```
Rob-Hague:
```suggestion
            Assert.AreEqual("", cmd.Result);
```

? (if not, the arguments should be reversed at least)

src/Renci.SshNet/SshCommand.cs
```diff
@@ -356,13 +357,12 @@ public string EndExecute(IAsyncResult asyncResult)
...
+        public void CancelAsync(bool forceKill = false)
         {
-            if (_channel is not null && _channel.IsOpen && _asyncResult is not null)
-            {
-                // TODO: check with Oleg if we shouldn't dispose the channel and uninitialize it ?
-                _channel.Dispose();
-            }
+            var signal = forceKill ? "KILL" : "TERM";
+            _ = _channel?.SendExitSignalRequest(signal, coreDumped: false, "Command execution has been cancelled.", "en");
+            _ = _commandCancelledWaitHandle.Set();
```
Rob-Hague: I think we should set `_asyncResult.IsCompleted = true` here as well (before this line)

src/Renci.SshNet/SshCommand.cs
```diff
@@ -356,13 +357,12 @@ public string EndExecute(IAsyncResult asyncResult)
         /// <summary>
         /// Cancels command execution in asynchronous scenarios.
         /// </summary>
-        public void CancelAsync()
+        /// <param name="forceKill">if true send SIGKILL instead of SIGTERM.</param>
+        public void CancelAsync(bool forceKill = false)
```
Rob-Hague, line 368: Perhaps we drop the `Async` suffix here (and obsolete `CancelAsync`) as it is not an async method. @WojciechNagorski?

Rob-Hague (commented):
Would it be possible to add a test for forceKill = true as well?

src/Renci.SshNet/SshCommand.cs
zeotuan: Sure. I am doing some further debugging. It seem that we might have to also execute the ```_callback``` as well as Flushing the ```OutputStream``` + ```ExntednedOutputStream``` .

On my local testing, ```Channel.Close``` event is never called and when I try to make ```SendExitSignalRequest``` wait for a response, I got ```Session Operation has Timeout``` Exception. Wondering if anyone else can recreate it.

Rob-Hague: Is it because of the call to `UnsubscribeFromEventsAndDisposeChannel(_channel);` in `EndExecute`?

Perhaps `_commandCancelledWaitHandle` should be removed? If I read https://datatracker.ietf.org/doc/html/rfc4254#section-6.10 correctly, we should always get SSH_MSG_CHANNEL_CLOSE after the signal, so it "should" just work.

edit: actually, not sure I did read it correctly

zeotuan: You are right! A proper implementation is to just send SIGTERM/KILL and wait for a response.

However, when I tried removing the _commandCancelledWaitHandle and made SendExitSignalRequest wait for a response, it never happened. If I just make SendExitSignalRequest send the message without waiting for a response, it will be blocked until the command actually finishes. 

My server openSSH version: OpenSSH_8.2p1 Ubuntu-4ubuntu0.11 so those signals are probably already supported.
I will takes a deeper look at the log.

src/Renci.SshNet/SshCommand.cs
```diff
@@ -349,20 +352,49 @@ public string EndExecute(IAsyncResult asyncResult)
...
             }
+
+            return Task.Run(() =>
+                         {
+                             var signaledElement = WaitHandle.WaitAny(new[] { _asyncResult.AsyncWaitHandle }, timeout);
+                             if (signaledElement == WaitHandle.WaitTimeout)
+                             {
+                                 _ = _commandCancelledWaitHandle?.Set();
+                             }
+                         });
```
Rob-Hague: I think the version at 43a57e12 is better. Spawning a thread to block on a wait is not ideal.

I am thinking we should not try and wait for a response after all (I'm not convinced there is one now).

As for the callback, I am not sure but I think if we have cancelled the operation, we should not execute the callback. (if one cancels a Task-based operation with a CancellationToken, then we get a TaskCanceledException and continuations are not executed)

src/Renci.SshNet/Common/SshOperationCancelledException.cs
Rob-Hague, line 1: Can we just use [System.OperationCanceledException](https://learn.microsoft.com/en-us/dotnet/api/system.operationcanceledexception)?

Rob-Hague (approved):

WojciechNagorski (approved):
Unfortunately, I don't have much time for this project at the moment. Only PRs that have @Rob-Hague's approval are verified and merged by me.

PR #1362, opened 2024-03-23
SftpClient: handle the SFTP session being closed by the server

### LINKED ISSUE #369: System.InvalidOperationException: The session is not open.
Hi 
I am running on renci latest version 2016.1.0.0 

Trying to upload/poll/download and delete in sequence using 4 different static SFTP connections for these individual operations. This all happens in multi-threaded environment with a throttle of 100 parallel threads. 

The dispose happens at once, only when all the threads are finished. 

Problem : After a few files I start getting the following error in either of these operations (upload/poll/download/delete). 
```

_SFTP - Error in SFTP Upload. Retry attempt number : 9. Exception : System.InvalidOperationException: The session is not open.
   at Renci.SshNet.SubsystemSession.EnsureSessionIsOpen()
   at Renci.SshNet.SubsystemSession.SendData(Byte[] data)
   at Renci.SshNet.Sftp.SftpSession.SendRequest(SftpRequest request)
   at Renci.SshNet.Sftp.SftpSession.RequestRealPath(String path, Boolean nullOnError)
   at Renci.SshNet.Sftp.SftpSession.GetCanonicalPath(String path)
   at Renci.SshNet.SftpClient.InternalUploadFile(Stream input, String path, Flags flags, SftpUploadAsyncResult asyncResult, Action`1 uploadCallback)
   at Renci.SshNet.SftpClient.UploadFile(Stream input, String path, Action`1 uploadCallback)_
```

**Following is the GetInstance method to create the 4 static connection on first request:**
```
public SftpClient GetInstance(string instanceID)
        {


            if (_edmSftpClient == null || _edmSftpClient.ContainsKey(instanceID) == false)
            {
                lock (((ICollection)_edmSftpClient).SyncRoot)
                {
                    if (_edmSftpClient == null || _edmSftpClient.ContainsKey(instanceID) == false)
                    {
                        string _domain = _credential.Domain.Replace("sftp://", "").Replace("ftp://", "");
         
                        ConnectionInfo ConnNfo = new ConnectionInfo(_domain, 22, _credential.UserName, new AuthenticationMethod[]
                            {
                                new PasswordAuthenticationMethod(_credential.UserName, _credential.Password),                             
                                new PrivateKeyAuthenticationMethod(_credential.UserName,_pkFileClient.ToArray()),
                                    }
                                );

                        logger.Info(string.Format("Creating SFTP for blockInstance: {0}", instanceID));
                        if (_credential.Domain != null)
                            _edmSftpClient.Add(instanceID, new SftpClient(ConnNfo) { OperationTimeout = TimeSpan.FromDays(1), KeepAliveInterval = TimeSpan.FromDays(1) });
                        else
                            throw new Exception("SFTP Domain not set.");
                    }
                }
            }
            if (_edmSftpClient[instanceID].IsConnected == false)
            {
                lock (((ICollection)_edmSftpClient).SyncRoot)
                {
                    if (_edmSftpClient[instanceID].IsConnected == false)
                    {
                        _edmSftpClient[instanceID].ConnectionInfo.MaxSessions = 99999999;
                        _edmSftpClient[instanceID].ConnectionInfo.Timeout = TimeSpan.FromDays(1);
                        _edmSftpClient[instanceID].Connect();
                    }
                }
            }
            _edmSftpClient[instanceID].ErrorOccurred += (object sender, ExceptionEventArgs e) =>
            {
                logger.Error(string.Format("_edmSftpClient.ErrorOccurred event captured with exception {0}", e != null && e.Exception != null ? e.Exception.ToString() : "no error message recorded"));
            };
            return _edmSftpClient[instanceID];
        }
```

**And here is the sample Upload method that I have written. Trying for 10 retries:**
 ```
public void Upload(string remotePath, string remoteFileName, byte[] stream)
        {
            EDMSftpClient _sftpClient = null;
            int _retryCount = 10;
            while (true)
            {
                try
                {
                    logger.Info(string.Format("Uploading File : {0}. RemotePath : {1}", remoteFileName, remotePath));

                    _sftpClient = new EDMSftpClient(Credential);
                    _sftpClient._pkFileClient = pkFile;
                    string r = GetRemotePathNoServer(remotePath, remoteFileName);
                    var _uInstance = _sftpClient.GetInstance("U" + _blockInstanceID);
                    _uInstance.UploadFile(new MemoryStream(stream), r, (x) =>
                    {
                        logger.Info(string.Format("Uploaded File : {0}. RemotePath : {1}", remoteFileName, remotePath));
                    });
                    break;
                }
                catch (Exception ex)
                {
                    _retryCount--;
                    logger.Error(string.Format("Error in SFTP Upload. Retry attempt number : {0}. Exception : {1}", _retryCount, ex));
                    if (_retryCount == 0)
                        throw;
                    Thread.Sleep(1000 * _retryCount);

                }
            }
        }
```

This happens intermittently after some time. Have tried playing around with properties like ConnectionInfo.MaxSessions, KeepAliveInterval, ConnectionInfo.Timeout. But no luck. 
And the confusion is the the _edmSftpClient[instanceID].IsConnected property evaluates to true while I get this error.

Can you please suggest where am I going wrong. Could it be SFTP server setting issue or a code bug.

### LINKED ISSUE #843: Discussion: How to reliably detect if SftpClient is still connected
I need to implement SFTP connection pooling. As part of this I need to have long running ```SftpClient```s and a way to check if the client is still connected. Initially I used ```SftpClient.IsConnected``` for this purpose. However, even if ```SftpClient.IsConnected``` returned true, I many times got this exception when trying to use the ```SftpClient``` instance:
```
System.InvalidOperationException: The session is not open.
   at Renci.SshNet.SubsystemSession.EnsureSessionIsOpen()
   at Renci.SshNet.SubsystemSession.SendData(Byte[] data)
   at Renci.SshNet.Sftp.SftpSession.SendMessage(SftpMessage sftpMessage)
   at Renci.SshNet.Sftp.SftpSession.SendRequest(SftpRequest request)
...
```

Here is the class diagram of the relevant portion of SSH.NET: https://gitmind.com/app/flowchart/75ace876ac90aec4dbce3b61347c6cf9

Based on this architecture, it seems to me that the described problem is caused by ```SftpClient``` not taking into account the ```SubsystemSession```'s state. As an intermediate workaround I made ```BaseClient.IsConnected``` virtual and implemented it in ```SftpClient``` as:
```C#
        public override bool IsConnected
        {
            get
            {
                return base.IsConnected && _sftpSession.IsOpen;
            }
        }
```
Initial tests show this solves my problem, but further testing is needed to really claim this. 

Can someone with more insight into the codebase confirm my reasoning so far? I would also be thankful if the one could answer the questions that pop up in my mind looking at this class diagram:
 - There are a lot of classes named *Session in this diagram... can someone explain what these classes do conceptually? Is the naming good?
 - Is there a reason for Session and SubsystemSession to hold their own respective status? Does/should a change in one's state relate to a change in the other's state?
 - Why do BaseClient and Session each hold ConnectionInfo? Are these two the same?

### LINKED ISSUE #1153: SftpClient.IsConnect returns true even if session is closed.
This is similar to the discussion in https://github.com/sshnet/SSH.NET/issues/843

We have a connection pool implemented for the SftpClient and we reconnect whenever SftpClient.IsConnected() returns false.
However we have noticed issues where SftpClient.IsConnected() still returns true in case the server has closed the session and we eventually get the following error when using the client for any operation
```
System.InvalidOperationException: The session is not open. at Renci.SshNet.Common.AsyncResult.EndInvoke() at Renci.SshNet.SftpClient.EndListDirectory(IAsyncResult asyncResult)
```
Is this an expected behaviour? If not, can we try to implement the fix mentioned #843 or expose a method to determine if the session is open or not?

mus65:
If the server closes the SFTP session but keeps the TCP connection open, this currently causes IsConnected to return true, but any operation fails with "the session is not open".

SftpClient.IsConnected now also check sftpSession.IsOpen. Connect() and ConnectAsync() were reworked to take into account that the Session may already/still be open, but the SFTP session may not. This is needed so a reconnect works.

fixes #843 
fixes #1153
fixes #369

### DIFF
--- a/src/Renci.SshNet/BaseClient.cs
+++ b/src/Renci.SshNet/BaseClient.cs
@@ -237,13 +237,9 @@ namespace Renci.SshNet
 
             // The session may already/still be connected here because e.g. in SftpClient, IsConnected also checks the internal SFTP session
             var session = Session;
-            if (session is null)
+            if (session is null || !session.IsConnected)
             {
                 Session = CreateAndConnectSession();
-            }
-            else if (!session.IsConnected)
-            {
-                session.Connect();
             }
 
             try
@@ -305,13 +301,9 @@ namespace Renci.SshNet
 
             // The session may already/still be connected here because e.g. in SftpClient, IsConnected also checks the internal SFTP session
             var session = Session;
-            if (session is null)
+            if (session is null || !session.IsConnected)
             {
                 Session = await CreateAndConnectSessionAsync(cancellationToken).ConfigureAwait(false);
-            }
-            else if (!session.IsConnected)
-            {
-                await session.ConnectAsync(cancellationToken).ConfigureAwait(false);
             }
 
             try
--- a/src/Renci.SshNet/BaseClient.cs
+++ b/src/Renci.SshNet/BaseClient.cs
@@ -72,7 +72,7 @@ namespace Renci.SshNet
         /// <see langword="true"/> if this client is connected; otherwise, <see langword="false"/>.
         /// </value>
         /// <exception cref="ObjectDisposedException">The method was called after the client was disposed.</exception>
-        public bool IsConnected
+        public virtual bool IsConnected
         {
             get
             {
@@ -228,14 +228,23 @@ namespace Renci.SshNet
             // forwarded port with a client instead of with a session
             //
             // To be discussed with Oleg (or whoever is interested)
-            if (IsSessionConnected())
+            if (IsConnected)
             {
                 throw new InvalidOperationException("The client is already connected.");
             }
 
             OnConnecting();
 
-            Session = CreateAndConnectSession();
+            // The session may already/still be connected here because e.g. in SftpClient, IsConnected also checks the internal SFTP session
+            var session = Session;
+            if (session is null)
+            {
+                Session = CreateAndConnectSession();
+            }
+            else if (!session.IsConnected)
+            {
+                session.Connect();
+            }
 
             try
             {
@@ -287,14 +296,23 @@ namespace Renci.SshNet
             // forwarded port with a client instead of with a session
             //
             // To be discussed with Oleg (or whoever is interested)
-            if (IsSessionConnected())
+            if (IsConnected)
             {
                 throw new InvalidOperationException("The client is already connected.");
             }
 
             OnConnecting();
 
-            Session = await CreateAndConnectSessionAsync(cancellationToken).ConfigureAwait(false);
+            // The session may already/still be connected here because e.g. in SftpClient, IsConnected also checks the internal SFTP session
+            var session = Session;
+            if (session is null)
+            {
+                Session = await CreateAndConnectSessionAsync(cancellationToken).ConfigureAwait(false);
+            }
+            else if (!session.IsConnected)
+            {
+                await session.ConnectAsync(cancellationToken).ConfigureAwait(false);
+            }
 
             try
             {
--- a/src/Renci.SshNet/SftpClient.cs
+++ b/src/Renci.SshNet/SftpClient.cs
@@ -109,6 +109,22 @@ namespace Renci.SshNet
         }
 
         /// <summary>
+        /// Gets a value indicating whether this client is connected to the server and
+        /// the SFTP session is open.
+        /// </summary>
+        /// <value>
+        /// <see langword="true"/> if this client is connected and the SFTP session is open; otherwise, <see langword="false"/>.
+        /// </value>
+        /// <exception cref="ObjectDisposedException">The method was called after the client was disposed.</exception>
+        public override bool IsConnected
+        {
+            get
+            {
+                return base.IsConnected && _sftpSession.IsOpen;
+            }
+        }
+
+        /// <summary>
         /// Gets remote working directory.
         /// </summary>
         /// <exception cref="SshConnectionException">Client is not connected.</exception>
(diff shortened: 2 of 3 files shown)


src/Renci.SshNet/BaseClient.cs
```diff
@@ -228,14 +228,23 @@ public void Connect()
...
+            // The session may already/still be connected here because e.g. in SftpClient, IsConnected also checks the internal SFTP session
+            var session = Session;
+            if (session is null)
+            {
+                Session = CreateAndConnectSession();
+            }
+            else if (!session.IsConnected)
+            {
+                session.Connect();
+            }
```
Rob-Hague: Previously, if `Session.IsConnected` were false, we would create a new Session. I think we should keep that the same.

```suggestion
            var session = Session;
            if (session is null || !session.IsConnected)
            {
                Session = CreateAndConnectSession();
            }
```

src/Renci.SshNet/BaseClient.cs
```diff
@@ -287,14 +296,23 @@ public async Task ConnectAsync(CancellationToken cancellationToken)
...
+            // The session may already/still be connected here because e.g. in SftpClient, IsConnected also checks the internal SFTP session
+            var session = Session;
+            if (session is null)
+            {
+                Session = await CreateAndConnectSessionAsync(cancellationToken).ConfigureAwait(false);
+            }
+            else if (!session.IsConnected)
+            {
+                await session.ConnectAsync(cancellationToken).ConfigureAwait(false);
+            }
```
Rob-Hague:
```suggestion
            var session = Session;
            if (session is null || !session.IsConnected)
            {
                Session = await CreateAndConnectSessionAsync(cancellationToken).ConfigureAwait(false);
            }
```

Rob-Hague:
I tagged @IgorMilavec as co-author for the initial/partial fix.

src/Renci.SshNet/BaseClient.cs
mus65: fixed

src/Renci.SshNet/BaseClient.cs
mus65: fixed

Rob-Hague (approved):
Thanks

WojciechNagorski (approved):

PR #1363, opened 2024-03-24
Handle unknown channel messages correctly

mus65:
See discussion #1218 . Some servers send custom channel messages like 'keepalive@proftpd.org' as keep alive messages. This currently causes a NotSupportedException.

According to the spec https://datatracker.ietf.org/doc/html/rfc4254#section-5.4 :

"If the request is not recognized or is not
supported for the channel, SSH_MSG_CHANNEL_FAILURE is returned."

Send a failure message back instead of throwing an exception.

The  test is mostly copy&paste from ChannelTest_OnSessionChannelRequestReceived_OnRequest_Exception. I also reproduced the issue and tested the fix with an actual ProFTPD server locally.

### DIFF
--- a/src/Renci.SshNet/Channels/Channel.cs
+++ b/src/Renci.SshNet/Channels/Channel.cs
@@ -714,8 +714,14 @@ namespace Renci.SshNet.Channels
                     }
                     else
                     {
-                        var reply = new ChannelFailureMessage(LocalChannelNumber);
-                        SendMessage(reply);
+                        var unknownRequestInfo = new UnknownRequestInfo(e.Message.RequestName);
+                        unknownRequestInfo.Load(e.Message.RequestData);
+
+                        if (unknownRequestInfo.WantReply)
+                        {
+                            var reply = new ChannelFailureMessage(LocalChannelNumber);
+                            SendMessage(reply);
+                        }
                     }
                 }
                 catch (Exception ex)
--- a/src/Renci.SshNet/Channels/Channel.cs
+++ b/src/Renci.SshNet/Channels/Channel.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Globalization;
 using System.Net.Sockets;
 using System.Threading;
 
@@ -715,8 +714,8 @@ namespace Renci.SshNet.Channels
                     }
                     else
                     {
-                        // TODO: we should also send a SSH_MSG_CHANNEL_FAILURE message
-                        throw new NotSupportedException(string.Format(CultureInfo.CurrentCulture, "Request '{0}' is not supported.", e.Message.RequestName));
+                        var reply = new ChannelFailureMessage(LocalChannelNumber);
+                        SendMessage(reply);
                     }
                 }
                 catch (Exception ex)
--- a/src/Renci.SshNet/Channels/Channel.cs
+++ b/src/Renci.SshNet/Channels/Channel.cs
@@ -719,7 +719,7 @@ namespace Renci.SshNet.Channels
 
                         if (unknownRequestInfo.WantReply)
                         {
-                            var reply = new ChannelFailureMessage(LocalChannelNumber);
+                            var reply = new ChannelFailureMessage(RemoteChannelNumber);
                             SendMessage(reply);
                         }
                     }
--- a/test/Renci.SshNet.Tests/Classes/Channels/ChannelTest_OnSessionChannelRequestReceived_HandleUnknownMessage.cs
+++ b/test/Renci.SshNet.Tests/Classes/Channels/ChannelTest_OnSessionChannelRequestReceived_HandleUnknownMessage.cs
@@ -17,6 +17,9 @@ namespace Renci.SshNet.Tests.Classes.Channels
         private uint _localWindowSize;
         private uint _localPacketSize;
         private uint _localChannelNumber;
+        private uint _remoteChannelNumber;
+        private uint _remoteWindowSize;
+        private uint _remotePacketSize;
         private ChannelStub _channel;
         private IList<ExceptionEventArgs> _channelExceptionRegister;
         private UnknownRequestInfoWithWantReply _requestInfo;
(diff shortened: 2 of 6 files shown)


test/Renci.SshNet.Tests/Classes/Channels/ChannelTest_OnSessionChannelRequestReceived_HandleUnknownMessage.cs
```diff
@@ -0,0 +1,70 @@
...
+        public override string RequestName
+        {
+            get
+            {
+                return nameof(UnknownRequestInfo);
+            }
+        }
+
+    }
+}
```
mus65, line 70: This test is mostly copy&paste from ChannelTest_OnSessionChannelRequestReceived_OnRequest_Exception. I also reproduced the issue and tested the fix with an actual ProFTPD server locally.

Rob-Hague:
The whole paragraph is (emphasis mine):

```
   If 'want reply' is FALSE, no response will be sent to the request.
   **Otherwise**, the recipient responds with either
   SSH_MSG_CHANNEL_SUCCESS, SSH_MSG_CHANNEL_FAILURE, or request-specific
   continuation messages.  If the request is not recognized or is not
   supported for the channel, SSH_MSG_CHANNEL_FAILURE is returned.
```

To me, that says we should only send SSH_MSG_CHANNEL_FAILURE if 'want reply' is true.

i.e.

```diff

                         // Raise request specific event
                         OnRequest(requestInfo);
                     }
-                    else
+                    else if (e.Message.WantReply)
                     {
                         var reply = new ChannelFailureMessage(LocalChannelNumber);
                         SendMessage(reply);

```

Unfortunately, `WantReply` does not exist on `ChannelRequestMessage`, but instead on `RequestInfo` which is a wrapper for the "type-specific data" in the SSH_MSG_CHANNEL_REQUEST. I guess this was done because 'want reply' often depends on the type of the request when sending one. I see two options:

1. Plumb `WantReply` onto `ChannelRequestMessage` instead of `RequestInfo`
2. Add your `UnknownRequestInfo` into the production code in order to load the 'want reply' value; or
2.2 Un-abstract `RequestInfo` for the same purpose

2 is probably easiest. What do you think?

mus65:
> 2\. Add your `UnknownRequestInfo` into the production code in order to load the 'want reply' value;

@Rob-Hague Pushed. Moved UnknownRequestInfo to production code and it checks WantReply now. The test still needs its own class to set WantReply=true .

zybexXL:
This RFC clarification (draft) may be relevant here, though it's a slightly different issue arising from the same RFC paragraph:
http://www.watersprings.org/pub/id/draft-sgtatham-secsh-closure-race-00.html

This issue has been discussed before in other SSH packages:
https://libssh2.org/mail/libssh2-devel-archive-2012-01/0000.shtml

And this implementation also agrees with your interpretation of WantReply:
https://github.com/hudson/ganymed-ssh-2/blob/master/src/main/java/ch/ethz/ssh2/channel/ChannelManager.java#L1315

Rob-Hague:
> This RFC clarification (draft) may be relevant here

Thanks. The relevant text is:

```
Specifically, users of [RFC4254] should replace the first sentence of the third paragraph of section 5.4:

If 'want reply' is FALSE, no response will be sent to the request.

with the text:

If 'want reply' is FALSE, or if the sender has already sent SSH_MSG_CHANNEL_CLOSE for the channel, then the sender MUST NOT send a reply.
```

I think we are covered here because `SendMessage` has a check on `IsOpen` which is false when we have sent SSH_MSG_CHANNEL_CLOSE:

https://github.com/sshnet/SSH.NET/blob/db3d7e8d03d0a852cd7d16e1dee95a72939de317/src/Renci.SshNet/Channels/Channel.cs#L479-L488

https://github.com/sshnet/SSH.NET/blob/db3d7e8d03d0a852cd7d16e1dee95a72939de317/src/Renci.SshNet/Channels/Channel.cs#L547-L575

src/Renci.SshNet/Channels/Channel.cs
```diff
@@ -715,8 +714,14 @@ private void OnChannelRequest(object sender, MessageEventArgs<ChannelRequestMess
...
                     else
                     {
-                        // TODO: we should also send a SSH_MSG_CHANNEL_FAILURE message
-                        throw new NotSupportedException(string.Format(CultureInfo.CurrentCulture, "Request '{0}' is not supported.", e.Message.RequestName));
+                        var unknownRequestInfo = new UnknownRequestInfo(e.Message.RequestName);
+                        unknownRequestInfo.Load(e.Message.RequestData);
+
+                        if (unknownRequestInfo.WantReply)
+                        {
+                            var reply = new ChannelFailureMessage(LocalChannelNumber);
```
Rob-Hague: I think this should be:

```suggestion
                            var reply = new ChannelFailureMessage(RemoteChannelNumber);
```

Unfortunately, it seems like these lines are wrong in SshCommand (469 and 477):

https://github.com/sshnet/SSH.NET/blob/db3d7e8d03d0a852cd7d16e1dee95a72939de317/src/Renci.SshNet/SshCommand.cs#L461-L481

mus65: Fixed it in Channel.cs . Should I fix SshCommand.cs as well? (Seems unrelated to this issue to me).

Rob-Hague: Thanks, if it is a one-line fix, it would be good to fix it here. If it is more effort, then you can leave it.

mus65: Pushed it (had to extend IChannel interface with RemoteChannelNumber).

Rob-Hague (approved):
Thanks

WojciechNagorski (approved):

anandgmenon:
@Rob-Hague/ @WojciechNagorski Can we have a release with this fix please?

Rob-Hague:
Probably some time in May

anandgmenon:
@Rob-Hague can we please have a release for this?

Rob-Hague:
Yes, soon

anandgmenon:
Hey @Rob-Hague can you provide a tentative ETA on when you're planning to do the next release?

PR #1364, opened 2024-03-27
Support AesGcm cipher

### LINKED ISSUE #477: AES-GCM
Any chance you could add support in the Encryptions for CipherInfo of AES-GCM's?   I have seen amazing performance boosts with my testing of this cipher using FileZilla Client as a test bench.   Thanks for the consideration.

### LINKED ISSUE #555: AES-GCM Cipher Mode
Hoping someone out there has added this, really want to get this implemented but having a rough time and no luck getting it working

### LINKED ISSUE #773: Support aes256-gcm encryption ?
Is it possible to implement aes256-gcm Encryption Method ?

### LINKED ISSUE #774: is it possible to add AES 128-GCM Cipher Suite or ssh agent forwarding utility?
I could not find any of these two in ssh.net 

 AES 128-GCM Cipher Suite or ssh agent forwarding utility

Any one knows how to add them ?

### LINKED ISSUE #792: AES-GCM 128 into SSH-NET using MS AES-GCM class
I have integrated AES-GCM 128 into SSH-NET using MS AES-GCM class under netstandard 2.1, I have tested it and it's fully functioning, wonder if I can share my changes back to the community so that the community will maintain my changes for the near future

### LINKED ISSUE #994: Request to verify it with TLS 1.3 ciphers and check on Renci.SshNet.dll
Helo Team,

RHEL 8.6 is not supporting the SSH tunnel via Renci.SshNet.dll, possibly because of the latest TLS 1.3 ciphers

This happens when old communication is via SSL and TLS 1.1 is disabled

Is it possible to update with the latest ciphers and rebuilt the project?

TLS v1.3 supports 5 cipher suites.
– TLS_AES_128_GCM_SHA256
– TLS_AES_256_GCM_SHA384
– TLS_CHACHA20_POLY1305_SHA256
– TLS_AES_128_CCM_SHA256
– TLS_AES_128_CCM_8_SHA256

current dll version I am using is:

image

### LINKED ISSUE #1356: we have a requirement to enable the following ciphers (aes128-gcm@openssh.com, aes256-gcm@openssh.com, chacha20-poly1305@openssh.com)
we have a requirement to enable the following ciphers (aes128-gcm@openssh.com, aes256-gcm@openssh.com, chacha20-poly1305@openssh.com). Let me know how we can accommodate this request.

scott-xu:
This PR adds support for aes128-gcm<nolink>@openssh.com, aes256-gcm<nolink>@openssh.com ciphers.
This PR does not add chacha20-poly1305<nolink>@openssh.com cipher.
Fix https://github.com/sshnet/SSH.NET/issues/792 Fix https://github.com/sshnet/SSH.NET/issues/1356 Fix https://github.com/sshnet/SSH.NET/issues/774 Fix https://github.com/sshnet/SSH.NET/issues/773 Fix https://github.com/sshnet/SSH.NET/issues/555 Fix https://github.com/sshnet/SSH.NET/issues/477 Fix https://github.com/sshnet/SSH.NET/issues/994


src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.cs
```diff
@@ -0,0 +1,105 @@
...
+        {
+            // [outbound sequence][packet length field][padding length field sz][payload][random paddings]
+            // [--4 bytes(offset)][------4 bytes------][-------------------Plain Text--------------------]
+            var packetLengthField = new ReadOnlySpan<byte>(input, offset, 4);
+            var plainText = new ReadOnlySpan<byte>(input, offset + 4, length - 4);
+
+            var cipherText = new byte[length - 4];
+            var tag = new byte[TagSize];
+
+            _aesGcm.Encrypt(_nonce, plainText, cipherText, tag, packetLengthField);
```
zybexXL, line 74: You can avoid the BlockCopy() calls below by using a ReadOnlySpan for the cipherText and tag ouputs as well. This considerably increases performance, while also allocating only one buffer instead of two.

WojciechNagorski:
I will review in next few days.

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.cs
```diff
@@ -0,0 +1,101 @@
...
+
+            return result;
+        }
+
+        /// <inheritdoc/>
+        public override byte[] Decrypt(byte[] input, int offset, int length)
+        {
+            // [inbound sequence][packet length field][padding length field sz][payload][random paddings][Authenticated TAG]
+            // [-----4 bytes----][----4 bytes(offset)][------------------Cipher Text--------------------][-------TAG-------]
+            var packetLengthField = new ReadOnlySpan<byte>(input, offset - 4, 4);
```
Rob-Hague: It's quite unpleasant/unintuitive to assume there is data to process in `input` before `offset`. Is there a nicer way to do it?

Could packet_length (i.e. the associated data) start at `offset`?

At the very least, there should be some explicit `///` xmldoc here explaining the usage.

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.cs
```diff
@@ -0,0 +1,101 @@
...
+            var cipherText = new ReadOnlySpan<byte>(input, offset, length);
+            var tag = new ReadOnlySpan<byte>(input, offset + length, TagSize);
```
Rob-Hague: Same here, in assuming there is space beyond `offset + length`

Rob-Hague (commented):
The OpenSSH version of AES-GCM (https://github.com/openssh/openssh-portable/blob/43e7c1c07cf6aae7f4394ca8ae91a3efc46514e2/PROTOCOL#L82) states:

```
Additionally, if AES-GCM is selected as the cipher [...] there doesn't have to be a matching MAC.
```

could you please implement that?

scott-xu:
> The OpenSSH version of AES-GCM (https://github.com/openssh/openssh-portable/blob/43e7c1c07cf6aae7f4394ca8ae91a3efc46514e2/PROTOCOL#L82) states:
> 
> ```
> Additionally, if AES-GCM is selected as the cipher [...] there doesn't have to be a matching MAC.
> ```
> 
> could you please implement that?

I think the "matching MAC" means the matching MAC algorithm. See code here: https://github.com/sshnet/SSH.NET/pull/1364/files#diff-8cb58eb5577b3c26c820c320ea463952036f2aeadf28470a6265575168756164R1520-R1528

Rob-Hague:
That's good but I think if we offer:

```
Ciphers: aes128-gcm@openssh.com
HMACs: hmac-sha2-256
```

and they offer:

```
Ciphers: aes128-gcm@openssh.com
HMACs: hmac-sha1
```

then we are going to error out here despite not using the HMAC

https://github.com/sshnet/SSH.NET/blob/db3d7e8d03d0a852cd7d16e1dee95a72939de317/src/Renci.SshNet/Security/KeyExchange.cs#L99-L107

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.cs
scott-xu: I understand your concern. Typically, the cipher only decrypts "cipher text" which starts from offset and ends before MAC (tag). However, `AesGcm` requires the packet_length (aad) and MAC (tag) to do the decryption and authentication which is **before** and **after** the cipher text.
The `Decrypt(...)` method of `AesGcm` is called (and only called) from https://github.com/sshnet/SSH.NET/blob/develop/src/Renci.SshNet/Session.cs#L1302 which guarantees the data layout.

zybexXL: > The `Decrypt(...)` method of `AesGcm` is called (and only called) from https://github.com/sshnet/SSH.NET/blob/develop/src/Renci.SshNet/Session.cs#L1302 which guarantees the data layout.

Still, adding a bounds check could be a good thing for future proofing.

scott-xu: There are some options here:
1. keep it as is but add bounds check
2. in `Session` class, treat aead cipher as special and set offset before packet_length and include MAC(tag) in length
3. in `Session` class, treat aead cipher as special and set offset to 0 and include all in length (consider that ChaCha20Poly1305 requires sequence block if I read the [spec](https://datatracker.ietf.org/doc/html/draft-josefsson-ssh-chacha20-poly1305-openssh-00) correctly)

scott-xu: Just checked the source code of the constructor about `ReadOnlySpan<T>`. It does check the bounds:

- It casts the `start` to ulong/uint to make sure it is positive.
- It checks the `length` to no larger than array length - offset.

```
        /// <summary>
        /// Creates a new read-only span over the portion of the target array beginning
        /// at 'start' index and ending at 'end' index (exclusive).
        /// </summary>
        /// <param name="array">The target array.</param>
        /// <param name="start">The index at which to begin the read-only span.</param>
        /// <param name="length">The number of items in the read-only span.</param>
        /// <remarks>Returns default when <paramref name="array"/> is null.</remarks>
        /// <exception cref="System.ArgumentOutOfRangeException">
        /// Thrown when the specified <paramref name="start"/> or end index is not in the range (&lt;0 or &gt;Length).
        /// </exception>
        [MethodImpl(MethodImplOptions.AggressiveInlining)]
        public ReadOnlySpan(T[]? array, int start, int length)
        {
            if (array == null)
            {
                if (start != 0 || length != 0)
                    ThrowHelper.ThrowArgumentOutOfRangeException();
                this = default;
                return; // returns default
            }
#if TARGET_64BIT
            // See comment in Span<T>.Slice for how this works.
            if ((ulong)(uint)start + (ulong)(uint)length > (ulong)(uint)array.Length)
                ThrowHelper.ThrowArgumentOutOfRangeException();
#else
            if ((uint)start > (uint)array.Length || (uint)length > (uint)(array.Length - start))
                ThrowHelper.ThrowArgumentOutOfRangeException();
#endif

            _pointer = new ByReference<T>(ref Unsafe.Add(ref MemoryMarshal.GetArrayDataReference(array), (nint)(uint)start /* force zero-extension */));
            _length = length;
        }
```

WojciechNagorski:
@Rob-Hague Can I ask you for a review? My time is limited now.

scott-xu:
I renamed the source branch and this PR is closed automatically. I just created another PR https://github.com/sshnet/SSH.NET/pull/1369

zybexXL:
You mean #1369.
This PR could be reopened to keep the existing comment history.

scott-xu:
@zybexXL  thanks! I corrected the comments just now. Unfortunately, this PR cannot be reopened. The button is grayed out.

scott-xu:
Let me try again.

zybexXL:
Admins/Moderators should be able to reopen it.

Rob-Hague:
Presumably, you will at least need to restore the branch ref with the same name as before

zybexXL:
I think the branch is to remain deleted, it's just about reopening the PR.

scott-xu:
The PR is closed but this thread is not deleted. The history is not lost.
Is there anything can't be done without this PR to reopen?

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.cs
scott-xu: > There are some options here:
> 
> 1. keep it as is but add bounds check
> 2. in `Session` class, treat aead cipher as special and set offset before packet_length and include MAC(tag) in length
> 3. in `Session` class, treat aead cipher as special and set offset to 0 and include all in length (consider that ChaCha20Poly1305 requires sequence block if I read the [spec](https://datatracker.ietf.org/doc/html/draft-josefsson-ssh-chacha20-poly1305-openssh-00) correctly)

Personally I prefer option1, because it is consistent for all kinds of ciphers. No special handing in `Session` class. The current implementation in https://github.com/sshnet/SSH.NET/pull/1369 is option2. It defers the issue when add `ChaCha20Poly1305` in the future.

Life will be much easier if there's no such many options :upside_down_face:

scott-xu: Changed back to option1 with some tiny enhancements in the latest commit.

PR #1365, opened 2024-03-28
add optional execution timestamp

michaelSant0s:
When executing a command that outputs many lines of text its not possible to know when there was which output. In situations where you want to know how much time it took between outputs it would be helpful to see the exact time when the output occured. 

This is why i want to add an optional timestamp before the output of a command.

### DIFF
--- a/src/Renci.SshNet/SshClient.cs
+++ b/src/Renci.SshNet/SshClient.cs
@@ -239,6 +239,19 @@ namespace Renci.SshNet
         }
 
         /// <summary>
+        /// Creates the command.
+        /// </summary>
+        /// <param name="commandText">The command text.</param>
+        /// <param name="logExecutionTimeStamp">Returns a timestamp of the time the command was executed with the result.</param>
+        /// <returns><see cref="SshCommand"/> object with execution time in the result.</returns>
+        public SshCommand CreateCommand(string commandText, bool logExecutionTimeStamp)
+        {
+            EnsureSessionIsOpen();
+
+            return new SshCommand(Session, commandText, ConnectionInfo.Encoding, logExecutionTimeStamp);
+        }
+
+        /// <summary>
         /// Creates and executes the command.
         /// </summary>
         /// <param name="commandText">The command text.</param>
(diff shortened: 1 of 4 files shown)


Rob-Hague:
Thanks for the PR @michaelSant0s. I understand the use case but I'm not sure this is something worth building in to the library.

You might already be able to achieve the desired behaviour by wrapping `SshCommand.OutputStream` with a `System.IO.StreamReader`?

michaelSant0s:
> Thanks for the PR @michaelSant0s. I understand the use case but I'm not sure this is something worth building in to the library.
> 
> You might already be able to achieve the desired behaviour by wrapping `SshCommand.OutputStream` with a `System.IO.StreamReader`?

Thank you for helping me out! This works exactly like i need it to work.

michaelSant0s:
If somebody wants to have the same feature, here is an extension function i created with the help of @Rob-Hague 's informations. 


```
namespace Renci.SshNet
{
    public static class SshNetExtension
    {
        public static async Task<string> ExecuteWithTimestamp(this SshClient sshClient, string command)
        {
            using SshCommand sshCommand = sshClient.CreateCommand(command);
            IAsyncResult async = sshCommand.BeginExecute();

            Stream stream = sshCommand.OutputStream;
            StreamReader streamReader = new StreamReader(stream);

            string result = "";

            bool completed = false;
            Task streamReaderTask = Task.Run(() =>
            {
                while (!completed)
                {
                    if (streamReader.Peek() > 0)
                    {
                        result += $"[{DateTime.Now}]: {streamReader.ReadLine()}\n";
                    }
                    else
                    {
                        Thread.Sleep(5);
                    }
                }
            });

            string lastLine = sshCommand.EndExecute(async);
            result += $"[{DateTime.Now}]: {lastLine}\n";

            completed = true;
            await streamReaderTask;

            // using StreamReader will dispose it too early
            streamReader.Dispose();
            return result;
        }
    }
}

```

zybexXL:
You should use BeginRead with a callback instead of waking up the thread 200 times per second to poll the reader:
https://stackoverflow.com/questions/2784878/continuously-reading-from-a-stream

PR #1366, opened 2024-03-30
Implement OpenSSH strict key exchange extension

### LINKED ISSUE #1285: Terrapin attack and strict KEX
There is a recently published attack on SSH named [Terrapin](https://terrapin-attack.com/).

Is SSH.NET affected by Terrapin (does SSH.NET support any of the vulnerable encryption modes; does SSH.NET support the strict key exchange countermeasure)?

scott-xu:
The PR implements the algorithm described in section 1.10 of https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.

All integration tests passed, especially `*-etm@openssh.com` HMAC variants.

Resolves https://github.com/sshnet/SSH.NET/issues/1285

[Terrapin Scanner](https://github.com/RUB-NDS/Terrapin-Scanner) report **without** this PR:

```
Terrapin_Scanner_Windows_amd64.exe --listen 2222
Listening for incoming client connection on 127.0.0.1:2222
================================================================================
==================================== Report ====================================
================================================================================

Remote Banner: SSH-2.0-Renci.SshNet.SshClient.0.0.1

ChaCha20-Poly1305 support:   false
CBC-EtM support:             true

Strict key exchange support: false

The scanned peer is VULNERABLE to Terrapin.

Note: This tool is provided as is, with no warranty whatsoever. It determines
      the vulnerability of a peer by checking the supported algorithms and
      support for strict key exchange. It may falsely claim a peer to be
      vulnerable if the vendor supports countermeasures other than strict key
      exchange.

For more details visit our website available at https://terrapin-attack.com
```

[Terrapin Scanner](https://github.com/RUB-NDS/Terrapin-Scanner) report **with** this PR:

```
Terrapin_Scanner_Windows_amd64.exe --listen 2222
Listening for incoming client connection on 127.0.0.1:2222
================================================================================
==================================== Report ====================================
================================================================================

Remote Banner: SSH-2.0-Renci.SshNet.SshClient.0.0.1

ChaCha20-Poly1305 support:   false
CBC-EtM support:             true

Strict key exchange support: true

The scanned peer supports Terrapin mitigations and can establish
connections that are NOT VULNERABLE to Terrapin. Glad to see this.
For strict key exchange to take effect, both peers must support it.

Note: This tool is provided as is, with no warranty whatsoever. It determines
      the vulnerability of a peer by checking the supported algorithms and
      support for strict key exchange. It may falsely claim a peer to be
      vulnerable if the vendor supports countermeasures other than strict key
      exchange.

For more details visit our website available at https://terrapin-attack.com
```


src/Renci.SshNet/ConnectionInfo.cs
```diff
@@ -378,6 +378,7 @@ public ConnectionInfo(string host, int port, string username, ProxyTypes proxyTy
                     { "diffie-hellman-group14-sha256", () => new KeyExchangeDiffieHellmanGroup14Sha256() },
                     { "diffie-hellman-group14-sha1", () => new KeyExchangeDiffieHellmanGroup14Sha1() },
                     { "diffie-hellman-group1-sha1", () => new KeyExchangeDiffieHellmanGroup1Sha1() },
+                    { "kex-strict-c-v00@openssh.com", null },
```
Rob-Hague: As you said (somewhere), a `Dictionary<,>` maintains its order as an implementation detail, when only adding and not removing. In any case, we should explicitly filter this out in the algorithm selection in `KeyExchange`

src/Renci.SshNet/Session.cs
```diff
@@ -1446,6 +1464,18 @@ internal void OnKeyExchangeInitReceived(KeyExchangeInitMessage message)
 
             _keyExchangeCompletedWaitHandle.Reset();
 
+            if (_isInitialKex && message.KeyExchangeAlgorithms.Contains("kex-strict-s-v00@openssh.com"))
```
Rob-Hague, line 1463: This should also check that we have it defined in our algorithms (in case someone has removed it)

src/Renci.SshNet/Session.cs
```diff
@@ -1446,6 +1464,18 @@ internal void OnKeyExchangeInitReceived(KeyExchangeInitMessage message)
...
+                DiagnosticAbstraction.Log(string.Format("[{0}] Enabling strict key exchange extension.", ToHex(SessionId)));
+
+                if (_inboundPacketSequence != 1)
+                {
+                    throw new SshConnectionException("KEXINIT was not the first packet during strict key exchange", DisconnectReason.KeyExchangeFailed);
+                }
+            }
+
             // Disable messages that are not key exchange related
             _sshMessageFactory.DisableNonKeyExchangeMessages();
```
Rob-Hague: This disables most messages, but it doesn't disable SSH_MSG_IGNORE. We need to do that for strict KEX

Rob-Hague (commented):
It looks good so far. I know that the integration tests are testing this in the positive case (i.e. they are using strict KEX), but it would make sense to add some tests in the negative cases assuming a bad actor, e.g.

* When we have strict KEX established, and we receive SSH_MSG_IGNORE (or other) during the key exchange. We should check that the connection fails.
* Conversely, when we don't have strict KEX established, we should not fail when receiving SSH_MSG_IGNORE (or other) during the key exchange
* When we have strict KEX established, and the peer does not reset the sequence number. We should check that the connection fails.
* Maybe some other cases?

test/Renci.SshNet.Tests/Classes/SessionTest_ConnectedBase.cs is probably a good starting point for simulating connections (either deriving from that or copying stuff out)

src/Renci.SshNet/ConnectionInfo.cs
scott-xu: With latest commit, it only adds "kex-strict-c-v00<nolink/>@openssh.com" in the very first key exchange message.

src/Renci.SshNet/Session.cs
scott-xu: With latest commit, it adds "kex-strict-c-v00<nolink/>@openssh.com" in the very first key exchange message. There's nowhere to remove it.

src/Renci.SshNet/Session.cs
scott-xu: Good catch. Fixed in the latest commit

src/Renci.SshNet/Session.cs
```diff
@@ -1335,6 +1333,11 @@ private Message ReceiveMessage(Socket socket)
 
             _inboundPacketSequence++;
 
+            if (_inboundPacketSequence == uint.MaxValue && _isInitialKex)
+            {
+                throw new SshConnectionException("Inbound packet sequence number is about to wrap during initial key exchange.", DisconnectReason.KeyExchangeFailed);
+            }
```
scott-xu, line 1350: This piece of code is mirrored from https://github.com/hierynomus/sshj/pull/917/comtmits/94fcc960e0fb198ddec0f7efc53f95ac627fe083 and https://github.com/openssh/openssh-portable/commit/1edb00c58f8a6875fad6a497aa2bacf37f9e6cd5.
However, I don't quite understand the motivation.
We've already made sure the kex init is the first packet.

Rob-Hague: I guess in theory an attacker could inject uint.MaxValue messages of their choosing (before KEXINIT) which would be undetected by the subsequent check on the sequence number after receiving NEWKEYS

Probably worth a comment with a link to the openssh implementation at least?

test/Renci.SshNet.Tests/Classes/SessionTest_Connecting_ServerSendsDebugMessageAfterKexInit_NoStrictKex.cs
```diff
@@ -0,0 +1,50 @@
...
+
+            // MitM drops server message
+            ServerOutboundPacketSequence++;
+        }
+
+        [TestMethod]
+        public void ThrowsSshException()
+        {
+            // Should we allow debug message during kex in non-strict-kex mode?
+            // Probably better to keep this behavior as is, unless someone strongly disagree.
```
Rob-Hague: I agree there is no need to change it. Although I don't think it's worth asserting the current behaviour

test/Renci.SshNet.Tests/Classes/SessionTest_Connecting_ServerSendsMaxIgnoreMessagesBeforeKexInit.cs
```diff
@@ -0,0 +1,33 @@
...
+            var ignore = ignoreMessage.GetPacket(8, null);
+            for (uint i = 0; i < uint.MaxValue; i++)
+            {
+                // MitM sends ignore message to client
+                _ = ServerSocket.Send(ignore, 4, ignore.Length - 4, SocketFlags.None);
+            }
+        }
+
+        [TestMethod]
+        [Ignore("It takes hours to send 4294967295 ignore messages.")]
```
Rob-Hague: Delete it then? I don't think anyone will ever run this

Rob-Hague (commented):
Thanks for all the tests. I will run this in production for a week or two

src/Renci.SshNet/Session.cs
scott-xu: I have the same feeling that it tries to ensure the integrity of key exchange, thought I can't think of a hack model here.

test/Renci.SshNet.Tests/Classes/SessionTest_Connected.cs
```diff
@@ -30,6 +31,31 @@ public void ClientVersionIsRenciSshNet()
...
+        {
+            Assert.IsTrue(ServerBytesReceivedRegister.Count > 0);
+
+            var kexInitMessage = new KeyExchangeInitMessage();
+            kexInitMessage.Load(ServerBytesReceivedRegister[0], 4 + 1 + 1, ServerBytesReceivedRegister[0].Length - 4 - 1 - 1);
+            Assert.IsTrue(kexInitMessage.KeyExchangeAlgorithms.Contains("kex-strict-c-v00@openssh.com"));
+        }
+
+        [TestMethod]
+        public void ShouldNotIncludeStrictKexPseudoAlgorithmInSubsequenceKex()
```
Rob-Hague:
```suggestion
        public void ShouldNotIncludeStrictKexPseudoAlgorithmInSubsequentKex()
```

Rob-Hague (approved):

Rob-Hague:
Thanks

Issue #719, opened 2020-08-24
Dictionary enumeration order is relied upon in ConnectionInfo despite being undefined behaviour


MattJamesChampion:
In [`ConnectionInfo.cs`](https://github.com/sshnet/SSH.NET/blob/develop/src/Renci.SshNet/ConnectionInfo.cs) there are a number of attributes of type `IDictionary` that rely on the ordering that entities were added to the dictionary, even though the documentation for `Dictionary` states that [the order in which the items are returned is undefined](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2#remarks).

As a concrete example, take [`HostKeyAlgorithms`](https://github.com/sshnet/SSH.NET/blob/cefdc203d98cd890815e029bc759bc43ec5a9643/src/Renci.SshNet/ConnectionInfo.cs#L60). Modifying `HostKeyAlgorithms` as part of `ConnectionInfo` appears to be the method used to specify which algorithms are used and in what order. As an example, [reversing the order that the collection contents are added in](https://gist.github.com/MattJamesChampion/976a093739f47116bfd533209a640243) produces different host key fingerprints, which indicates that the order matters internally. As well as that, [the `Test_KeyExchangeInitMessage_Load` test uses `SequenceEqual` assertions to check the contents of `ServerHostKeyAlgorithms`](https://github.com/sshnet/SSH.NET/blob/7691cb0b55f5e0de8dc2ad48dd824419471ab710/src/Renci.SshNet.Tests/Classes/Messages/Transport/KeyExchangeInitMessageTest.cs#L35) which imply that the order is expected to be consistent.

Since the order that `Dictionary` items are enumerated is undefined according to the documentation/standard, it's possible that a valid implementation could return values in a different order and therefore cause unexpected and inconsistent results. Using the above example, if a different host key algorithm is successfully negotiated first, the returned fingerprint would be different and could lead to the connection not being trusted (but only when the algorithms are enumerated in a certain order).

The [`OrderedDictionary`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.ordereddictionary) class and the [`IOrderedDictionary`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.iordereddictionary) interface are potential replacements since they guarantee a consistent order, although they aren't drop-in replacements.

warwickmm:
This seems pretty important.  Can we get some more attention on this issue?

scott-xu:
[...]

PR #1369, opened 2024-04-04
Add support for AEAD AES 128/256 GCM Ciphers (.NET 6.0 onward only)

### LINKED ISSUE #477: AES-GCM

### LINKED ISSUE #555: AES-GCM Cipher Mode

### LINKED ISSUE #773: Support aes256-gcm encryption ?

### LINKED ISSUE #774: is it possible to add AES 128-GCM Cipher Suite or ssh agent forwarding utility?

### LINKED ISSUE #792: AES-GCM 128 into SSH-NET using MS AES-GCM class

### LINKED ISSUE #994: Request to verify it with TLS 1.3 ciphers and check on Renci.SshNet.dll

scott-xu:
This PR adds support for **aes128-gcm<nolink>@openssh.com** and **aes256-gcm<nolink>@openssh.com** described in https://datatracker.ietf.org/doc/html/rfc5647

Resolves https://github.com/sshnet/SSH.NET/issues/792
Contributes to https://github.com/sshnet/SSH.NET/issues/1356
Resolves https://github.com/sshnet/SSH.NET/issues/774
Resolves https://github.com/sshnet/SSH.NET/issues/773
Resolves https://github.com/sshnet/SSH.NET/issues/555
Resolves https://github.com/sshnet/SSH.NET/issues/477
Resolves https://github.com/sshnet/SSH.NET/issues/994
Closes https://github.com/sshnet/SSH.NET/pull/877

**Notes:**
1. This PR does **not** add support for **chacha20-poly1305<nolink>@openssh.com** described in https://datatracker.ietf.org/doc/html/draft-josefsson-ssh-chacha20-poly1305-openssh-00 because it requires `ChaCha20` to encrypt/decrypt the packet length field.
    The BCL adds support for [ChaCha20Ploy1305](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.chacha20poly1305) since .NET 6.0, but **no** standalone `ChaCha20` till now (2024-04-04).
    At the beginning, I created a branch named as "**aes-gcm-and-chacha20-poly1305**" in my fork and created a PR https://github.com/sshnet/SSH.NET/pull/1364. Then when I realize there's no direct way to implement **chacha20-poly1305<nolink>@openssh.com**, I renamed the branch to **aesgcm**. The original PR is closed automatically after renaming. I have to create this new PR. Please refer the previous PR for review comment history. Thank @zybexXL @Rob-Hague for reviewing.
2. The AES-GCM ciphers are inserted right after AES-CTR ciphers but before AES-CBC ciphers, which is kind of similar with OpenSSH:
      ```
      debug2: ciphers ctos: chacha20-poly1305@openssh.com,aes128-ctr,aes192-ctr,aes256-ctr,aes128-gcm@openssh.com,aes256-gcm@openssh.com
      debug2: ciphers stoc: chacha20-poly1305@openssh.com,aes128-ctr,aes192-ctr,aes256-ctr,aes128-gcm@openssh.com,aes256-gcm@openssh.com
      ```
      Although `Dictionary`'s order is not defined, from observation, it is in the same order with add. Anyway that would be another topic, see https://github.com/sshnet/SSH.NET/issues/719


Rob-Hague:
Generally it looks good to me but I'll take a closer look soon.

One thing is that `S.S.C.AesGcm` has an [IsSupported](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.aesgcm.issupported) property, but only on >=NET 6. It probably makes sense to be checking that. I would suggest guarding AES-GCM support on `#if NET6_0_OR_GREATER` in order to keep it simple there (i.e. drop the support for AES-GCM on netstandard2.1). That would also mean only including the cipher in the ConnectionInfo if it is supported.

(side note: if you don't want to close #1356, replace "fix" with a non-keyword e.g. "contributes to")

src/Renci.SshNet/ConnectionInfo.cs
```diff
@@ -55,7 +55,9 @@ public class ConnectionInfo : IConnectionInfoInternal
         /// <summary>
         /// Gets supported encryptions for this connection.
         /// </summary>
+#pragma warning disable CA1859 // Use concrete types when possible for improved performance
```
Rob-Hague, line 58: We should probably delete this at some point:

https://github.com/sshnet/SSH.NET/blob/7b58819537798bec53e4193a3e5f395bcf018023/.editorconfig#L681

cc @jscarle

src/Renci.SshNet/Security/Cryptography/Cipher.cs
```diff
@@ -13,6 +13,14 @@ public abstract class Cipher
         /// </value>
         public abstract byte MinimumSize { get; }
 
+        /// <summary>
+        /// Gets the tag (MAC) size.
+        /// </summary>
+        /// <value>
+        /// The tag (MAC) size.
+        /// </value>
+        public virtual int TagSize { get; }
```
Rob-Hague, line 23: Could it be done another way? It doesn't seem so good to put this on the base `Cipher` class

You had an `AeadCipher` class before. Putting this property there might be a bit better at least. Or perhaps replace `bool IsAead` with `int? AeadTagSize` on `CipherInfo` - null means it is not AEAD?

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.cs
```diff
@@ -0,0 +1,137 @@
...
+        /// <param name="offset">The zero-based offset in <paramref name="input"/> at which to begin encrypting.</param>
+        /// <param name="length">The number of bytes to encrypt from <paramref name="input"/>.</param>
+        /// <returns>
+        /// The packet length field + cipher text + tag.
+        /// </returns>
+        public override byte[] Encrypt(byte[] input, int offset, int length)
+        {
+            // [outbound sequence field][packet length field][padding length field sz][payload][random paddings]
+            // [----4 bytes----(offset)][------4 bytes------][----------------Plain Text---------------(length)]
+            var packetLengthField = new ReadOnlySpan<byte>(input, offset, 4);
```
Rob-Hague: Could you please put the comments explaining the usage into the /// xmldoc? e.g. `The first 4 bytes of the input are treated as the associated data for the algorithm.` (and perhaps, rename `packetLengthField` to `associatedData`)

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.cs
```diff
@@ -0,0 +1,137 @@
...
+        /// <param name="offset">The zero-based offset in <paramref name="input"/> at which to begin decrypting and authenticating.</param>
+        /// <param name="length">The number of bytes to decrypt and authenticate from <paramref name="input"/>.</param>
+        /// <returns>
+        /// The packet length field + plain text.
+        /// </returns>
+        public override byte[] Decrypt(byte[] input, int offset, int length)
+        {
+            // [inbound sequence field][packet length field][padding length field sz][payload][random paddings][Authenticated TAG]
+            // [----4 bytes---(offset)][------4 bytes------][------------------Cipher Text--------------------][---TAG---(length)]
+            var packetLengthField = new ReadOnlySpan<byte>(input, offset, 4);
```
Rob-Hague: Similarly, move the explanation to /// xmldoc

src/Renci.SshNet/Security/Cryptography/Cipher.cs
scott-xu: [SSHJ](https://github.com/hierynomus/sshj) puts TagSize at BaseCipher. See https://github.com/hierynomus/sshj/blob/master/src/main/java/net/schmizz/sshj/transport/cipher/BaseCipher.java#L65

Using `AeadCipher` requires `as` operation in `Session` which seems not generic?
`int? AeadTagSize` implies AEAD when value is null which seems not intuitive?

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.cs
scott-xu: Well, if it is named as `associatedData`, we need to explain that it comes from `packetLengthField`; If it is named as `packetLengthField`, we need to explain that it is treated as `associatedData`. :rofl:

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.cs
scott-xu: Actually, I once put the xml doc comments. There are 2 small issues
1. I once put the structure in xml doc `remarks` section. However, it requires a period at the end which means I need add some paragraphs after `[---TAG---(length)]`
   See https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1629.md
3. xml doc comments are primarily for lib consumers. This kind of details might be more suitable for lib developers. Is there any existing simlar details in xml doc comments?

jscarle: > Actually, I once put the xml doc comments. There are 2 small issues
> 
> 1. I once put the structure in xml doc `remarks` section. However, it requires a period at the end which means I need add some paragraphs after `[---TAG---(length)]`
>    See https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1629.md
> 2. xml doc comments are primarily for lib consumers. This kind of details might be more suitable for lib developers. Is there any existing simlar details in xml doc comments?

You could try using the `<code>` element: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/recommended-tags#code

scott-xu: @jscarle Thanks. I did think to put it inside `<code>` element. However I doubted it is code. It is quite different with C#. I'm not native English speaker but just now I realized the word "code" can be very generic.

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.cs
scott-xu: I'm using named parameter in the latest commit, so that it should be very clear that the packet length field is treated as associated data.

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.cs
scott-xu: Done

src/Renci.SshNet/Security/Cryptography/Cipher.cs
Rob-Hague: I'd probably prefer

> `as` operation in `Session`

(it could even be enforced that the cipher is `AeadCipher` in `KeyExchange.CreateCipher`)

but in the interest of progress, if you want to do it this way, let's at least flesh out the comment (suggestion in different thread because github...)

Rob-Hague:
```suggestion
        /// <summary>
        /// Gets the size of the authentication tag for ciphers which implement Authenticated Encryption (AE).
        /// </summary>
        /// <returns>
        /// When this <see cref="Cipher"/> implements Authenticated Encryption, the size, in bytes,
        /// of the authentication tag included in the encrypted message.
        /// </returns>
        public virtual int TagSize { get; }
```

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.cs
```diff
@@ -0,0 +1,171 @@
...
+        /// <returns>
+        /// The decrypted data with below format:
+        /// <code>
+        ///   [padding length field sz][payload][random paddings]
+        ///   [--------------------Plain Text-------------------]
+        /// </code>
+        /// </returns>
+        public override byte[] Decrypt(byte[] input, int offset, int length)
+        {
+            var packetLengthField = new ReadOnlySpan<byte>(input, 4, 4);
```
Rob-Hague, line 125: Was it necessary to revert back to this way ? 🙁 It looked OK before

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.cs
Rob-Hague: Thanks for adding comments 👍

src/Renci.SshNet/Security/Cryptography/Cipher.cs
scott-xu: Could you please help me understand *it could even be enforced that the cipher is AeadCipher in KeyExchange.CreateCipher*

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.cs
scott-xu: Hmm... just want to share that it will have problem when deal with **chacha20-poly1305<nolink/>@openssh.com** described https://datatracker.ietf.org/doc/html/draft-josefsson-ssh-chacha20-poly1305-openssh-00#section-4
 > When receiving a packet, the length must be decrypted first.  When 4
 > bytes of ciphertext length have been received, they may be decrypted
 > using the K_1 key, a nonce consisting of the packet sequence number
 > encoded as a uint64 under the usual SSH wire encoding and a zero
 > block counter to obtain the plaintext length.

My first reaction is to make `byte[] Cipher.Decrypt(byte[] input)` virtual and **override** it in
1. AesGcmCipher and just return the input data without doing anything
2. ChaCha20Poly1305Cipher so that it decrypts the packet data length. However, it requires the packet sequence number as the nonce. How can we transfer the the packet sequence number to ChaCha20Poly1305Cipher？

src/Renci.SshNet/Security/Cryptography/Cipher.cs
Rob-Hague: Yes, sorry, I was thinking something like 

```c#
            // Create server cipher
            var cipher = _serverCipherInfo.Cipher(serverKey, serverVector);

            if (isAead && cipher is not AeadCipher)
            {
                throw new Exception("...");
            }

            return cipher;
```

but it is still not very elegant. You can leave it how it is

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.cs
Rob-Hague: I don't have a good understanding of chacha20-poly1305, but it sounds like it might be worth having a more general abstraction for building & deserializing packets, rather than trying to force these different schemes through the existing `Cipher` API.

For now, can we either put this bit back so that packet_length starts at offset, or just throw if `offset != 8` to make the behaviour more explicit. And then make AesGcmCipher internal so that it's harder to use directly? Then we can change it later

scott-xu: That's a good idea.

scott-xu: Actually, it might be good to set all ciphers to internal. They are SSH specific and are not intended to be consumed externally.
However, it's hard to change public APIs to internal as it is breaking change. We have no idea how they are being used.

scott-xu: Using `Debug.Assert(offset == 8, "The offset must be 8");`  should be sufficient.

Rob-Hague: Yes, let's leave the others. Aes is used externally, e.g. https://github.com/darinkes/SshNet.Keygen/blob/b230c43c53ceec442c3bb563957cd53bd03cac3e/SshNet.Keygen/SshKeyEncryption/SshKeyEncryptionAes256.cs#L74

Most of the rest I would like to delete

Rob-Hague (approved):
Thank you

Rob-Hague:
fyi you can click this to see your PRs without needing to assign yourself:

image

scott-xu:
Thanks

Dlyftservies:
Thanks

Issue #1271, opened 2023-12-10
Third party libraries


WojciechNagorski:
Does anyone know why SSH.NET uses copies of third-party libraries:
- BouncyCastle: Based on Version 1.8.3 from http://www.bouncycastle.org/csharp/ (currently there is 2.2.1)
- Chaos.NaCl: https://github.com/CodesInChaos/Chaos.NaCl/commit/2c861348dc45369508e718aa08611c53b53553db

source of information: https://github.com/sshnet/SSH.NET/pull/496#issuecomment-573557849

I wonder if it would be better to use Nuget. We would receive security updates, bug fixes, and optimizations.

I looked through the code coverage and for the most part, the copied code is not covered by tests.

@scott-xu @Rob-Hague @drieseng @jacobslusser

Rob-Hague:
I don't know why it was copied, but one possible explanation and downside of using the nuget package: the bouncycastle binary is nearly 7 megabytes.

The size wouldn't be a blocker for me (I would prefer the nuget). I imagine we wouldn't need Chaos.NaCl: I think bouncycastle could be used for Ed25519

> I looked through the code coverage and for the most part, the copied code is not covered by tests.

Most of the internal code is unused: #1140

Rob-Hague:
cc @darinkes

jacobslusser:
@WojciechNagorski, I think you raise an excellent question. If we are counting votes, I would be in favor of using the third-party nuget packages instead of copying the code.

darinkes:
At that time there were no usable NuGets (Chaos.NaCl still hasnt) and BouncyCastle is a huge bloat we just needed a very small part of. Thats why we went the route to import only needed stuff.

drieseng:
I'm ok with switching to NuGet. 

I suppose the reasons for including the source code were:
* Reduce the number of - direct or indirect - dependencies to the absolute minimum hereby avoid dll/assembly hell.
* Reduce the on-disk foodprint of SSH.NET.
* Support for legacy target frameworks.

Perhaps the last one was - at that time - the most important reason.

WojciechNagorski:
The BouncyCastle code has been removed. I think we can get rid of Chaos.NaCI now.

PR #1370, opened 2024-04-05
Replace internal BouncyCastle with NuGet package

mus65:
Contributes to #1271 .

This was as simple as adding the NuGet package, deleting the old code and replacing the namespaces. It's just a matter of whether the tests pass and you actually want to do this. I don't think there was a clear decision in #1271 .


mus65:
Guess it wasn't this simple after all, a lot of tests fail with a NullReferenceException. I'm gonna take a look within the next few days. Converted to draft.

mus65:
tests fixed  with b7f6cd4b4f86a2c8c87ec0d8fee2a4f354683f58

scott-xu:
How about leveraging BCL's `ECDiffieHellman.DeriveRawSecretAgreement` for .NET 8 onward? See https://github.com/dotnet/runtime/issues/71613

mus65:
Thanks, that's good to know. Unfortunately this API is not supported on Windows Server 2012, so even net8.0 builds would still need the BouncyCastle reference.

Not saying this shouldn't be done. But since this wouldn't reduce the dependency to BouncyCastle it imho shouldn't block this PR and could be done separately.

darkoperator:
Out of curiosity what is the lowest supported version of the Windows. Windows 2012 is EOL

scott-xu:
@mus65 I created a separate PR https://github.com/sshnet/SSH.NET/pull/1371. Might bring a little bit conflicts with your PR but it should be straightforward to resolve and should not block this PR.
@darkoperator See this please: https://github.com/sshnet/SSH.NET/pull/1371#issuecomment-2041099786

Rob-Hague:
Thanks. Indeed there was no decision on this, and I don't have a preference either way (which probably means I'd lean towards not doing anything). I think I'll leave this one for a while for others to chime in.

One thing: this library has a similar internal dependency on some Chaos.NaCl code for ed25519. Perhaps as part of this change (if it were to be merged), that code could be removed as well and replaced with some additional usage of BouncyCastle. Would you be interested in investigating that?

> Guess it wasn't this simple after all

FYI, running the integration tests locally is very simple - just install Docker Desktop and it should "just work". It's a lot faster than relying on CI.

mus65:
> One thing: this library has a similar internal dependency on some Chaos.NaCl code for ed25519. Perhaps as part of this change (if it were to be merged), that code could be removed as well and replaced with some additional usage of BouncyCastle. Would you be interested in investigating that?

I actually had a look at that, but unfortunately I'm not familiar enough with cryptography to do this.

Side note: the internal BouncyCastle doesn't support Ed25519 yet, so in order to get rid of Chaos.NaCI, either this PR needs to be merged first or the internal BouncyCastle needs to be updated. I also found this [BouncyCastle issue](https://github.com/bcgit/bc-csharp/issues/188) about some implementation difference with NaCl (I don't know if this is relevant for SSH.NET).

Rob-Hague:
Just curious to see latest CI

lifeincha0s:
Chiming in... Adding BouncyCastle as a NuGet package would add all of the lib features, or am I wrong about that? Wouldn't that be a bit excessive since only a very limited subset is actually implemented and NET has been supporting more and more crypto natively.

mus65:
Updated to BouncyCastle 2.3.1 which fixed [multiple CVEs](https://www.bouncycastle.org/csharp/#DOWNLOAD20301). These CVEs are obviously unfixed in the internal implementation. I can't tell whether these actually affect SSH.NET, but either way this imho would be another reason in favor of the NuGet package.

If one of the CVEs affects SSH.NET, fixing it would require updating the internal implementation and making a new release. With the NuGet package, any application using SSH.NET could simply update to the newer BouncyCastle version to get the security fixes.

Rob-Hague:
> Wouldn't that be a bit excessive since only a very limited subset is actually implemented

Right, if we are only using it for ECDH it is arguably not worth it. But if we can also use it for ed25519, bcrypt, standard DH(?) and Chachapoly(?) then there is more of a case for it.

> I can't tell whether these actually affect SSH.NET, but either way this imho would be another reason in favor of the NuGet package.

They don't look like they affect our usage but your point stands either way.

The hesitation stems from that the BC assembly does not play well with trimming. I've opened https://github.com/bcgit/bc-csharp/pull/534 to partly address that and a full fix is possible. I'd like to see if that gets anywhere. If it does then we will be able to say if you care about the extra size then you should trim your app, and if you don't care then you don't care. Right now a trimmed BC assembly is still 4MB.

If the BC trimming changes don't get anywhere then I guess we just bite the bullet.

mus65:
Updated to BouncyCastle 2.4.0 which already includes @Rob-Hague's trimming fixes. :smile:

Rob-Hague:
I think we should take this for this release. I will come back to it after some other stuff. I want to a) experiment with replacing the X25519 Na.Cl code as well; and b) check the size of an AOT app, perhaps with some check in CI

WojciechNagorski:
IMO, we should use BouncyCastle Nuget and mark all old releases as unsafe. Let's look at how many vulnerabilities they resolved in old versions. Then we can implement #1416 PR using this library.
@Rob-Hague Do you agree or would you prefer to update our source to the latest version?

Rob-Hague (approved):
Agreed, I won't hold this up any longer, so let's take it. Renci.SshNet.AotCompatibilityTestApp.exe is 9.0MB without this change and 9.8MB with this change

WojciechNagorski (approved):
LGTM

WojciechNagorski:
I think we can get rid of Chaos.NaCI now.

scott-xu:
Chaos.NaCl can be removed if merge https://github.com/sshnet/SSH.NET/pull/1416, https://github.com/sshnet/SSH.NET/pull/1447 and https://github.com/sshnet/SSH.NET/pull/1448. @WojciechNagorski @Rob-Hague

Rob-Hague:
Nice, I'll review in the next few days

PR #1375, opened 2024-04-16
Fix potential side-channel timing attack issue

scott-xu:
The PR fixes a potential side-channel timing attack issue when verify HMAC and verify DigitalSignature.


Rob-Hague:
What's the attack? The signature verification does not involve any private data, and in both cases we will drop the connection once the verification fails. What am I missing?

scott-xu:
AFAK, the timing attack is based on enumerous failures.

scott-xu:
https://github.com/openssh/openssh-portable/blob/88351eca17dcc55189991ba60e50819b6d4193c1/mac.c#L222
FYI. OpenSSH also prevents timing attack when verify MAC.

scott-xu:
I come up with this PR when read [Chaos.Nacl](https://github.com/CodesInChaos/Chaos.NaCl) project.
The readme.md says: (forget the typos 🙄)
> 
> ```
> public static bool ContantTimeEquals(byte[] x, byte[] y)
> ```
> Checks if the contents of the two arrays are the same and returns truie if they are equal.
Throws an expection if their lengthes differ.
> 
> The runtime of this method does not depend on the contents of the arrays. Using constant time prevents timing attacks that allow an attacker to learn if the arrays have a common prefix. It is important to use such a constant time comparison when verifying MACs.

Rob-Hague (approved):
Ok :shrug:

PR #1382, opened 2024-04-25
Tweak AsyncSocketListener in tests

Rob-Hague:
Some CI runs have been crashing lately from within AsyncSocketListener. This moves a bit of disconnection & exception handling around. It probably won't fix the underlying tests which would have otherwise failed, but it might stop the process from outright crashing.

Also reference the same code for the integration tests.


test/Renci.SshNet.Tests/Common/AsyncSocketListener.cs
```diff
@@ -322,23 +284,36 @@ void ConnectionDisconnected()
                 }
                 catch (ObjectDisposedException)
                 {
-                    // TODO On .NET 7, sometimes we get ObjectDisposedException when _started but only on appveyor, locally it works
-                    ConnectionDisconnected();
```
Rob-Hague, line 326: This was the main problem. If we have received ObjectDisposedException, we don't want to call `Shutdown` on the socket because we will get another ObjectDisposedException, and that will crash the process.

WojciechNagorski:
This is a very difficult piece of code. Sockets will behave differently on Linux and Windows and in the .NET Framework and modern .NET. In addition, different tests use the `AsyncSocketListener` differently. But I agree that it should be a little easier.

Rob-Hague:
Yeah, I figured this at least removes one way of crashing. There might still be more

KodakDave00119 (approved):

test/Renci.SshNet.IntegrationTests/Renci.SshNet.IntegrationTests.csproj
```diff
@@ -8,6 +8,10 @@
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
   </PropertyGroup>
 
+  <ItemGroup>
+    <Compile Include="$(ProjectDir)../Renci.SshNet.Tests/Common/AsyncSocketListener.cs" />
```
WojciechNagorski, line 12: Great idea 👍

WojciechNagorski (approved):

PR #1392, opened 2024-05-05
Enable nullable on NetConf/Scp/SshClient

mus65:
Followup #1339


mus65:
CI failure looks unrelated. Windows works and I can't reproduce it locally on Linux.

src/Renci.SshNet/ScpClient.cs
```diff
@@ -747,7 +784,14 @@ private void InternalDownload(IChannelSession channel, Stream input, FileSystemI
...
 
-                    currentDirectoryFullName = new DirectoryInfo(currentDirectoryFullName).Parent.FullName;
+                    var currentDirectoryParent = new DirectoryInfo(currentDirectoryFullName).Parent;
+
+                    if (currentDirectoryParent is null)
+                    {
+                        break;
+                    }
+
+                    currentDirectoryFullName = currentDirectoryParent.FullName;
```
Rob-Hague: I think this actually fixes a NullReferenceException when downloading a directory to C:\\. It would be more explicit if the `directoryCounter == 0` check were first (I can't add as a suggestion):

```c#
directoryCounter--;

if (directoryCounter == 0)
{
    break;
}

var currentDirectoryParent = new DirectoryInfo(currentDirectoryFullName).Parent;

Debug.Assert(currentDirectoryParent is not null, $"Should be {directoryCounter} levels deeper than {startDirectoryFullName}.");

currentDirectoryFullName = currentDirectoryParent.FullName;

continue;
```

mus65: pushed. Had to add a `!` to `currentDirectoryParent!.FullName` though because legacy frameworks don't have the nullable annotation on `Debug.Assert`. :(

mus65: @Rob-Hague I just pushed another change to simply disable the nullable warnings on old frameworks since these kind of false positives would become very annoying when migrating more of the codebase to nullable.

Rob-Hague (approved):
Thanks

PR #1399, opened 2024-05-16
Fix a few issues with PipeStream

### LINKED ISSUE #12: PipeStream returns 0 instead of blocking when running a command with intermittent output

### LINKED ISSUE #142: SshCommand OutputStream has no way to directly set BlockLastReadBuffer
When reading the stdout of a process executed through SshCommand the output stream will throw an exception if you read beyond what's available in the buffer and it seems to be by design.

Although in my opinion this is the wrong default, there's no direct way to change the property at runtime as the exposed interface is the standard `Stream`.

Could the default be changed to more sensible blocking behavior or there be a public method/property that can be used to change it at runtime?

### LINKED ISSUE #164: Cannot get streaming stdout/stderr and wait for remote process to exit

### LINKED ISSUE #440: PipeStream not supporting Read offsets > 0

### LINKED ISSUE #1045: No response due to the file size over the default MaxBufferLength of PipeStream
The default MaxBufferLength of PipeStream is _maxBufferLength = 200 * 1024 * 1024
The issue occurs when there is a file in linux, whose sizes are over the default MaxBufferLength.
Steps:
1. Given a JsonData1.txt, whose sizes are 210MB ( > 200MB), and a JsonData2.txt, whose sizes are 190MB (< 200MB)
2. Copy both JsonData1.txt and JsonData2.txt  into /test/ in your linux.
3. Test the JsonData by attached test.zip file.
4. Config commandtext= "cat /test/JsonData1.txt" in app.config, and config other items.
5. Run it, then you will get a time out issue. Actually, it should spend about 17s, if you change _maxBufferLength = **300** * 1024 * 1024
6. Config commandtext= "cat /test/JsonData2.txt" in app.config.
7. Run it, then you will get an available timespan.

[Test.zip](https://github.com/sshnet/SSH.NET/files/10011454/Test.zip)

Rob-Hague:
Apply similar treatment to PipeStream as #1322 did to ShellStream

PipeStream now behaves much more Stream-like. In particular, it performs partial reads (instead of blocking until a certain amount of data is available), blocks until data is available (instead of returning 0 prematurely) and removes the Stream-unlike properties `BlockLastReadBuffer` and `MaxBufferLength`.

Sadly I gave up trying to make a benchmark compatible with all the quirks of the previous implementation, but a dumb throughput test (reading and writing simultaneously) shows about 5.2GB/s with this implementation compared to 140MB/s previously.

Some cleanup of its usage in the library followed.

contributes to #650 (does not make any of the suggested source breaking changes - still worth considering)
closes #440
closes #164
closes #12
closes #1045
closes #142

of existing PRs:
closes #1070
closes #924
closes #374
closes #13


test/Renci.SshNet.IntegrationTests/SshTests.cs
```diff
@@ -297,8 +289,16 @@ public void Ssh_Command_IntermittendOutput_OutputStream()
...
+                        // command.Result (also returned from EndExecute) consumes OutputStream,
+                        // which we've already read from, so Result will be empty.
+                        // TODO consider the suggested changes in https://github.com/sshnet/SSH.NET/issues/650
+
+                        //Assert.AreEqual(expectedResult, actualResult);
+                        //Assert.AreEqual(expectedResult, command.Result);
+
+                        // For now just assert the current behaviour.
+                        Assert.AreEqual(0, actualResult.Length);
+                        Assert.AreEqual(0, command.Result.Length);
```
Rob-Hague, line 301: I was initially planning on getting this PR in, then getting a release out. Now I am thinking of addressing #650 beforehand by taking the break changing `string Execute()` to `void Execute()` in order to stop `Execute` from consuming `OutputStream`.

That way, we can get the CancelAsync changes (#1345), this PR with its fixes and subtle behaviour changes, and the source-breaking changes wrapped up in one release. Probably easier for consumers.

WojciechNagorski (approved):
LGTM!

Rob-Hague:
Thank you

vladd:
Thank you for fixing all these problems! Is there any estimation, in which release is the fix going to be included?

Rob-Hague:
I have a couple more fixes/changes I'd like to make to SshCommand before the next release, and then we can make one then. So I would say June, but the last time I answered that question I said May and it won't be May :-)

PR #1400, opened 2024-05-16
dotnet pack in CI

Rob-Hague:
and delete the old build script


Rob-Hague:
Before (from https://www.nuget.org/packages/SSH.NET):

image

After (https://ci.appveyor.com/project/drieseng/ssh-net/builds/49850281/job/8p8klyxta9gy1r3c/artifacts)

image

Minor annoyance: the artifacts download as *.zip so you have to change the file extension. Not sure if there is a way to make them download as e.g. *.nupkg

WojciechNagorski (approved):

PR #1401, opened 2024-05-17
use Regex Source Generator for .NET 7+

### LINKED ISSUE #1131: Use regex source generators when target .NET 7+
Use conditional compilation to switch to regex source generators when targeting .NET 7+.
When done, remove the SYSLIB1045 "exclusion" from the **.editorconfig** in the **Renci.SshNet** project.

mus65:
fixes #1131

### DIFF
--- a/src/Renci.SshNet/Connection/HttpConnector.cs
+++ b/src/Renci.SshNet/Connection/HttpConnector.cs
@@ -35,25 +35,17 @@ namespace Renci.SshNet.Connection
         private const string HttpHeaderPattern = @"(?<fieldName>[^\[\]()<>@,;:\""/?={} \t]+):(?<fieldValue>.+)?";
 
 #if NET7_0_OR_GREATER
+        private static readonly Regex HttpResponseRegex = GetHttpResponseRegex();
+        private static readonly Regex HttpHeaderRegex = GetHttpHeaderRegex();
+
         [GeneratedRegex(HttpResponsePattern)]
-        private static partial Regex HttpResponseRegex();
+        private static partial Regex GetHttpResponseRegex();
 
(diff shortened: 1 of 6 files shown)


src/Renci.SshNet/Connection/HttpConnector.cs
```diff
@@ -29,18 +29,40 @@ namespace Renci.SshNet.Connection
...
+            return HttpResponseRe;
+        }
+
+        private static readonly Regex HttpHeaderRe = new Regex(HttpHeaderPattern, RegexOptions.Compiled);
+
+        private static Regex HttpHeaderRegex()
+        {
+            return HttpHeaderRe;
+        }
+#endif
```
Rob-Hague: Slightly more compact this way?

```suggestion
#if NET7_0_OR_GREATER
        private static readonly Regex HttpResponseRegex = GetHttpResponseRegex();
        private static readonly Regex HttpHeaderRegex = GetHttpHeaderRegex();

        [GeneratedRegex(HttpResponsePattern)]
        private static partial Regex GetHttpResponseRegex();

        [GeneratedRegex(HttpHeaderPattern)]
        private static partial Regex GetHttpHeaderRegex();
#else
        private static readonly Regex HttpResponseRegex = new Regex(HttpResponsePattern, RegexOptions.Compiled);
        private static readonly Regex HttpHeaderRegex = new Regex(HttpHeaderPattern, RegexOptions.Compiled);
#endif
```

mus65: Makes sense, pushed.

Rob-Hague (approved):
Thanks

PR #1402, opened 2024-05-17
fix flaky Sftp_BeginUploadFile test

mus65:
this test can randomly fail because it assumes that the callback has been called when the AsyncWaitHandle was set. But this is not necessarily the case because AsyncResult.SetAsCompleted does it the other way around.

example: https://ci.appveyor.com/project/drieseng/ssh-net/builds/49831002/job/1237d4lg46j22pf0

### DIFF
--- a/test/Renci.SshNet.IntegrationTests/SftpTests.cs
+++ b/test/Renci.SshNet.IntegrationTests/SftpTests.cs
@@ -132,8 +132,13 @@ namespace Renci.SshNet.IntegrationTests
                     using (var memoryStream = new MemoryStream(Encoding.ASCII.GetBytes(content)))
                     {
                         IAsyncResult asyncResultCallback = null;
+                        var callbackCalled = new ManualResetEvent(false);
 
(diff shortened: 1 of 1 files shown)


test/Renci.SshNet.IntegrationTests/SftpTests.cs
```diff
@@ -132,8 +132,13 @@ public void Sftp_BeginUploadFile()
                     using (var memoryStream = new MemoryStream(Encoding.ASCII.GetBytes(content)))
                     {
                         IAsyncResult asyncResultCallback = null;
+                        var callbackCalled = new ManualResetEvent(false);
```
Rob-Hague:
```suggestion
                        using var callbackCalled = new ManualResetEventSlim(false);
```

mus65: pushed

Rob-Hague (approved):
Thanks

PR #1403, opened 2024-05-17
fix flaky Connectivity Tests

mus65:
these two tests were not waiting for the interruption (all other usages of BreakConnections() did). This could cause random failures because e.g. the wrong exception is thrown.

example: https://ci.appveyor.com/project/drieseng/ssh-net/builds/49831433/job/4urb0upx63bvxcgu


mus65:
CI failure is yet another flaky test that should be fixed by #1404

Rob-Hague (approved):
Thanks

PR #1412, opened 2024-05-23
Remove unnecessary DNS lookup in Connect

Rob-Hague:
The library currently performs a DNS lookup of the desired host, takes the first returned IP address and connects to that. Instead, we can just pass the hostname down to System.Net.Sockets which will do the right thing, potentially trying multiple addresses if needed.


WojciechNagorski (approved):
LGTM!

PR #1414, opened 2024-05-25
fix flaky SFTP file time tests

mus65:
These asserts currently assume that DateTime.Now and the creation timestamp on the server happen within the same second.

Note that I didn't change the asserts below where an explicit timestamp is set. From my understanding, these should not be affected.

example: https://ci.appveyor.com/project/drieseng/ssh-net/builds/49850731/job/u6lk0cyipbxkp4jd


Rob-Hague:
Thanks, I'm pretty sure I just deleted these asserts in some similar tests but maybe I missed some. I'm out for the next week or so so may not come back until then

Rob-Hague:
Yeah, let's just delete them. The real asserts are a few lines lower.

mus65:
I removed them.

Rob-Hague (approved):
Thanks

PR #1417, opened 2024-05-28
Ref System.Memory for net462 and netstandard2.0

scott-xu:
Use `System.Buffers.Binary.BinaryPrimitives` instead of `Renci.SshNet.Common.Pack`

More can be done once this PR is merged, since `Span<T>` and `Memory<T>` is available in all SSH.NET target frameworks.


src/Renci.SshNet/Common/SshDataStream.cs
```diff
@@ -316,7 +323,6 @@ internal byte[] ReadBytes(int length)
             return data;
         }
 
-#if NETSTANDARD2_1 || NET6_0_OR_GREATER
```
Rob-Hague, line 319: Could you replace this method with [Stream.ReadExactly](https://learn.microsoft.com/dotnet/api/system.io.stream.readexactly) on NET7+, and keep a polyfill on other targets?

src/Renci.SshNet/Common/SshDataStream.cs
```diff
@@ -56,20 +56,47 @@ public bool IsEndOfData
...
+            {
+                var numRead = Read(sharedBuffer, 0, buffer.Length);
+
+                new ReadOnlySpan<byte>(sharedBuffer, 0, numRead).CopyTo(buffer);
+                return numRead;
+            }
+            finally
+            {
+                System.Buffers.ArrayPool<byte>.Shared.Return(sharedBuffer);
+            }
```
Rob-Hague: Let's remove the try-finally. It is generally considered not worthwhile - see here for some discussion https://github.com/dotnet/runtime/issues/71249#issuecomment-1165597563

Obviously it doesn't really matter here - just establishing a convention

src/Renci.SshNet/Common/SshDataStream.cs
Rob-Hague: actually, don't worry about it

src/Renci.SshNet/Common/SshDataStream.cs
Rob-Hague:
```suggestion
            
            var numRead = Read(sharedBuffer, 0, buffer.Length);

            sharedBuffer.AsSpan(0, numRead).CopyTo(buffer);
            
            System.Buffers.ArrayPool<byte>.Shared.Return(sharedBuffer);
            
            return numRead;
```

src/Renci.SshNet/Common/SshDataStream.cs
```diff
@@ -56,20 +56,47 @@ public bool IsEndOfData
...
+            var sharedBuffer = System.Buffers.ArrayPool<byte>.Shared.Rent(buffer.Length);
+            try
+            {
+                buffer.CopyTo(sharedBuffer);
+                Write(sharedBuffer, 0, buffer.Length);
+            }
+            finally
+            {
+                System.Buffers.ArrayPool<byte>.Shared.Return(sharedBuffer);
+            }
```
Rob-Hague:
```suggestion
            var sharedBuffer = System.Buffers.ArrayPool<byte>.Shared.Rent(buffer.Length);
            
            buffer.CopyTo(sharedBuffer);
            
            Write(sharedBuffer, 0, buffer.Length);
            
            System.Buffers.ArrayPool<byte>.Shared.Return(sharedBuffer);
```

Rob-Hague (approved):
Let's get spanning

PR #1419, opened 2024-05-30
Support creating Shell(Stream) without PTY

### LINKED ISSUE #1418: Support creating shell without PTY
OpenSSH client supports `-T` option to disable pseudo-terminal allocation. Can we add this feature to SSH.NET as well?
With this option, the output from remote does not contain control sequences.
It is not user-friendly but **program-friendly**. Program at client side does not need to parse/draw the screen with control sequences to get the correct output.

There's already `SshCommand` feature which is similar with this. But `SshCommand` does not support interactions. `SshCommand` runs and ends.

https://www.man7.org/linux/man-pages/man1/ssh.1.html
`-T      Disable pseudo-terminal allocation.`

scott-xu:
Fixes https://github.com/sshnet/SSH.NET/issues/1418

### DIFF
--- a/src/Renci.SshNet/ShellStream.cs
+++ b/src/Renci.SshNet/ShellStream.cs
@@ -20,6 +20,8 @@ namespace Renci.SshNet
     /// </summary>
     public class ShellStream : Stream
     {
+        private const int DefaultBufferSize = 1024;
+
         private readonly ISession _session;
         private readonly Encoding _encoding;
         private readonly IChannelSession _channel;
@@ -96,7 +98,7 @@ namespace Renci.SshNet
         /// <exception cref="SshException">The pseudo-terminal request was not accepted by the server.</exception>
         /// <exception cref="SshException">The request to start a shell was not accepted by the server.</exception>
         internal ShellStream(ISession session, string terminalName, uint columns, uint rows, uint width, uint height, IDictionary<TerminalModes, uint> terminalModeValues, int bufferSize)
-               : this(session, bufferSize, disablePTY: false)
+               : this(session, bufferSize, noTerminal: false)
         {
             try
             {
@@ -127,7 +129,7 @@ namespace Renci.SshNet
         /// <exception cref="SshException">The channel could not be opened.</exception>
         /// <exception cref="SshException">The request to start a shell was not accepted by the server.</exception>
         internal ShellStream(ISession session, int bufferSize)
-            : this(session, bufferSize, disablePTY: true)
+            : this(session, bufferSize, noTerminal: true)
         {
             try
             {
@@ -150,10 +152,14 @@ namespace Renci.SshNet
         /// </summary>
         /// <param name="session">The SSH session.</param>
         /// <param name="bufferSize">The size of the buffer.</param>
-        /// <param name="disablePTY">Disables pseudo terminal allocation or not.</param>
+        /// <param name="noTerminal">Disables pseudo terminal allocation or not.</param>
         /// <exception cref="SshException">The channel could not be opened.</exception>
-        private ShellStream(ISession session, int bufferSize, bool disablePTY)
+        private ShellStream(ISession session, int bufferSize, bool noTerminal)
         {
+            if (bufferSize == -1)
+            {
+                bufferSize = DefaultBufferSize;
+            }
 #if NET8_0_OR_GREATER
             ArgumentOutOfRangeException.ThrowIfNegativeOrZero(bufferSize);
 #else
@@ -177,7 +183,7 @@ namespace Renci.SshNet
             _readBuffer = new byte[bufferSize];
             _writeBuffer = new byte[bufferSize];
 
-            _noTerminal = disablePTY;
+            _noTerminal = noTerminal;
         }
 
         /// <summary>
--- a/src/Renci.SshNet/ShellStream.cs
+++ b/src/Renci.SshNet/ShellStream.cs
@@ -96,30 +96,8 @@ namespace Renci.SshNet
         /// <exception cref="SshException">The pseudo-terminal request was not accepted by the server.</exception>
         /// <exception cref="SshException">The request to start a shell was not accepted by the server.</exception>
         internal ShellStream(ISession session, string terminalName, uint columns, uint rows, uint width, uint height, IDictionary<TerminalModes, uint> terminalModeValues, int bufferSize)
+               : this(session, bufferSize, disablePTY: false)
         {
-#if NET8_0_OR_GREATER
-            ArgumentOutOfRangeException.ThrowIfNegativeOrZero(bufferSize);
-#else
-            if (bufferSize <= 0)
-            {
-                throw new ArgumentOutOfRangeException(nameof(bufferSize));
-            }
-#endif
-
-            _encoding = session.ConnectionInfo.Encoding;
-            _session = session;
-            _carriageReturnBytes = _encoding.GetBytes("\r");
-            _lineFeedBytes = _encoding.GetBytes("\n");
-
-            _channel = _session.CreateChannelSession();
-            _channel.DataReceived += Channel_DataReceived;
-            _channel.Closed += Channel_Closed;
-            _session.Disconnected += Session_Disconnected;
-            _session.ErrorOccured += Session_ErrorOccured;
-
-            _readBuffer = new byte[bufferSize];
-            _writeBuffer = new byte[bufferSize];
-
             try
             {
                 _channel.Open();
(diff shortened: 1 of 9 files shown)


src/Renci.SshNet/ShellStream.cs
```diff
@@ -119,25 +177,7 @@ internal ShellStream(ISession session, string terminalName, uint columns, uint r
             _readBuffer = new byte[bufferSize];
             _writeBuffer = new byte[bufferSize];
 
-            try
```
Rob-Hague, line 122: Keep this here? Don't think they need to be in separate constructors

src/Renci.SshNet/ShellStream.cs
```diff
@@ -848,7 +888,7 @@ public override void Write(byte[] buffer, int offset, int count)
         /// <exception cref="ObjectDisposedException">The stream is closed.</exception>
         public void WriteLine(string line)
         {
-            Write(line + "\r");
+            Write(line + (_disablePTY ? "\n" : "\r"));
```
Rob-Hague: Why?

src/Renci.SshNet/SshClient.cs
```diff
@@ -313,7 +331,7 @@ public Shell CreateShell(Stream input, Stream output, Stream extendedOutput, str
         /// <exception cref="SshConnectionException">Client is not connected.</exception>
         public Shell CreateShell(Stream input, Stream output, Stream extendedOutput)
         {
-            return CreateShell(input, output, extendedOutput, string.Empty, 0, 0, 0, 0, terminalModes: null, 1024);
+            return CreateShell(input, output, extendedOutput, 1024);
```
Rob-Hague: This looks like a behavioural breaking change. Not sure if we should take it

src/Renci.SshNet/SshClient.cs
```diff
@@ -353,6 +371,37 @@ public Shell CreateShell(Encoding encoding, string input, Stream output, Stream
...
+        /// <param name="encoding">The encoding to use to send the input.</param>
+        /// <param name="input">The input.</param>
+        /// <param name="output">The output.</param>
+        /// <param name="extendedOutput">The extended output.</param>
+        /// <param name="bufferSize">Size of the internal read buffer.</param>
+        /// <returns>
+        /// Returns a representation of a <see cref="Shell" /> object.
+        /// </returns>
+        /// <exception cref="SshConnectionException">Client is not connected.</exception>
+        public Shell CreateShell(Encoding encoding, string input, Stream output, Stream extendedOutput, int bufferSize)
```
Rob-Hague: This is the 8th overload of `CreateShell`, let's leave this one out for now

src/Renci.SshNet/SshClient.cs
```diff
@@ -450,6 +499,21 @@ public ShellStream CreateShellStream(string terminalName, uint columns, uint row
...
 
+        /// <summary>
+        /// Creates the shell stream.
+        /// </summary>
+        /// <param name="bufferSize">The size of the buffer.</param>
+        /// <returns>
+        /// The created <see cref="ShellStream"/> instance.
+        /// </returns>
+        /// <exception cref="SshConnectionException">Client is not connected.</exception>
+        public ShellStream CreateShellStream(int bufferSize)
```
Rob-Hague: Let's make it an optional argument so we can give it a default value when it is not supplied (-1 means default)

```suggestion
        public ShellStream CreateShellStream(int bufferSize = -1)
```

test/Renci.SshNet.IntegrationTests/SshTests.cs
```diff
@@ -77,6 +78,23 @@ public void Ssh_ShellStream_Exit()
...
         }
 
+        [TestMethod]
+        public void Ssh_CreateShellStream_WithoutPseudoTerminal()
+        {
+            using (var client = new SshClient(_connectionInfoFactory.Create()))
+            {
+                client.Connect();
+
+                using (var shellStream = client.CreateShellStream(bufferSize: 1024))
```
Rob-Hague: This test also passes if I replace this line with `CreateShellStream("term", 80, 80, 80, 80, 1024)`. Can you make it only pass with the non-terminal-allocating version?

test/Renci.SshNet.IntegrationTests/SshTests_TTYDisabled.cs
```diff
@@ -0,0 +1,122 @@
...
+                {
+                    var shell = client.CreateShell(input, output, extOutput);
+
+                    shell.Start();
+
+                    var inputWriter = new StreamWriter(input, Encoding.ASCII, 1024);
+                    inputWriter.WriteLine("echo $PATH");
+
+                    var outputReader = new StreamReader(output, Encoding.ASCII, false, 1024);
+                    Console.WriteLine(outputReader.ReadToEnd());
```
Rob-Hague: Assert on something instead?

test/Renci.SshNet.IntegrationTests/SshTests.cs
```diff
@@ -171,6 +189,32 @@ public void Ssh_CreateShell()
...
+        {
+            using (var client = new SshClient(_connectionInfoFactory.Create()))
+            {
+                client.Connect();
+
+                using (var input = new MemoryStream())
+                using (var output = new MemoryStream())
+                using (var extOutput = new MemoryStream())
+                {
+                    var shell = client.CreateShell(input, output, extOutput);
```
Rob-Hague: Also here, make it only pass without pseudo terminal

Rob-Hague (commented):
Not sure just yet whether the behaviour should be differentiated from the default by more than just an overload, i.e. whether they should be methods with different names e.g. `CreateShellNoTerminal`

src/Renci.SshNet/ShellStream.cs
scott-xu: > Keep this here? Don't think they need to be in separate constructors

You want to remove the constructor `private ShellStream(ISession session, int bufferSize, bool disablePTY)` and copy it's content in the two internal constructors?

src/Renci.SshNet/ShellStream.cs
scott-xu: By default, the terminal driver [translates carriage return to line feed on input](https://www.man7.org/linux/man-pages/man3/termios.3.html#:~:text=ICRNL%20%20Translate%20carriage%20return%20to%20newline%20on%20input%20(unless%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20IGNCR%20is%20set).)

src/Renci.SshNet/SshClient.cs
scott-xu: Previously, it will swallow the exception if failed to send pseudo terminal request when start `Shell`. (Note it will throw exception when instantiate `ShellStream`)

src/Renci.SshNet/SshClient.cs
scott-xu: Or rename it to `CreateShellNoTerminal`?

src/Renci.SshNet/SshClient.cs
scott-xu: There are many methods take `bufferSize` argument. Should we change them all?
image

src/Renci.SshNet/SshClient.cs
scott-xu: Or `CreateShellWithoutTerminal`?

test/Renci.SshNet.IntegrationTests/SshTests.cs
scott-xu: Sorry I don't understand.
This is a integration test and we want to make sure "user can create a shell stream without allocating pseudo terminal and can successfully send/receive".
We can add another integration test to make sure "user can create a shell stream with pseudo terminal and can successfully send/receive", which I believe is already covered by test method `Ssh_ShellStream_Exit`
They don't affect each other, do they?

Rob-Hague: What I mean is, presumably you have a use case for this functionality, but so far I don't see any demonstration of its use, and how its behaviour might differ to the terminal-allocating version (other than "it does not fail when the server disables PTY").

You said in the issue that this functionality can be used to exclude control sequences. Can you write a test to prove that?

edit: I'm not so familiar with this functionality, that's why it would be nice to see an example

src/Renci.SshNet/ShellStream.cs
Rob-Hague: Ok, thanks, can you put that in a comment

src/Renci.SshNet/SshClient.cs
Rob-Hague: That's good. What I mean is that this now will not allocate the pseudo terminal where previously it did, and it would not be nice to cause a behaviour change there (if the behaviour is indeed different)

src/Renci.SshNet/SshClient.cs
Rob-Hague: Yes, it might be safest if the new functionality is exposed via `CreateShellNoTerminal` and `CreateShellStreamNoTerminal` just to make a clear distinction.

The overload on this particular line does not seem very useful to me, but if you use it you can keep it.

src/Renci.SshNet/SshClient.cs
Rob-Hague: No I don't think any others are worth changing, just this new API. I'm guessing most of these overloads were added before optional arguments existed

src/Renci.SshNet/ShellStream.cs
Rob-Hague: Ok, I see the parameters for SendPseudoTerminalRequest are not supplied to this constructor. It's fine how it is

src/Renci.SshNet/SshClient.cs
scott-xu: Then other methods will throw exception if buffer size is -1, whereas this method will treat -1 as 1024. Is that okay for you?

test/Renci.SshNet.IntegrationTests/SshTests.cs
scott-xu: Okay I get what you mean. Will update the tests.

src/Renci.SshNet/SshClient.cs
Rob-Hague: You could add

```c# 
if (bufferSize == -1)
{
    bufferSize = 1024;
}
```

before the `ArgumentOutOfRangeException.ThrowIfNegativeOrZero` in the constructor to make it consistent for all overloads

scott-xu: How about making the optional `bufferSize` default to 1024 rather than -1?

scott-xu:
```C#
Shell CreateShellNoTerminal(Stream input, Stream output, Stream extendedOutput, int bufferSize = 1024)

ShellStream CreateShellStreamNoTerminal(int bufferSize = 1024)
```

Rob-Hague: -1 is better because then we can change the default in the future as an implementation detail. See StreamWriter https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/System.Private.CoreLib/src/System/IO/StreamWriter.cs#L93

scott-xu: Ok, that makes sense. Thanks.

Rob-Hague (approved):
Thanks

PR #1423, opened 2024-06-05
A couple of changes/fixes in SshCommand

### LINKED ISSUE #95: ExitSignalName
Hello.

Can You tell me what happened to property "ExitSignalName" in Renci.SshNet.SshCommand ?
Where can i find a source code where it was still there ?

Best regards,
Mieszko

Rob-Hague:
1. in `CancelAsync`, we were sending `exit-signal` channel request, when in fact we should be sending `signal` and the server will respond with `exit-signal`. This explains why we did not see any response or channel close in #1345 (@zeotuan fyi). Add a corresponding `string? ExitSignal` property which has a value when the server sends one. (closes #95 - seems like it existed once upon a time).
2. Previously, we were setting the `IAsyncResult.WaitHandle` only when the command had completed successfully, or otherwise once `EndExecute` had been called. But we should be setting it as soon as the command completes whether successfully, with an error or via cancellation. This refactors the logic by removing the extra wait handles and consolidating the completion logic onto the `IAsyncResult.WaitHandle`. With this refactoring the logic intentionally looks (with some squinting) similar to a `Task`. Adding `ExecuteAsync` should only be a small step from here.
3. Take a breaking change on `ExitStatus` by changing the return type from `int` to `int?`, in order to distinguish between an exit status of 0 and when no exit status has been returned (because for example, an exit signal was returned instead). This is the second commit.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -688,7 +688,7 @@ dotnet_diagnostic.CA1852.severity = none
 # CA1859: Change return type for improved performance
 #
 # By default, this diagnostic is only reported for private members.
-dotnet_code_quality.CA1859.api_surface = all
+dotnet_code_quality.CA1859.api_surface = private,internal
 
 # CA2208: Instantiate argument exceptions correctly
 # https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2208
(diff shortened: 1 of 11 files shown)


WojciechNagorski:
Is it ready for review?

Rob-Hague:
Yes, thanks

WojciechNagorski (approved):

Rob-Hague:
Thanks

PR #1426, opened 2024-06-14
Add SshCommand.ExecuteAsync

### LINKED ISSUE #650: SshCommand: OutputStream & ExtendedOutputStream should only return 0 when channel is closed

Rob-Hague:
After the previous change (#1423), this basically entails swapping out the IAsyncResult for a TaskCompletionSource and hooking up the cancellation/timeout logic.

As with the prior Begin/End implementation, the initiation of the command (SendExecRequest) happens synchronously, so there's a bit of room for improvement there, but otherwise it is the Task-based async that we know and like.

I chose to make it void (Task)- returning instead of string like in the existing overloads, so that OutputStream is not automatically consumed (and encoded as a string) when that may not be desired. As in #650, I was initially considering changing the other overloads to be void-returning as well, but decided that it was not worth the break since most people will probably want to change over to ExecuteAsync anyway.

closes #650 

supersedes a couple of PRs:
closes #937 
closes #1343


test/Renci.SshNet.IntegrationTests/SshClientTests.cs
```diff
@@ -91,29 +91,6 @@ public void Send_InputStream_to_Command_OneByteAtATime()
             }
         }
 
-        [TestMethod]
-        public void Send_InputStream_to_Command_InputStreamNotInUsingBlock_StillWorks()
```
Rob-Hague, line 95: I deleted this test because now it hangs (executes indefinitely). It was nice that it didn't hang before, but I don't think that behaviour is compatible with a TAP implementation. If you are sending a stream of data to `cat` without telling it when you have finished, and you aren't using any cancellation or timeout feature, then it is surely only logical that the command would keep executing forever.

WojciechNagorski (approved):

PR #1427, opened 2024-06-15
fix build with .NET 9.0 SDK

mus65:
SSH.NET is currently not buildable with .NET 9 because of newly introduced warnings CA1872 and CA2022, see separate commits.


Rob-Hague (approved):
Thanks

One day we'll stop every diagnostic being an error ...

PR #1441, opened 2024-07-09
Allow empty softwareversion string (#822)

### LINKED ISSUE #822: Exception thrown when the server SSH identification is missing its"softwareversion" part
We have actually run into a server in the wild which only provides an identification string of "SSH-2.0-" and omits the "softwareversion" part required by the standard. This leads to SSH.NET throwing an exception

`Renci.SshNet.Common.SshConnectionException: The server response contains a null character at position 0x0000000B:

  00000000  53 53 48 2D 32 2E 30 2D 0D 0A 00                 SSH-2.0-...

A server must not send a null character before the Protocol Version Exchange is complete.`

I've seen that there had been some discussion about less strict parsing in #749, with a suggestion of a potential strict/lenient property to allow for non-conforming servers that do not provide the software version part. Any thoughts on actually going forward with such a leniency option?

We've contacted the server operator about their non-conformance, but no idea if they have any interest in fixing this as apparently other clients are more forgiving about a missing software version. FileZilla for example connects without issues.

doerig:


Rob-Hague (approved):
Great, thanks

PR #1442, opened 2024-07-11
Drop legacy algorithms part 1

Rob-Hague:
This drops some of the algorithms long-considered legacy/insecure.

The idea is both to improve the theoretical security of the library by not offering these algorithms, and to improve the practical security of the library by not having hand-written, barely tested crypto code.

The overarching goal is for the library to have minimal exposure to crypto implementation, relying firstly on the .NET base libraries, and secondly on third-party providers, such as BouncyCastle.

This change covers deleting the cipher algorithms arcfour, blowfish, twofish, cast. It covers deleting the MD5-based and truncated HMAC algorithms.

These were all disabled in OpenSSH server (sshd) in 2014[^1]:

> sshd(8): The default set of ciphers and MACs has been altered to
> remove unsafe algorithms. In particular, CBC ciphers and arcfour*
> are disabled by default.
>
> The full set of algorithms remains available if configured
> explicitly via the Ciphers and MACs sshd_config options.

and in the client in 2016[^2]:

> This release disables a number of legacy cryptographic algorithms
> by default in ssh:
>
>  * Several ciphers blowfish-cbc, cast128-cbc, all arcfour variants
>    and the rijndael-cbc aliases for AES.
>
>  * MD5-based and truncated HMAC algorithms.
>
> These algorithms are already disabled by default in sshd.

This change also drops PKCS5Padding, which is a line-for-line copy of PKCS7Padding, and StreamCipher, which is now unused (and useless anyway).

[^1]: https://www.openssh.com/txt/release-6.7
[^2]: https://www.openssh.com/txt/release-7.2


WojciechNagorski:
Maybe we should move them to the SSH.NET.Unsafe nuget package. If anyone wanted to use them, it would be possible

Rob-Hague:
I think these are so old that it's not worth any extra effort unless there is some unexpected demand for it. It seems reasonable for someone to stay on an old version of SSH.NET if they want to use old algorithms

WojciechNagorski:
I agree 👍

WojciechNagorski (approved):

WojciechNagorski:
I approved.

Rob-Hague:
Thanks

PR #1450, opened 2024-07-21
[AesGcm] Falls back to use BouncyCastle if BCL doesn't support

scott-xu:


scott-xu:
This PR's code structure is very similar with `AesCipher`'s implementation as well. Ready to review and merge @Rob-Hague @WojciechNagorski

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.BclImpl.cs
```diff
@@ -0,0 +1,72 @@
...
+                    _aesGcm.Decrypt(nonce, cipherText, tag, output, associatedData);
+                }
+#if NET8_0_OR_GREATER
+                catch (AuthenticationTagMismatchException)
+#else
+                catch (CryptographicException ex) when (ex.Message == "The computed authentication tag did not match the input authentication tag.")
+#endif
+                {
+                    throw new SshConnectionException("MAC error", DisconnectReason.MacError);
+                }
```
Rob-Hague: This seems pretty fragile, a) catching an undocumented exception and b) catching based on the (unlocalized) message text. I suggest just passing the caught exception as the inner exception

```suggestion
                catch (CryptographicException ex)
                {
                    throw new SshConnectionException("MAC error", DisconnectReason.MacError, ex);
                }
```

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.BouncyCastleImpl.cs
```diff
@@ -0,0 +1,52 @@
...
+
+                var plainTextLength = _cipher.ProcessBytes(input, cipherTextOffset, cipherTextLength + _tagSize, output, plainTextOffset);
+                try
+                {
+                    _ = _cipher.DoFinal(output, plainTextLength);
+                }
+                catch (InvalidCipherTextException)
+                {
+                    throw new SshConnectionException("MAC error", DisconnectReason.MacError);
+                }
```
Rob-Hague:
```suggestion
                catch (InvalidCipherTextException ex)
                {
                    throw new SshConnectionException("MAC error", DisconnectReason.MacError, ex);
                }
```

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.BouncyCastleImpl.cs
```diff
@@ -0,0 +1,52 @@
...
+            {
+                _key = key;
+                _tagSize = tagSize;
+                _cipher = new GcmBlockCipher(new AesEngine());
+            }
+
+            public override void Encrypt(byte[] nonce, byte[] input, int plainTextOffset, int plainTextLength, int associatedDataOffset, int associatedDataLength, byte[] output, int cipherTextOffset)
+            {
+                var parameters = new AeadParameters(new KeyParameter(_key), _tagSize * 8, nonce, input.Take(associatedDataOffset, associatedDataLength));
+                _cipher.Init(forEncryption: true, parameters);
```
Rob-Hague: Seems a shame we need to initialise new parameters for each packet. Can the `KeyParameter` argument be a field? And is there a way to use `GcmCipher.ProcessAadBytes` to avoid the copy of the associated data?

Rob-Hague: This applies to the line above (`new AeadParameters(...`)

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.cs
```diff
@@ -12,10 +10,15 @@ namespace Renci.SshNet.Security.Cryptography.Ciphers
...
+    internal sealed partial class AesGcmCipher : SymmetricCipher, IDisposable
     {
+        private const int PacketLengthFieldLength = 4;
         private readonly byte[] _iv;
-        private readonly AesGcm _aesGcm;
+#if NET6_0_OR_GREATER
+        private readonly Impl _impl;
+#else
+        private readonly BouncyCastleImpl _impl;
+#endif
```
Rob-Hague, line 22: nit, looks like it can be simplified

```suggestion
        private readonly Impl _impl;
```

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.BclImpl.cs
scott-xu: `AuthenticationTagMismatchException` is a documented exception. See https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.authenticationtagmismatchexception?view=net-8.0

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.cs
scott-xu: Code analysis doesn't like your suggestion

scott-xu: See https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1859

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.BouncyCastleImpl.cs
scott-xu: > And is there a way to use `GcmCipher.ProcessAadBytes` to avoid the copy of the associated data?

The limitation comes from BouncyCastle. The `AeadParameters`'s constructor doesn't take associated data with offset and length.

scott-xu: Oh I see what you mean. Yes, we can use `GcmCipher.ProcessAadBytes`

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.BclImpl.cs
Rob-Hague: It is not documented on the Encrypt/Decrypt methods (rightly or wrongly): https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.aesgcm.decrypt?view=net-8.0

Since AuthenticationTagMismatchException already derives from CryptographicException, I'm not sure there's any reason for the code to differ between .NET6/8

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.cs
Rob-Hague: Ok, as the classes are sealed then there is unlikely to be a difference in performance, but there's no need to fight the code analysis

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.BclImpl.cs
scott-xu: Do we want to distinguish between Mac Error and other error? See https://github.com/dotnet/runtime/issues/67082

Rob-Hague: I don't think so - it's fatal for the connection either way

scott-xu: Then it could fail with "Mac Error" where the underlying error is not "Mac Error".

Rob-Hague (approved):

PR #1451, opened 2024-07-21
Clean up Abstractions / use HashData

mus65:
- replace usages of GenerateRandom(byte[]) with GenerateRandom(int)
- remove Create() functions and make use of static HashData methods on .NET 6+, see https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1850
- Remove unused ReflectionAbstraction


Rob-Hague:
I wrote a very similar branch just the other day 🙂 https://github.com/sshnet/SSH.NET/compare/develop...Rob-Hague:SSH.NET:hashoneshot

May I suggest taking the `CryptoAbstraction.Hash[..]` pattern from that branch. Seems like the ideal place for the ifdefs

mus65:
ha, bad timing. :smile:

yeah, that's better, PrivateKeyFile.cs was kind of messy with all the ifdefs. I cherry-picked your changes and added you as co-auhor.

Rob-Hague (approved):

Issue #1406, opened 2024-05-17
SSH.Net 2024.0.0 works for UWP but does not work for Android and iOS


Karteeka18:
### Discussed in https://github.com/sshnet/SSH.NET/discussions/1405

<div type='discussions-op-text'>

<sup>Originally posted by **Karteeka18** May 17, 2024</sup>
We have integrated SSH.Net 2024.0.0 from Nuget for Xamarin UWP, Xamarin. IOS and Mono Android.  UWP works fine. BUT iOS and Android are throwing exceptions. Any ideas or suggestions would be helpful.
Visual Studio 2022 C#

              using (SftpClient sftp = new SftpClient(host, port, user, pass))
                {
                    sftp.Connect();

                    if (!sftp.Exists(path))
                        ok = false;
                    sftp.Disconnect();
                }

Throws an Exception on sftp.Connect().

**iOS Exception** 
{System.NotImplementedException: The method or operation is not implemented.   at System.Security.Cryptography.ECDsa.Create (System.Security.Cryptography.ECParameters parameters) [0x00000] in /Library/Frameworks/Xamarin.iOS.framework/Versions/Current/src/Xamari…} </div>

Rob-Hague:
It's not clear to me which runtime version your app is using and which target platform for the library.

You would get a definitive answer by including that in an issue in the https://github.com/dotnet/runtime repository, although I'm afraid the answer might simply be that ECDsa is not supported on your chosen version/platform.

PR #1220, opened 2023-10-21
appveyor test

Rob-Hague:

PR #1456, opened 2024-07-25
Enable all net48 integration tests and stabilise tests

scott-xu:


Rob-Hague:
We can merge this one next assuming it passes. No need to create and resolve conflicts in the other branches

scott-xu:
Deal 🥂

scott-xu:
`public static System.Security.Cryptography.ECDsa Create(System.Security.Cryptography.ECParameters parameters)` applies to net47 and above.

See https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.ecdsa.create?view=net-8.0#system-security-cryptography-ecdsa-create(system-security-cryptography-ecparameters)

scott-xu:
Unfortunately, Mono doesn't implement `ECDsa`. See https://github.com/mono/mono/blob/main/mcs/class/referencesource/System.Core/System/Security/Cryptography/ECDsa.cs#L32

WojciechNagorski:
So we can run them on Windows. What do you think?

Rob-Hague:
> Unfortunately, Mono doesn't implement `ECDsa`. See https://github.com/mono/mono/blob/main/mcs/class/referencesource/System.Core/System/Security/Cryptography/ECDsa.cs#L32

It is the reason behind #1406 

> So we can run them on Windows. What do you think?

I tried a while back. The problem is running a Linux container on a Windows host, which doesn't seem so easy on CI machines (the way to do it is WSL2 AFAIK but most Windows server images do not come with WSL2). But I'd welcome anyone to try because I'd very much like to see it happen.

Rob-Hague:
It might be possible now that Windows Server 2022 includes WSL2. The problems I ran into were seemingly related to LCOW (https://github.com/sshnet/SSH.NET/pull/1220#issuecomment-1773930792). I wouldn't know if it is possible to run docker via WSL2 rather than LCOW in appveyor (I don't really know much about this stuff)

Rob-Hague:
I think there's some work involved here to make sure we don't introduce more unstable tests

scott-xu:
> > Unfortunately, Mono doesn't implement `ECDsa`. See https://github.com/mono/mono/blob/main/mcs/class/referencesource/System.Core/System/Security/Cryptography/ECDsa.cs#L32
> 
> It is the reason behind #1406
> 

We can consider using BouncyCastle as a fallback when in Mono. (https://www.mono-project.com/docs/faq/technical/#how-can-i-detect-if-am-running-in-mono)

Then we can split the code into 3 implementations.
1. Bcl implementation (.NET Framework 4.7+ and not Mono)
2. Cng implementation (.NET Framework 4.6.2 and not Mono)
3. BouncyCastle implementation (.NET Framework and Mono)

I can have a try when I have time.

scott-xu:
It seems not easy to split `EcdsaKey` into three implementations because it is public already with `ECDsaCng Ecdsa` and `CngAlgorithm HashAlgorithm` properties when target to .NET Framework.

Rob-Hague:
I kind of expected mono to use a .NET Standard build, but I guess it can use a .NET Framework build as well. The error in #1406 comes from calling `ECDsa.Create(ECParameters)` which only occurs #if !NET_FRAMEWORK.

Anyway I think it's fine to tweak the public members at this level if needed, especially if it doesn't work in the first place.

I'll run the CI on this a couple more times to check for any more misbehaving tests

test/Renci.SshNet.IntegrationTests/SshTests.cs
```diff
@@ -655,17 +655,32 @@ public void Ssh_RemotePortForwarding()
             var endpoint1 = new IPEndPoint(ipv4HostAddress, 10000);
             var endpoint2 = new IPEndPoint(ipv4HostAddress, 10001);
 
+            var isBytesReceivedOnLister1 = false;
+            var isBytesReceivedOnLister2 = false;
```
Rob-Hague:
```suggestion
            var isBytesReceivedOnListener1 = false;
            var isBytesReceivedOnListener2 = false;
```

scott-xu:
Any idea to stabilize this test? @Rob-Hague 
```
  Failed Test_ExecuteAsync_CancellationToken [1 s]
  Error Message:
   Assert.AreEqual failed. Expected:<TERM>. Actual:<(null)>. 
  Stack Trace:
     at Renci.SshNet.IntegrationTests.OldIntegrationTests.SshCommandTest.Test_ExecuteAsync_CancellationToken() in /home/appveyor/projects/ssh-net/test/Renci.SshNet.IntegrationTests/OldIntegrationTests/SshCommandTest.cs:line 139
```
https://ci.appveyor.com/project/drieseng/ssh-net/builds/50322490/job/yo6pooiew5j86wcw

The 500ms wait may not enough in unit test.
See https://github.com/sshnet/SSH.NET/blob/develop/src/Renci.SshNet/SshCommand.cs#L446

Rob-Hague:
Hmm, I would not expect the 500ms to be relevant here because we know OpenSSH implements signals. I think we can ignore that one and see if it happens more often

Rob-Hague (approved):
Thanks

PR #1459, opened 2024-07-27
Stabilise tests

scott-xu:


Rob-Hague:
Put it into #1456 instead?

PR #1462, opened 2024-08-04
Revert mono/.NET Framework integration tests

Rob-Hague:
This combination is not a worthwhile use of time

Ref https://github.com/sshnet/SSH.NET/pull/1458#issuecomment-2267458881 and #1456

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,14 +21,12 @@ for:
     - echo build
     - dotnet build -f net8.0 -c Debug test/Renci.SshNet.Tests/Renci.SshNet.Tests.csproj
     - dotnet build -f net8.0 -c Debug test/Renci.SshNet.IntegrationTests/Renci.SshNet.IntegrationTests.csproj
-    - dotnet build -f net48 -c Debug test/Renci.SshNet.IntegrationTests/Renci.SshNet.IntegrationTests.csproj
 
   test_script:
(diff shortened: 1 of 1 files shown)


scott-xu (approved):
It's okay for me.
However, regarding https://github.com/sshnet/SSH.NET/pull/1461, we have to run .NET Framework tests (can use `--filter Name~ECDsa`) at Linux with Mono to verify if it fixes https://github.com/sshnet/SSH.NET/issues/1406.

Rob-Hague:
Ok but note that Xamarin is now EOL, so it also may not be worth your time. And sockets can't run in the browser so mono/WebAssembly is not relevant to us - I don't think there are any officially supported, commonly used scenarios under mono which are relevant to us

WojciechNagorski (approved):

scott-xu:
Can we keep **Ecdh** and **Zlib**? We need to make sure .NET Framework specific code to be tested.
`--filter "Name~Ecdh|Name~Zlib"`

We don't need to worry about **ChaCha20Poly1305** as there's no .NET Framework specific code for it. It is covered by .NET 8.0 integration test.

Rob-Hague:
If it doesn't cause any issues, then fine. I'll try it on the DSA branch. Otherwise it is easy enough to run them properly locally

PR #1465, opened 2024-08-06
appveyor test

Rob-Hague:


Rob-Hague:
This image is probably what we would want but it's not part of the free tier: https://help.appveyor.com/discussions/questions/61307-windows-server-2022-vms

I did ask about pricing but no response, which I can only assume means "quite a lot" since there is no price listed at https://ci.appveyor.com/billing

PR #1468, opened 2024-08-10
Drop net7.0 target

Rob-Hague:
.NET 7 is EOL since May. The only .NET 7 features we use are `ObjectDisposedException.ThrowIf` (moved to a throw helper) and some newer regex features.

This feels a bit weird, but I suppose it is the expected course of action.


WojciechNagorski (dismissed):
Can we make a release without this and merge it later for next release?

Rob-Hague:
Sure

drieseng:
Does it hurt us to keep supporting .NET 7.0? Do we want to be so agressive in removing support for TFMs that are EOL?
+1 for all other changes.

Rob-Hague:
It is obviously not a great hinderance atm, as there is not much net7.0-specific code. That's also a reason to drop it without ill effect.

IMO it is not much more nuanced than: if Microsoft don't support it, then why should we?

Anyway, let's do another release and re-evaluate for the one after.

mus65:
Also note that dropping net7.0 doesn't actually prevent net7.0 applications from using SSH.NET . They will just fall back to the ~~netstandard2.1~~ net6.0 assembly, which will probably work fine.

mus65:
Might want to drop net6.0 as well since it's also EOL. Also netstandard2.1 is imho not worth targeting anymore. I've seen very few libraries targeting it and even MS libraries don't.

drieseng (approved):

Rob-Hague:
will do a separate change for net6.0

PR #1480, opened 2024-08-30
Add .NET 9 target

mus65:
I had to disable a SonarSource analyzer and made use of the new `Lock` type, see separate Commits.

CI is expected to fail because of missing .NET 9 binaries in AppVeyor. Build and tests pass locally.

Can be merged when .NET 9 final is released and AppVeyor supports it.


Rob-Hague:
Can we have a basic Lock polyfill to save on the #ifdefs? e.g.

```c#
#if !NET9_0_OR_GREATER

namespace Renci.SshNet.Common;

internal sealed class Lock
{
    public bool TryEnter() => Monitor.TryEnter(this);
    public void Exit() => Monitor.Exit(this);
}

```

mus65:
Makes sense, pushed.

drieseng:
@mus65, you'll also want to update **global.json**.

.editorconfig
```diff
@@ -183,6 +183,9 @@ dotnet_diagnostic.S2971.severity = none
 # This is rather harmless.
 dotnet_diagnostic.S3218.severity = none
 
+# S3236: Remove this argument from the method call; it hides the caller information.
+dotnet_diagnostic.S3236.severity = none
```
drieseng, line 199: Where was this (falsely) reported?

appveyor.yml
```diff
@@ -19,15 +19,15 @@ for:
           
   build_script:
     - echo build
-    - dotnet build -f net8.0 -c Debug test/Renci.SshNet.Tests/Renci.SshNet.Tests.csproj
-    - dotnet build -f net8.0 -c Debug test/Renci.SshNet.IntegrationTests/Renci.SshNet.IntegrationTests.csproj
+    - dotnet build -f net9.0 -c Debug test/Renci.SshNet.Tests/Renci.SshNet.Tests.csproj
```
drieseng: Do we want to switch from .NET 8.0 (LTS) to .NET 9.0 (STS, currently still in preview)?
Does AppVeyor even support .NET 9.0?

test/Renci.SshNet.AotCompatibilityTestApp/Renci.SshNet.AotCompatibilityTestApp.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net8.0</TargetFramework>
+    <TargetFramework>net9.0</TargetFramework>
```
drieseng: Do we want this?

test/Renci.SshNet.Benchmarks/Renci.SshNet.Benchmarks.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net8.0</TargetFramework>
+    <TargetFramework>net9.0</TargetFramework>
```
drieseng, line 5: Do we want this?

test/Renci.SshNet.IntegrationBenchmarks/Renci.SshNet.IntegrationBenchmarks.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net8.0</TargetFramework>
+    <TargetFramework>net9.0</TargetFramework>
```
drieseng, line 5: Do we want this?

test/Renci.SshNet.IntegrationTests/Renci.SshNet.IntegrationTests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net48;net8.0</TargetFrameworks>
+    <TargetFrameworks>net48;net9.0</TargetFrameworks>
```
drieseng: Do we want this?

drieseng:
@Rob-Hague, are planning on merging this before the release of .NET 9.0? In theory, there may even be API changes - I'm thinking of the **Lock** class - between now and the release. It's highly unlikely though.

.editorconfig
mus65: I mentioned this in the commit (should have written it in the PR in hindsight). This causes false positives with `Debug.Assert` when a custom message is supplied. This is only a problem since .NET 9 because of this change: https://github.com/dotnet/core/blob/main/release-notes/9.0/preview/preview7/libraries.md#debugassert-now-reports-assert-condition-by-default

appveyor.yml
mus65: > Do we want to switch from .NET 8.0 (LTS) to .NET 9.0 (STS, currently still in preview)?

Imho it makes sense to test against the latest .NET to catch any regressions. Would even be better to test against all of them, but that's probably not worth the CI overhead. That's up to you.

I didn't want this to be merged while .NET 9 is still in preview, that's why it's still draft.

> Does AppVeyor even support .NET 9.0?

it doesn't. I'm just hoping that it supports it when .NET 9 final is released. If not, the SDK could probably be installed manually somehow.

mus65:
@drieseng @Rob-Hague I updated global.json . I had to reference RC1 explicitly though, so I have to change it when final is released.

I left it a draft because it's still preview and wouldn't expect it to be merged before .NET 9 final. RC1 is already released, so I wouldn't expect any more API changes.

Rob-Hague:
Yeah, we will merge once .NET 9 is released

Newer targets can be used on appveyor with apt-get, as in the opposite of #1352

test/Renci.SshNet.AotCompatibilityTestApp/Renci.SshNet.AotCompatibilityTestApp.csproj
Rob-Hague: Fine either way

test/Renci.SshNet.Benchmarks/Renci.SshNet.Benchmarks.csproj
Rob-Hague: Yes

test/Renci.SshNet.IntegrationBenchmarks/Renci.SshNet.IntegrationBenchmarks.csproj
Rob-Hague: Yes

test/Renci.SshNet.IntegrationTests/Renci.SshNet.IntegrationTests.csproj
Rob-Hague: Let's keep a net8.0 target just for local testing

```suggestion
    <TargetFrameworks>net48;net8.0;net9.0</TargetFrameworks>
```

mus65:
This is now only blocked by AppVeyor not supporting .NET 9. apt-get is not going work either because the dotnet9 package is not available in Ubuntu 22.04 (see https://launchpad.net/ubuntu/+source/dotnet9 ).

@Rob-Hague has there been any thought about moving to GitHub actions?

Rob-Hague:
I seem to remember @WojciechNagorski trying it but I can't find a PR so may be imagining

Rob-Hague:
Feel free to give a go though

Rob-Hague (approved):

PR #1484, opened 2024-09-05
fix "client not connected" after SFTP reconnect

### LINKED ISSUE #1474: SFTP: "client not connected" after server closed connection
As reported by multiple people here https://github.com/sshnet/SSH.NET/issues/843#issuecomment-2242437779 , #1362 does not completely fix the general issue. A "client not connected" exception can still happen after checking for `IsConnected` .

I will try to reproduce this and create a PR.

mus65:
if the server closes the session and the client reconnects, this currently leads to a broken state because the session is re-created, but the SFTP subsession is not and still references the old session.

This causes all operations to fail with "client not connected" or even throwing the "An established connection was aborted by the server." exception of the old session.

Always re-create the SFTP subsession to fix this.

Also noticed another issue while looking into this: if the session already exists, but it not connected anymore, the old instance was never disposed.

fixes #1474

### DIFF
--- a/src/Renci.SshNet/BaseClient.cs
+++ b/src/Renci.SshNet/BaseClient.cs
@@ -240,6 +240,11 @@ namespace Renci.SshNet
             var session = Session;
             if (session is null || !session.IsConnected)
             {
+                if (session is not null)
+                {
+                    DisposeSession(session);
+                }
+
                 Session = CreateAndConnectSession();
             }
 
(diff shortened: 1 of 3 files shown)


WojciechNagorski:
image

Rob-Hague:
I think those are unrelated. I can take a closer look later

mus65:
Tests all pass locally on linux, so probably unrelated.

Rob-Hague (approved):

Rob-Hague:
Does it fix https://github.com/sshnet/SSH.NET/issues/1474#issuecomment-2325899186 as well?

mus65:
yes, this is also caused by the SFTP subsession referencing the old session object (and therefore re-throwing its old exception).

PR #1487, opened 2024-09-07
[Private Key] Support more ciphers for OpenSSH private key decryption

### LINKED ISSUE #1483: Exception when initializing PrivateKeyFile with type aes256-gcm@openssh.com
Hey there, I am trying to connect to an SFTP server using an OpenSSH key. This OpenSSH key uses the cipher aes256-gcm@openssh.com. When initializing a new instance of PrivateKeyFile with the path to the private key and private key passphrase, the library throws an exception: `Cipher 'aes256-gcm@openssh.com' is not supported for an OpenSSH key.`

I was under the impression 2024.1 added support for this specific cipher. Does the key need to be initialized differently?

I am using .NET 8 on macOS 14.6.1.

scott-xu:
With this PR, SSH.NET supports below ciphers when decrypt OpenSSH private key.

* 3des-cbc
* aes128-cbc
* aes192-cbc
* aes256-cbc
* aes128-ctr
* aes192-ctr
* aes256-ctr
* aes128-gcm<span></span>@openssh.com
* aes256-gcm<span></span>@openssh.com
* chacha20-poly1305<span></span>@openssh.com

The list is now the same with the output of command `ssh -Q cipher`.

There's a `-Z` option when generate key pair using ssh-keygen[^1].

> -Z cipher
>         Specifies the cipher to use for encryption when writing
>         an OpenSSH-format private key file.  The list of
>         available ciphers may be obtained using "ssh -Q cipher".
>         The default is “aes256-ctr”.

Fixes https://github.com/sshnet/SSH.NET/issues/1483

[^1]: https://www.man7.org/linux/man-pages/man1/ssh-keygen.1.html


src/Renci.SshNet/PrivateKeyFile.cs
```diff
@@ -518,38 +561,76 @@ private static Key ParseOpenSshV1Key(byte[] keyFileData, string passPhrase)
...
+                // apply the kdf to derive a key and iv from the passphrase
+                var passPhraseBytes = Encoding.UTF8.GetBytes(passPhrase);
+                var keyiv = new byte[keyLength + ivLength];
+                new BCrypt().Pbkdf(passPhraseBytes, salt, rounds, keyiv);
+
+                var key = keyiv.Take(keyLength);
+                var iv = keyiv.Take(keyLength, ivLength);
+
+                var cipher = cipherInfo.Cipher(key, iv);
+                var cipherData = keyReader.ReadBytes(privateKeyLength + cipher.TagSize);
```
Rob-Hague, line 616: Can you point me to where the `+ TagSize` comes from?

scott-xu: Not quite sure if I understand your question... The AEAD cipher will generate tag when encrypt data. In this case (OpenSSH), the tag is appended to the cipher data. When decrypt and verify authenticity, we need to read not only cipher data but also the tag.

Rob-Hague: I am asking where you have found that the tag is appended to the cipher data. In https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.key I just see `string	encrypted, padded list of private keys`

scott-xu: https://github.com/openssh/openssh-portable/blob/master/sshkey.c#L2951

Rob-Hague: Thanks

Rob-Hague (approved):

src/Renci.SshNet/PrivateKeyFile.cs
scott-xu: And https://github.com/openssh/openssh-portable/blob/master/cipher.c#L340

PR #1488, opened 2024-09-08
Migrate to NuGet Central Package Management

mus65:
This migrates to [Central Package Management](https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management) . Versions stay the same, so there is no functional difference.

Additional changes I made as part of this:
- replaced some usages of the  `[1.0.0,)` syntax with `1.0.0` since these are equivalent, see [docs](https://learn.microsoft.com/en-us/nuget/concepts/package-versioning?tabs=semver20sort#version-ranges)
- removed the testcontainers -> SSH.NET exclude from IntegrationTests . I can not reproduce this issue anymore. Even if it still woud be a problem, I think it would be better to use `VersionOverride` here since the version has no meaning anyway.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,6 +1,7 @@
 <Project>
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+    <CentralPackageVersionOverrideEnabled>false</CentralPackageVersionOverrideEnabled>
   </PropertyGroup>
   <ItemGroup>
     <PackageVersion Include="Appveyor.TestLogger" Version="2.0.0" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,0 +1,25 @@
+<Project>
+  <PropertyGroup>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageVersion Include="Appveyor.TestLogger" Version="2.0.0" />
+    <PackageVersion Include="BenchmarkDotNet" Version="0.13.12" />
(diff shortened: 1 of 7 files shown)


Rob-Hague:
Just want to double check the exclude, not sure why there would be no build warnings anymore

mus65:
I can only reproduce the original warning with Testcontainers <= 3.5.0 which depended on SSH.NET 2020.0.2 . With Testcontainers >= 3.6.0 (which depends on SSH.NET 2023.0.0), the warning disappears.

Still no idea why this makes a difference though.

Directory.Packages.props
```diff
@@ -0,0 +1,25 @@
+<Project>
+  <PropertyGroup>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
```
drieseng, line 3: Also set **CentralPackageVersionOverrideEnabled** to **false**. The whole idea of CPM is to manage versions centrally.

Directory.Packages.props
```diff
@@ -0,0 +1,25 @@
...
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageVersion Include="Appveyor.TestLogger" Version="2.0.0" />
```
drieseng, line 7: We should consider upgrading the following packages:
* BenchmarkDotNet
* Meziantou.Analyzer
* MSTest.*
* Moq
* SonarAnalyzer.CSharp

Lets do this in an another PR though.

drieseng (changes requested):

Directory.Packages.props
mus65: I wll do a separate PR.

drieseng (approved):

Rob-Hague (approved):
Thanks for checking the build warnings. Let's be optimistic that it won't reoccur

Issue #1090, opened 2023-02-28
Support PKCS#8


wertzui:
At the Moment only PKCS#1 RSA keys are supported.
Those begin with `-----BEGIN RSA PRIVATE KEY-----`

However PKCS#8 Keys are not supported.
Those begin with `-----BEGIN PRIVATE KEY-----`

I stumbled across this problem when using openssl to generate a key.
By default openssl will write using the PKCS#8 format.
Only if you give it the `-traditional` flag, it will write in PKCS#1 format.

The error I was getting with PKCS#8 was the following one, because the RegEx that is used to extract the algorithm did not match, because the algorithm is not present in  `-----BEGIN PRIVATE KEY-----`.
```
Renci.SshNet.Common.SshException: Invalid private key file.
   at Renci.SshNet.PrivateKeyFile.Open(Stream privateKey, String passPhrase)
   at Renci.SshNet.PrivateKeyFile..ctor(Stream privateKey)
```

PR #1490, opened 2024-09-10
Ref System.Formats.Asn1 nuget package

scott-xu:
This PR references [System.Formats.Asn1](https://www.nuget.org/packages/System.Formats.Asn1) instead of using home-made `DerData` class.

I'm trying to add PKCS#8 support after this PR. See [RFC5208](https://www.rfc-editor.org/rfc/rfc5208)
Contributes to https://github.com/sshnet/SSH.NET/issues/1090


src/Renci.SshNet/Renci.SshNet.csproj
```diff
@@ -50,6 +50,7 @@
 
   <ItemGroup>
     <PackageReference Include="BouncyCastle.Cryptography" Version="[2.4.0,)" />
+    <PackageReference Include="System.Formats.Asn1" Version="[8.0.1,)" />
```
mus65: This is part of the runtime on modern .NET, so it should only be needed for `net462`, `netstandard2.0` and `netstandard2.1`,

Rob-Hague: We can also remove the explicit System.Memory dependency since this package depends on it already (and System.Buffers - yay ArrayPool)

Rob-Hague:
Also delete ObjectIdentifier.cs?

src/Renci.SshNet/Renci.SshNet.csproj
scott-xu: Makes sense. Thanks for the suggestion. @mus65 @Rob-Hague

src/Renci.SshNet/Renci.SshNet.csproj
```diff
@@ -42,10 +42,10 @@
     <PackageReference Include="Nerdbank.GitVersioning" Version="3.7.70-alpha" PrivateAssets="all" />
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(TargetFramework)' == 'net462' or '$(TargetFramework)' == 'netstandard2.0' ">
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net462' or '$(TargetFramework)' == 'netstandard2.0' or '$(TargetFramework)' == 'netstandard2.1'">
```
Rob-Hague: We don't want/need to reference Microsoft.Bcl.AsyncInterfaces for netstandard2.1. We should have:

net462 and netstandard2.0: AsyncInterfaces, Asn1, BouncyCastle
netstandard2.1: Asn1, BouncyCastle
netN.0: BouncyCastle

src/Renci.SshNet/Renci.SshNet.csproj
Rob-Hague: > (and System.Buffers - yay ArrayPool)

I forgot that System.Memory already depends on System.Buffers. Not sure why the Asn1 package declares an explicit dependency on System.Buffers

src/Renci.SshNet/Renci.SshNet.csproj
scott-xu: Just out of curiosity, why `Microsoft.Bcl.AsyncInterfaces` targets netstandard2.1?

Rob-Hague: So that it does not result in unnecessary transitive dependencies (like System.Threading.Tasks.Extensions). It is actually just a facade which forwards to the in-box types: https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Bcl.AsyncInterfaces/src/Microsoft.Bcl.AsyncInterfaces.csproj

scott-xu: Yes, the `Microsoft.Bcl.AsyncInterfaces` for netstandard2.1 is just a facade, but `System.Formats.Asn1` for net6.0+ contains all types which is confused.
image
image

Rob-Hague (approved):
Nice

PR #1491, opened 2024-09-10
Miscellaneous cleanup/throw helpers

Rob-Hague:
These are some random changes I've accumulated in some branches, which I've extracted out to reduce diffs & conflicts. They are roughly:

Add a reference to PolySharp to allow nullable attributes and CallerArgumentExpressionAttribute etc.
Add a ThrowHelpers class
Add #nullable to PrivateKeyFile and cleanup PrivateKeyFileTest
Also change around some of the newly added keys - I have a use for this with certificate support.


Rob-Hague:
Ok to get a sign-off here?

Directory.Build.props
```diff
@@ -32,7 +32,7 @@
         Disable nullable warnings on old frameworks because of missing annotations.
   -->
   <PropertyGroup Condition=" !$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0')) ">
-    <NoWarn>$(NoWarn);CS8602</NoWarn>
+    <NoWarn>$(NoWarn);CS8602;CS8604;CS8777</NoWarn>
```
drieseng, line 35: Not possible to fix these?

src/Renci.SshNet/.editorconfig
```diff
@@ -188,3 +188,6 @@ dotnet_diagnostic.MA0040.severity = none
 # MA0042: Do not use blocking calls in an async method
 # duplicate of CA1849
 dotnet_diagnostic.MA0042.severity = none
+
+# S3236: Caller information arguments should not be provided explicitly
+dotnet_diagnostic.S3236.severity = none
```
drieseng, line 193: Why was this one reported? Don't have time to check myself now.

drieseng (approved):

Directory.Build.props
Rob-Hague: They are coming from string.IsNullOrEmpty and friends which are not annotated for nullable analysis on lower targets

src/Renci.SshNet/.editorconfig
Rob-Hague: It comes when one wants to provide a value for a parameter marked with `[CallerExpressionAttribute]`, e.g.

```c#
public TimeSpan Timeout
{
    get
    {
        return _timeout;
    }
    set
    {
        // EnsureValidTimeout(this TimeSpan timeSpan, [CallerArgumentExpression(nameof(timeSpan))] string? paramName = null)

        value.EnsureValidTimeout(nameof(Timeout));

        _timeout = value;
    }
}
```

I would probably suppress this one at the call sites, but #1480 will also bring more places where we don't want it (from Debug.Assert)

Rob-Hague:
Thanks

scott-xu:
> Also change around some of the newly added keys - I have a use for this with certificate support.

How about consolidating the keys. e.g. one key-pair per algorithm (and key size) , but with different format and different encryption for private key? It will be easier for integration test.

Rob-Hague:
It seems OK how it is - it was a bit of a pain to set everything up for #1498 so I would rather like to avoid changing it

PR #1492, opened 2024-09-12
Update dependencies

mus65:
This updates everything except SonarAnalyzer.CSharp (will do a separate PR for it since it's a lot more work).

Documentation changes are due to https://github.com/meziantou/Meziantou.Analyzer/blob/main/docs/Rules/MA0154.md


drieseng (approved):

drieseng:
@mus65, don't blindly change code to silence **SonarAnalyzer.CSharp**. It's known to have some false positives.

I may be asking too much here, but would it be possible to do a commit per rule that you're providing fixes for?

PR #1494, opened 2024-09-14
Update SonarAnalyzer.CSharp

mus65:
There a quite a few new analyzers. For easier review, I created a separate commit for each analyzer, including a link to the sonar source docs and additional notes from me.

Some of these may be worth disabling instead of adding suppressions. I opted for adding suppressions for now.


Rob-Hague:
IMHO S3993 (attribute usage) and S3431 (ExpectedException) have value, the others do not. But you can wait for other opinions - I don't like the "every analyzer" approach that we take

test/Renci.SshNet.IntegrationTests/AuthenticationTests.cs
```diff
@@ -90,18 +90,15 @@ public void Multifactor_PublicKeyWithPassPhrase()
         }
 
         [TestMethod]
-        [ExpectedException(typeof(SshPassPhraseNullOrEmptyException))]
         public void Multifactor_PublicKeyWithEmptyPassPhrase()
```
Rob-Hague: Could you delete this test or move it to PrivateKeyAuthenticationMethodTest.cs in the unit tests project? Not sure why it's an integration test

mus65: `Test_PrivateKey_SSH2_Encrypted_ShouldThrowSshPassPhraseNullOrEmptyExceptionWhenPassphraseIsNull` in `PrivateKeyFileTest.cs` already covers this, so I removed it.

drieseng:
@mus65 We should globally disable **S2325** (and revert the changes that you applied to silence this rule).

This is what I have in my **.editorconfig** at work:

```txt
# S2325: Methods and properties that don't access instance data should be static
# https://rules.sonarsource.com/csharp/RSPEC-2325
#
# This is a duplicate of CA1822, and reports false positives for ASP.NET Core Middleware
dotnet_diagnostic.S2325.severity = none
```

**CA1822** also allows you to configure the API surface that the rule applies to. We do this for example in the **.editorconfig** of the **Renci.SshNet** project. Also **S2325** does not appear to respect the `generated_code = true` .editorconfig setting which we set for **Bcrypt.cs** (as that code was imported by us).

mus65:
@drieseng done

Rob-Hague:
Sorry for the conflicts, can you also revert/suppress S127 then I'll merge

Rob-Hague (approved):
Thanks

PR #1496, opened 2024-09-16
Add support for OpenSSL PKCS#8 private key format

### LINKED ISSUE #1090: Support PKCS#8

scott-xu:
This PR adds support for PKCS#8 private key format (in practice, OpenSSL is used to generate the keys)

There are several options to add support for PKCS#8 private key format:
 1. Use [System.Security.Cryptography.Pkcs](https://github.com/dotnet/runtime/tree/main/src/libraries/System.Security.Cryptography.Pkcs). It is from Microsoft and it supports exactly the same targets frameworks as SSH.NET. However, it is shipped as NuGet only which means we need to add it as a dependency for all target frameworks.
 2. Call [`ImportPkcs8PrivateKey`](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.asymmetricalgorithm.importpkcs8privatekey?view=net-8.0) or [`ImportFromEncryptedPem`](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.asymmetricalgorithm.importfromencryptedpem?view=net-8.0) from relevant concrete class. Both methods applied to netstandard2.1+ and Core 3.0+. However, I couldn't find a way to detect the key type before call the methods, for example a static method to load PKCS#8 private key and returns BCL's `AsymmetricAlgorithm`.
 3. Use BouncyCastle's [PemReader](https://github.com/bcgit/bc-csharp/blob/master/crypto/src/openssl/PEMReader.cs) to load the private key and return BouncyCastle's `AsymmetricKeyParameter`. Or use BouncyCastle's underlying [PrivateKeyFactory](https://github.com/bcgit/bc-csharp/blob/master/crypto/src/security/PrivateKeyFactory.cs) to `CreateKey` / `DecryptKey` which also returns BouncyCastle's `AsymmetricKeyParameter`. Then we need to bridge BouncyCastle's `AsymmetricKeyParameter` with SSH.NET's `Key`. This option may add some complexity to map types and unnecessary memory allocations.
 4. My finally proposal is to use BouncyCastle to decrypt the PCKS#8 private key, read the PrivateKeyInfo Der structure and then read the parameters and key manually with `AsnReader` and finally construct the SSH.NET relevant concrete `Key` class.
 With this option, there's no extra NuGet dependency and performance is balanced.

I also considered using `Impl` pattern for `PrivateKeyFile` class, but haven't had time to try. We can continue evolving this PR or create a new PR if needs be.

We can also consider dropping support for OpenSSL traditional private key format and drop ssh.com private key format as they are old and weak.

Fix https://github.com/sshnet/SSH.NET/issues/1090


src/Renci.SshNet/Security/Cryptography/DsaKey.cs
```diff
@@ -108,31 +118,6 @@ public DsaKey(SshKeyData publicKeyData)
             DSA = LoadDSA();
         }
 
-        /// <summary>
-        /// Initializes a new instance of the <see cref="DsaKey"/> class.
-        /// </summary>
-        /// <param name="privateKeyData">DER encoded private key data.</param>
-        public DsaKey(byte[] privateKeyData)
```
Rob-Hague: Is it necessary to remove this?

Same applies to the other key constructors

src/Renci.SshNet/Security/Cryptography/EcdsaKey.cs
```diff
@@ -207,34 +206,12 @@ public EcdsaKey(SshKeyData publicKeyData)
         /// <summary>
         /// Initializes a new instance of the <see cref="EcdsaKey"/> class.
         /// </summary>
-        /// <param name="curve">The curve name.</param>
+        /// <param name="curve">The curve oid.</param>
```
Rob-Hague: That's a nasty break. Let's try to avoid it

test/Data/Key.PKCS8.DSA.pub
Rob-Hague, line 1: Is it possible to make OpenSSH public file formats instead? They are more useful e.g. for authorized_keys in the integration tests

Rob-Hague (commented):
The chosen approach seems reasonable. I don't think see why any of DsaKey, EcdsaKey, RsaKey types need to change (in particular, breaking changes)

src/Renci.SshNet/Security/Cryptography/DsaKey.cs
scott-xu: I remove it because there could be other DER data structures. The current Key constructors with byte array parameter assume the data is DER encoded _OpenSSL traditional format_ which is not always true. It could be PKCS#8 format as well.

In practice, most package consumers are not necessarily call Key constructors directly. It might not bring huge impacts. 

Anyway, to gracefully deprecate the constructors with byte array parameter, we can apply `Obsolete` attribute to the constructors and remove them in the future release. What do you think? @Rob-Hague

src/Renci.SshNet/Security/Cryptography/EcdsaKey.cs
scott-xu: By reading the DER data, we get the curve oid string directly. It looks strange if we convert curve oid string to curve name  in PrivateKeyFile and then convert name back to oid string in Ecdsakey.

src/Renci.SshNet/Security/Cryptography/DsaKey.cs
Rob-Hague: I am sorry, I do not think there is a reason to break the existing contract.

I agree there is probably not many people using these constructors, and if we were starting from scratch we might do it differently. And although we make breaking changes practically every release, I think we can justify each of them. I do not believe this one can be justified.

As library developers, when we make a breaking change we are telling the user: "we expect you to bear the pain for our work". As a user, I can sometimes accept that when the user experience is improved and when it is easy to react to the change. Otherwise I find it annoying and frankly inconsiderate from the developers.

Whether we make this change or not, we experience no pain, so we should not cause any pain to users. We can instead improve things here by clarifying in the documentation that this constructor expects an OpenSSL format.

src/Renci.SshNet/Security/Cryptography/EcdsaKey.cs
Rob-Hague: Yeah, it's strange and unfortunate, but it's the publicly exposed contract, so my other comment applies here too. If you like, you can change it to accept either the name *or* oid

src/Renci.SshNet/PrivateKeyFile.cs
```diff
@@ -717,6 +737,265 @@ private static Key ParseOpenSshV1Key(byte[] keyFileData, string passPhrase)
...
+        /// }
+        /// </code>
+        /// When generating a transfer encoding, generators SHOULD use
+        /// Distinguished Encoding Rules (DER) and receivers SHOULD be
+        /// prepared to handle Basic Encoding Rules (BER) and DER.
+        /// </remarks>
+        /// <returns>The <see cref="EcdsaKey" />.</returns>
+        private static EcdsaKey ParseECPrivateKey_SEC1(byte[] keyData)
+        {
+            var keyReader = new AsnReader(keyData, AsnEncodingRules.DER);
```
Rob-Hague: Based on the comment above, should this be

```suggestion
            var keyReader = new AsnReader(keyData, AsnEncodingRules.BER);
```

?

src/Renci.SshNet/Security/Cryptography/DsaKey.cs
scott-xu: Sorry I didn't explain well.
The current Key constructors with byte array parameter assume the data is DER encoded OpenSSL **traditional** PKCS#1 format.
OpenSSL use PKCS#8 by default. It only uses PKCS#1 when option "-traditional" is specified.
The newer "[genpkey](https://docs.openssl.org/3.0/man1/openssl-genpkey/)" command even doesn't show "-traditional" option in the guide.

src/Renci.SshNet/PrivateKeyFile.cs
scott-xu: DERs are subset of BERs. I think DERs are enough here since the data is generated using DERs?

src/Renci.SshNet/Security/Cryptography/DsaKey.cs
Rob-Hague: I understand the situation, I just don't think it is worth deleting the constructor or changing its behaviour, unless you can somehow detect which format it is using while reading. If not, then the constructor should stay the same

scott-xu: Then how about keep the constructor but apply the `Obsolete` attribute? Looking at the industry, the PKCS#1 is being replaced by PKCS#8.

src/Renci.SshNet/Security/Cryptography/EcdsaKey.cs
scott-xu: We need to either document the breaking change in the release notes, or keep the current behavior but increase the code complexity (accept **either name or oid**). 😢

src/Renci.SshNet/Security/Cryptography/DsaKey.cs
scott-xu: After some more thinking, I'm leaning towards to just delete those constructors.

We removed old target frameworks. People who are using old frameworks should target to new frameworks. See https://github.com/sshnet/SSH.NET/pull/1109
We removed legacy algorithms. People who are using legacy algorithms should use new algorithms. See https://github.com/sshnet/SSH.NET/pull/1442

This change only affects people who are directly calling the constructor with PKCS#1 data. They should use PKCS#8 data instead.

**Note**: People who are using `PrivateKeyFile` are not affected. I believe most people are using `PrivateKeyFile` instead of each `Key` class.

src/Renci.SshNet/PrivateKeyFile.cs
Rob-Hague: I am reading this part of the comment: "receivers SHOULD be prepared to handle Basic Encoding Rules (BER) and DER"

In my understanding, that means using AsnEncodingRules.BER to match the comment

src/Renci.SshNet/Security/Cryptography/EcdsaKey.cs
Rob-Hague: It should be the latter. Does not seem that complex to me

```diff
diff --git a/src/Renci.SshNet/Security/Cryptography/EcdsaKey.cs b/src/Renci.SshNet/Security/Cryptography/EcdsaKey.cs
index 70e2e1a9..6f275b20 100644
--- a/src/Renci.SshNet/Security/Cryptography/EcdsaKey.cs
+++ b/src/Renci.SshNet/Security/Cryptography/EcdsaKey.cs
@@ -271,17 +271,20 @@ private static Impl Import(string curve_oid, byte[] publickey, byte[]? privateke

         private static string GetCurveOid(string curve_s)
         {
-            if (string.Equals(curve_s, "nistp256", StringComparison.OrdinalIgnoreCase))
+            if (string.Equals(curve_s, "nistp256", StringComparison.OrdinalIgnoreCase)
+                || curve_s == ECDSA_P256_OID_VALUE)
             {
                 return ECDSA_P256_OID_VALUE;
             }

-            if (string.Equals(curve_s, "nistp384", StringComparison.OrdinalIgnoreCase))
+            if (string.Equals(curve_s, "nistp384", StringComparison.OrdinalIgnoreCase)
+                || curve_s == ECDSA_P384_OID_VALUE)
             {
                 return ECDSA_P384_OID_VALUE;
             }

-            if (string.Equals(curve_s, "nistp521", StringComparison.OrdinalIgnoreCase))
+            if (string.Equals(curve_s, "nistp521", StringComparison.OrdinalIgnoreCase)
+                || curve_s == ECDSA_P521_OID_VALUE)
             {
                 return ECDSA_P521_OID_VALUE;
             }
```

src/Renci.SshNet/Security/Cryptography/DsaKey.cs
Rob-Hague: This improvement does not require changing public API. It brings us zero benefits, and brings detriment to users

Rob-Hague (changes requested):
Please restore the public constructors

Sorry about the conflicts

src/Renci.SshNet/Security/Cryptography/DsaKey.cs
scott-xu: If everyone agrees, then I can add them back. Would you mind sharing your thoughts here? @drieseng @WojciechNagorski @mus65

src/Renci.SshNet/Security/Cryptography/EcdsaKey.cs
scott-xu: Sure from coding perspective, it is not complex. It is just the method signature **definition**. Now the first parameter could be either name or oid. Would it break single responsibility principle?

src/Renci.SshNet/PrivateKeyFile.cs
scott-xu: Actually I'm confused about the doc. Do you know why it says "BER **and** DER".

src/Renci.SshNet/Security/Cryptography/DsaKey.cs
mus65: I agree removing it is too much. But imho it should at least be obsoleted since it is ambigious and confusing (.NET 9 recently [made a similiar API obsolete](https://github.com/dotnet/core/blob/main/release-notes/9.0/preview/preview7/libraries.md#changes-to-x509-certificate-loading)). But then there should be an alternative public API people can move to (I assume there is? I haven't checked the whole PR).

src/Renci.SshNet/PrivateKeyFile.cs
Rob-Hague: I guess they are trying to be explicit but they do not make it any clearer by doing so

src/Renci.SshNet/Security/Cryptography/DsaKey.cs
scott-xu: > I agree removing it is too much. But imho it should at least be obsoleted since it is ambigious and confusing (.NET 9 recently [made a similiar API obsolete](https://github.com/dotnet/core/blob/main/release-notes/9.0/preview/preview7/libraries.md#changes-to-x509-certificate-loading)). But then there should be an alternative public API people can move to (I assume there is? I haven't checked the whole PR).

Thanks for sharing the info. I know I'm not alone.

BTW, Microsoft calls their new design **"one method, one purpose"** which is exactly what I want to apply for method `static string GetCurveOid(string curve_s)`.

Rob-Hague (approved):
Looks good, thanks for the changes

scott-xu:
TBH, I don't like my changes. Life is short. I may leave for a while.
Best wishes for SSH.NET

Rob-Hague:
Understood @scott-xu. I know it has not always been smooth sailing, but you make valuable contributions nonetheless. Thanks

PR #812, opened 2021-04-20
Add interface to SftpFile #120

wxtsxt:
Changes:
- Create `ISftpFile` interface. 
- Return ISftpFile from SftpClient instead of SftpFile. 
- Make ISftpClient interface disposable. Right now it is impossible to call Dispose() on the ISftpClient object

There is a breaking change in SftpClient as it returns interface ISftpFile instead of SftpFile.
Such change allows to create unit tests with mocks (sample using NSubstitude):
```csharp
            var file = Substitute.For<ISftpFile>();
            file.FullName.Returns("my file name");

            var sftpClient = Substitute.For<ISftpClient>();
            sftpClient.ListDirectory(myPath).Returns(new List<ISftpFile> { file });
```

PR #1508, opened 2024-09-26
Add interface to SshCommand

### LINKED ISSUE #1507: Add interface for SshCommand
Similar to #1486, it would be also be great to add an interface to SshCommand for testing, as this would allow mocking up the return value of SshClient.RunCommand().

eugencowie:
Changes:

* Add ISshCommand interface containing all public properties and methods of SshCommand
* Update SshClient methods to return ISshCommand instead of SshCommand

Closes #1507

### DIFF
--- a/src/Renci.SshNet/ISshClient.cs
+++ b/src/Renci.SshNet/ISshClient.cs
@@ -40,7 +40,7 @@ namespace Renci.SshNet
         /// <param name="commandText">The command text.</param>
         /// <returns><see cref="SshCommand"/> object.</returns>
         /// <exception cref="SshConnectionException">Client is not connected.</exception>
-        public SshCommand CreateCommand(string commandText);
+        public ISshCommand CreateCommand(string commandText);
 
         /// <summary>
         /// Creates the command to be executed with specified encoding.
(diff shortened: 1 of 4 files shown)


Rob-Hague:
I am a bit uneasy about this one because of the breaking changes it introduces

k0ka:
what exact breaking change are you referring to? As I understand the only problem might occur if the user had a code like this
```csharp
SshCommand command = client.CreateCommand("ls");
```

But this code would be quite rare as the modern preferred coding style is 
```csharp
var command = client.CreateCommand("ls");
```

Every release there are small breaking changes. This change won't be very big and would be caught during compile time. 

We might introduce another interface `ISshClient2` which returns `ISshCommand` instead of `SshCommand`. But it won't make much sense as we anyway have to replace the main interface `ISsshClient` with it.

Rob-Hague:
Yes that is the breaking change I am referring to.

> But this code would be quite rare as the modern preferred coding style is

It is not my preferred coding style :-)

The breaking changes in each release are considered based on how many people it is going to benefit vs inconvenience. You can see my thoughts on mocking the library here https://github.com/sshnet/SSH.NET/issues/1667#issuecomment-3065200031, but in short, I don't consider it an important feature for which is worth inconveniencing an unknown but possibly large number of people.

If we want to compare the proposed change here against prior changes:

1. #812. IMO this was a mistake. Even so, it is less inconvenient than this change due to implicit conversion in foreach loops (so `foreach (SftpFile file in client.ListDirectory(...))` still compiles)
2. #1628. Also a breaking change, but since these interfaces seem to only exist for people to mock, then it is more like "you get what you asked for"

PR #1509, opened 2024-09-28
Replace DiagnosticAbstration with Microsoft.Extensions.Logging.Abstractions

mus65:
This replaces the `DiagnosticAbstraction` with `Microsoft.Extensions.Logging.Abstractions` and allows consuming applications to pass their own `ILoggerFactory` to SSH.NET.

I don't expect this to be merged as is since there are a few things worth discussing:

#### Why do this?

Because it is the de-facto standard for modern and structured logging in .NET and allows consuming applications to easily integrate SSH.NET into their own logging. This also means users can more easily provide log messages in issues.

Also the currently used Trace API is pretty much [considered obsolete](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/logging-tracing#trace).

#### Injecting ILoggerFactory vs static instance

The imho best way to implement this would be to add an `ILoggerFactory` to the public constructors. The problem is that passing this through all the constructors and down to all the layers is quite a lot of work. I actually tried this and finally gave up when I realized that the logging in `KeyExchange.cs` meant that the `ConnectionInfo` constructors would also need an `ILoggerFactory` (because of the `KeyExchangeAlgorithms` dictionary).

So I opted for a static configuration class, similiar how [Npgsql](https://github.com/npgsql/npgsql/blob/main/src/Npgsql/NpgsqlLoggingConfiguration.cs) handles this. Imho this is "good enough" since having multiple `ILoggerFactory` instances in an application sounds like quite a niche use case.

Applications simply have to call `SshNetLoggingConfiguration.InitializeLogging(loggerFactory);` before using SSH.NET.

#### Compile-time logging source generation

While this implementation already allocates less than the old one (because it doesn't create the full string unless logging is enabled), things could be further improved with the [LoggerMessage attribute](https://learn.microsoft.com/en-us/dotnet/core/extensions/logger-message-generator). This would mean even better performance, but it also means writing a method for every single log message, which can be quite tedious. I would be fine with doing this, but decided to not bother unless it is decided to use it.

There are also quite a few unneccessary allocation because of the usage of `Session.ToHex()`. I didn't change this here since it could be optimized separately (even without this PR being merged).

#### New dependency

This obviously adds a new dependency to `Microsoft.Extensions.Logging.Abstractions` to SSH.NET (the `.Logging`and `Console`packages are only for the tests) . Imho this is not really an issue. This library is as minimal as it gets and most applications already pull this is in anyway, either directly or by another dependency needing it (ASP.NET Core requires it for example).


Rob-Hague:
I think it's the right way to go. The static init seems reasonable given the prior art in other libraries. I don't think the source generator is necessary (definitely tedious). Wrt the new dependency, we can always merge after the next release (hopefully soon) so there is not a big wave of dependencies in one release

mus65:
I just realized that this PR actually introduces new  `ToHex()` allocations because in release mode this was never called before  because of the `[Conditional("DEBUG")]` attribute on `DiagnosticAbstraction.Log`.

So I optimized this to only call `ToHex()` once instead of every log call. This could be even more improved by using a `Lazy` and checking `if (logger.IsEnabled())` before every message, but I'm not sure this is worth it.

src/Renci.SshNet/SshNetLoggingConfiguration.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        /// <summary>
+        /// Initializes the logging for SSH.NET.
+        /// </summary>
+        /// <param name="loggerFactory">The logger factory.</param>
+        public static void InitializeLogging(ILoggerFactory loggerFactory)
+        {
+            ThrowHelper.ThrowIfNull(loggerFactory);
+            LoggerFactory = loggerFactory;
+        }
```
Rob-Hague, line 24: Maybe allow null here as a way to turn it off?

```suggestion
        /// <param name="loggerFactory">
        /// The logger factory, or <see langword="null"/> to disable logging.
        /// </param>
        public static void InitializeLogging(ILoggerFactory? loggerFactory)
        {
            LoggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
        }
```

src/Renci.SshNet/Channels/ChannelDirectTcpip.cs
```diff
@@ -31,6 +33,7 @@ internal sealed class ChannelDirectTcpip : ClientChannel, IChannelDirectTcpip
         public ChannelDirectTcpip(ISession session, uint localChannelNumber, uint localWindowSize, uint localPacketSize)
             : base(session, localChannelNumber, localWindowSize, localPacketSize)
         {
+            _logger = SshNetLoggingConfiguration.LoggerFactory.CreateLogger<ChannelDirectTcpip>();
```
Rob-Hague, line 36: There is a curious difference between the generic and non-generic CreateLogger, see https://github.com/dotnet/aspnetcore/pull/50184 and related links. Basically, the non-generic seems to be the way to go

docfx/logging.md
```diff
@@ -0,0 +1,10 @@
+Logging
+=================
+
+SSH.NET uses the `Microsoft.Extensions.Logging` API to log diagnostic messages. In order to access the log messages of SSH.NET in your own application for diagnosis, simply register your own `ILoggerFactory` before using the SSH.NET APIs with the following code:
+
+```cs
+SshNetLoggingConfiguration.InitializeLogging(loggerFactory);
+```
```
Rob-Hague:
```suggestion
SSH.NET uses the `Microsoft.Extensions.Logging` API to log diagnostic messages. In order to access the log messages of SSH.NET in your own application for diagnosis, register your own `ILoggerFactory` before using the SSH.NET APIs, for example:

```cs
ILoggerFactory loggerFactory = LoggerFactory.Create(builder =>
{
    builder.SetMinimumLevel(LogLevel.Debug);
    builder.AddConsole();
});

Renci.SshNet.SshNetLoggingConfiguration.InitializeLogging(loggerFactory);
```

src/Renci.SshNet/SshNetLoggingConfiguration.cs
mus65: Imho that's what `NullLoggerFactory.Instance` is already there for. Also the usual way to turn off logs for specific components would be to configure the log level for the category (`Renci.SshNet`) to `LogLevel.None`.

src/Renci.SshNet/Channels/ChannelDirectTcpip.cs
mus65: hm, feels like something .NET should fix instead but I would be fine with changing it. It would require disabling [CA2263](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2263) though. Do you want me change it?

Rob-Hague:
It looks good but I think the 'info' level is too verbose. I don't think we need an info log for each packet we send or receive, it creates a lot of logs for a simple connect+download test

Would you mind reviewing the logs/levels and paring them back where it seems sensible? I know everything was already 'info', but it was only in debug builds and not shipped in the package 

For reference, packet send/receive logs are only visible under `ssh -vvv`

mus65:
You're right, I hadn't really thought about the levels. I made the following changes:

- downgraded the "Disposing" messages to debug.
- downgraded the key exchange messages to debug.
- downgraded the "Sending message" and "Received message" to debug. I also added `if (_logger.IsEnabled(LogLevel.Debug))` here because this is a hot path, so the array allocations by the params may actually be noticable. Imho this check is not worth it for the other "once per session" messages.
- downgraded the warnings to info. Imho failures on socket shutdown etc. are not worth a warning.
- changed the default min log level for the tests to Info.

Obviously all up for debate, feel free to suggest or make changes.

src/Renci.SshNet/Session.cs
```diff
@@ -1047,9 +1047,9 @@ internal void SendMessage(Message message)
                 WaitOnHandle(_keyExchangeCompletedWaitHandle.WaitHandle);
             }
 
-            if (_logger.IsEnabled(LogLevel.Debug))
+            if (_logger.IsEnabled(LogLevel.Trace))
             {
-                _logger.LogDebug("[{SessionId}] Sending message '{MessageType}' to server: '{Message}'.", SessionIdHex, message.GetType().Name, message);
+                _logger.LogTrace("[{SessionId}] Sending message {MessageName}({MessageNumber}) to server: '{Message}'.", SessionIdHex, message.MessageName, message.MessageNumber, message.ToString());
```
Rob-Hague, line 1052: I changed it to output e.g. "Sending message SSH_MSG_NEWKEYS(21)" as it is helpful to see the number as well. It is a bit duplicative atm because most messages include the message name in the `ToString`

also I added explicit `ToString` on the message in case e.g. a json logger would dump all the properties of the message. trying to strike a balance between useful information and private session information

Rob-Hague (approved):
Thanks, all seems sensible, I just tweaked a bit

src/Renci.SshNet/Session.cs
Rob-Hague: also, wondering if logger.BeginScope would have a valid use for tagging the SessionId to logs

mus65: Using a scope for SessionId would make sense. A logging scope is based on an AsyncLocal though, so the BeginScope would have to be called in quite a few places.

PR #1510, opened 2024-09-29
Fix sftp async methods not observing error conditions

### LINKED ISSUE #1497: Infinite timeout on async UploadFile with OpenAsync/CopyToAsync
I'm using the code here https://github.com/sshnet/SSH.NET/pull/819#issuecomment-1003769248 to have an async upload file and it works.

Unfortunately, if I'm uploading a file and close the connection on the remote end, the method will never finish. I would be expecting some exception to be thrown. Instead the thread remains stuck indefinitely, and it's not a good thing for server-side process doing this operation in batches.

I found [this decade old question](https://stackoverflow.com/questions/25736589/sftp-upload-using-sshnet-library-handling-connection-issues) and tried to set the `OperationTimeout` property to 30sec, but it doesn't seems to have any effect.

I'm aware I can have a CancellationToken, but this would be a way for me to trigger a cancellation from my end, not one that happens when the connection get closed remotely or if a timeout happen.

Rob-Hague:
The async methods on SftpClient (as implemented in SftpSession) were only completing on receiving a response in the sftp layer, or via cancellation. They were not paying any attention to the ErrorOccured, SessionDisconnected or ChannelClosed events, which would result in an infinite wait when one of these occured.

This fix waits asynchronously on the wait handles using ThreadPool.RegisterWaitForSingleObject. It also adds missing plumbing of OperationTimeout.

closes #1497

### DIFF
--- a/src/Renci.SshNet/Sftp/SftpSession.cs
+++ b/src/Renci.SshNet/Sftp/SftpSession.cs
@@ -523,28 +523,24 @@ namespace Renci.SshNet.Sftp
         /// A task that represents the asynchronous <c>SSH_FXP_OPEN</c> request. The value of its
         /// <see cref="Task{Task}.Result"/> contains the file handle of the specified path.
         /// </returns>
-        public async Task<byte[]> RequestOpenAsync(string path, Flags flags, CancellationToken cancellationToken)
+        public Task<byte[]> RequestOpenAsync(string path, Flags flags, CancellationToken cancellationToken)
         {
-            cancellationToken.ThrowIfCancellationRequested();
+            if (cancellationToken.IsCancellationRequested)
+            {
+                return Task.FromCanceled<byte[]>(cancellationToken);
+            }
 
             var tcs = new TaskCompletionSource<byte[]>(TaskCreationOptions.RunContinuationsAsynchronously);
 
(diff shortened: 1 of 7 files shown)


src/Renci.SshNet/Sftp/SftpSession.cs
Rob-Hague, line 1: Easiest to view with whitespace off: https://github.com/sshnet/SSH.NET/pull/1510/files?w=1

src/Renci.SshNet/SftpClient.cs
```diff
@@ -46,21 +46,21 @@ public class SftpClient : BaseClient, ISftpClient
         /// The timeout to wait until an operation completes. The default value is negative
         /// one (-1) milliseconds, which indicates an infinite timeout period.
         /// </value>
-        /// <exception cref="ObjectDisposedException">The method was called after the client was disposed.</exception>
```
drieseng, line 49: Please also update /// doc on **ISftpClient**.

drieseng (approved):

Rob-Hague:
Thanks

dsuesstr:
Is there a release date or workaround for this?

PR #1514, opened 2024-10-07
Improve shell stream

294797392:
Improve ShellStream

* Added SendWindowChangeRequest to ShellStream

* Added a new constructor to ShellStream

* SshClient adds an overload function for CreateShellStream

When creating ShellStream, the channel will be immediately opened in the constructor of ShellStream. At this time, the calling module may not have subscribed to ShellStream's DataReceived event, which may result in the loss of a small piece of data immediately after connection.
You can use ShellStream's new constructor to solve this problem

### DIFF
--- a/src/Renci.SshNet/ShellStream.cs
+++ b/src/Renci.SshNet/ShellStream.cs
@@ -94,6 +94,29 @@ namespace Renci.SshNet
         /// <param name="height">The terminal height in pixels.</param>
         /// <param name="terminalModeValues">The terminal mode values.</param>
         /// <param name="bufferSize">The size of the buffer.</param>
+        /// <param name="dataReceived">The DataReceived Handler.</param>
+        /// <param name="errorOccurred">The ErrorOccurred Handler.</param>
+        /// <exception cref="SshException">The channel could not be opened.</exception>
+        /// <exception cref="SshException">The pseudo-terminal request was not accepted by the server.</exception>
+        /// <exception cref="SshException">The request to start a shell was not accepted by the server.</exception>
+        internal ShellStream(ISession session, string terminalName, uint columns, uint rows, uint width, uint height, IDictionary<TerminalModes, uint> terminalModeValues, int bufferSize, EventHandler<ShellDataEventArgs> dataReceived, EventHandler<ExceptionEventArgs> errorOccurred)
+               : this(session, terminalName, columns, rows, width, height, terminalModeValues, bufferSize)
+        {
+            DataReceived = dataReceived;
+            ErrorOccurred = errorOccurred;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ShellStream"/> class.
+        /// </summary>
+        /// <param name="session">The SSH session.</param>
+        /// <param name="terminalName">The <c>TERM</c> environment variable.</param>
+        /// <param name="columns">The terminal width in columns.</param>
+        /// <param name="rows">The terminal width in rows.</param>
+        /// <param name="width">The terminal width in pixels.</param>
+        /// <param name="height">The terminal height in pixels.</param>
+        /// <param name="terminalModeValues">The terminal mode values.</param>
+        /// <param name="bufferSize">The size of the buffer.</param>
         /// <exception cref="SshException">The channel could not be opened.</exception>
         /// <exception cref="SshException">The pseudo-terminal request was not accepted by the server.</exception>
         /// <exception cref="SshException">The request to start a shell was not accepted by the server.</exception>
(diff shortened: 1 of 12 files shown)


294797392:
Why does AppVeyor fail?

294797392:
Why does appveyor fail?

src/Renci.SshNet/ShellStream.cs
```diff
@@ -83,6 +83,29 @@ private void AssertValid()
...
+        /// <param name="dataReceived">The DataReceived Handler.Please note that this event needs to be manually unsubscribed.</param>
+        /// <param name="errorOccurred">The ErrorOccurred Handler.Please note that this event needs to be manually unsubscribed.</param>
+        /// <exception cref="SshException">The channel could not be opened.</exception>
+        /// <exception cref="SshException">The pseudo-terminal request was not accepted by the server.</exception>
+        /// <exception cref="SshException">The request to start a shell was not accepted by the server.</exception>
+        internal ShellStream(ISession session, string terminalName, uint columns, uint rows, uint width, uint height, IDictionary<TerminalModes, uint> terminalModeValues, int bufferSize, EventHandler<ShellDataEventArgs> dataReceived, EventHandler<ExceptionEventArgs> errorOccurred)
+               : this(session, terminalName, columns, rows, width, height, terminalModeValues, bufferSize)
+        {
+            DataReceived += dataReceived;
+            ErrorOccurred += errorOccurred;
```
Rob-Hague: Since the other constructor will run before these two lines, how does this solve your problem?

Rob-Hague:
Hi @294797392 

Are you able solve your problem using the `Read` method instead of the `DataReceived` event? e.g.

```c#
byte[] buffer = new byte[1024];
int bytesRead;
while ((bytesRead = shellStream.Read(buffer) != 0)
{
    // Process "bytesRead" bytes from buffer
}

```

294797392:
Yes, using the Read function can solve my problem.
But I think it's best to optimize it, as other developers may also encounter this issue when using it.
Do you think it is necessary to add the SendWindowChangeRequest function?
@Rob-Hague

src/Renci.SshNet/ShellStream.cs
294797392: Sorry, there was a slight issue with the modifications I made

Rob-Hague:
Yes it seems reasonable to add SendWindowChangeRequest - there have been previous requests for it. Personally I would suggest to:

1) make it return `void`: the server does not respond to a window-change request, and our implementation in `ChannelSession` just returns true so it is not much use to return `bool`
2) add `ThrowHelper.ThrowObjectDisposedIf(_disposed, this);` at the start of the method
3) add at least a basic test. For example, in `test\Renci.SshNet.Tests\Classes\ShellStreamTest.cs` there could be a test which verifies the call to `IChannelSession` mock, and verifies the `ObjectDisposedException` when the ShellStream is closed

Let me know what you think

294797392:
I agree with your idea. 
I tried to modify the code and added the SendWindowChangeRequest method, but found that if it returns void, IDE0058 compilation error will occur.
If void is returned, there may be many modifications and it will also affect the consistency of the code (there are many methods in ChannelSession that directly return true). Perhaps having a return value would be better than having no return value, as a return value can prompt the caller about the execution result of the method you called, and in the future we can also add more error handling to the method (if necessary)!
And I believe that when developing the ChannelSession module initially, there must have been a reason for returning a boolean.
@Rob-Hague

Rob-Hague:
Ok, bool seems fine to me - I don't feel strongly about it

294797392:
> Ok, bool seems fine to me - I don't feel strongly about it

Do you agree with me changing it to a boolean return value? If you agree, I will submit another PR

Rob-Hague:
Yes

test/Renci.SshNet.Tests/Classes/ShellStreamTest.cs
```diff
@@ -123,6 +123,18 @@ public void Write_AfterDispose_ThrowsObjectDisposedException()
             Assert.ThrowsException<ObjectDisposedException>(() => shellStream.Write(bytes, 0, bytes.Length));
         }
 
+        [TestMethod]
+        public void SendWindowChangeRequest_ThrowsObjectDisposedException()
```
Rob-Hague, line 127: Could you also add a simple test which verifies the call to `ChannelSession.SendWindowChangeRequest`?

Rob-Hague (commented):
Could you please change `ChannelSession.SendWindowChangeRequest` to call `TrySendMessage` instead of `SendMessage`?

PR #1515, opened 2024-10-08
Adds DownloadFileAsync, UploadFileAsync and SynchronizeDirectoriesAsync overloads

Havunen:
Hi!

It would be awesome if SSH.NET package could provide these few async overload methods for more ergonomic developer experience. Previously there were implemented here https://github.com/JohnTheGr8/Renci.SshNet.Async but that package has not been updated for a while. https://github.com/JohnTheGr8/Renci.SshNet.Async/pull/17 

Hopefully this could be merged! Thanks!


Havunen:
That failing test seems to flaky, its failing also on other pull requests...

Rob-Hague:
Hi @Havunen 

I am definitely in favour of adding these overloads, but I think they need to have functional cancellation support with a `CancellationToken` parameter. I'm not sure how easy that's going to be.

Secondly, do you know why the tests were failing in the first commit? I think these ought to be investigated before taking the change.

Lastly (once the bigger problems are resolved), the optional parameters relating to task factory/creation etc. should be removed. These would expose details of the implementation which would not be desirable.

Let me know what you think

Havunen:
> Secondly, do you know why the tests were failing in the first commit? I think these ought to be investigated before taking the change.

Yeah I changed them to async - await but realized they were testing something while the upload was in progress so I reverted those

Havunen:
> I am definitely in favour of adding these overloads, but I think they need to have functional cancellation support with a CancellationToken parameter. I'm not sure how easy that's going to be. 

For that we maybe we need to create internal implementation for the methods so the cancellation can be checked there and then synchronous API has that async value as null

prajal55:
Any update on this issue?

Havunen:
I have been busy with all other work ongoing. Maybe somebody else can also add the Cancellation token support, it should not be too difficult there are already some methods using it.

sduffy77:
@Havunen Just checking in to see if this has been implemented or close to being ready? 
If not then I'll just create an extension method to support async.

Havunen:
I have been busy with all the other work, hopefully somebodyelse can add the cancellationToken support

Rob-Hague:
UploadFileAsync and DownloadFileAsync added in #1634. Would take a SynchronizeDirectoriesAsync with a CancellationToken, otherwise will close this PR for now. Thanks for pushing it along!

PR #1529, opened 2024-11-02
Avoid process crash due to unhandled exceptions in sending keepalive messages

SergeyTeplyakov:
`SendKeepAliveMessage` is called from the timer and if the method fails with an unhandled exception then the calling process crashes.

One of the Azure Core services faces those crashes periodically.

Here is the callstack from the crash:

```
mscorlib_ni!System.Runtime.InteropServices.SafeHandle.DangerousAddRef+0x0
mscorlib_ni!System.StubHelpers.StubHelpers.SafeHandleAddRef+0x21 [f:\dd\ndp\clr\src\BCL\system\stubhelpers.cs @ 1781]
System_Core!System.Security.Cryptography.CapiNative+UnsafeNativeMethods
System_Core!System.Security.Cryptography.CapiSymmetricAlgorithm.EncryptBlocks+0x67 [f:\dd\ndp\fx\src\Core\System\Security\Cryptography\CapiSymmetricAlgorithm.cs @ 258]
Renci_SshNet!Renci.SshNet.Security.Cryptography.Ciphers.AesCipher+CtrImpl [D:\dbs\sh\azcm\1022_220603\cmd\47\src\ext\Renci.SshNet\src\Security\Cryptography\Ciphers\AesCipher.CtrImpl.cs @ 77]
Renci_SshNet!Renci.SshNet.Session.SendMessage+0x193 [D:\dbs\sh\azcm\1022_220603\cmd\47\src\ext\Renci.SshNet\src\Session.cs @ 1170]
Renci_SshNet!Renci.SshNet.Session.TrySendMessage+0x13 [D:\dbs\sh\azcm\1022_220603\cmd\47\src\ext\Renci.SshNet\src\Session.cs @ 1371]
Renci_SshNet!Renci.SshNet.BaseClient.SendKeepAliveMessage+0xba [D:\dbs\sh\azcm\1022_220603\cmd\47\src\ext\Renci.SshNet\src\BaseClient.cs @ 492]
mscorlib_ni!System.Threading.ExecutionContext.RunInternal+0x172 [f:\dd\ndp\clr\src\BCL\system\threading\executioncontext.cs @ 980]
mscorlib_ni!System.Threading.ExecutionContext.Run+0x15 [f:\dd\ndp\clr\src\BCL\system\threading\executioncontext.cs @ 928]
mscorlib_ni!System.Threading.TimerQueueTimer.CallCallback+0xba [f:\dd\ndp\clr\src\BCL\system\threading\timer.cs @ 713]
mscorlib_ni!System.Threading.TimerQueueTimer.Fire+0x8b [f:\dd\ndp\clr\src\BCL\system\threading\timer.cs @ 670]
mscorlib_ni!System.Threading.TimerQueue.FireNextTimers+0x78 [f:\dd\ndp\clr\src\BCL\system\threading\timer.cs @ 425]
```

Essentially, there is a race condition in the `Disconnect` method with the timer clean up and the session disposal.
When the timer is disposed it is still possible for the timer callback to be executed by the thread pool thread, thats why the following code in `Disconnect` is not thread safe:

```csharp
public void Disconnect()
{
  // stop sending keep-alive messages before we close the session
  StopKeepAliveTimer();

  // dispose the SSH session
  DisposeSession();
}
```

Potentially, we can solve the race condition in a different way, but probably having a catch block and the trace in the timer callback is a good idea to avoid process crashes.


Rob-Hague:
Please can you target the `develop` branch?

PR #1250, opened 2023-11-21
Test integration tests

WojciechNagorski:

PR #1539, opened 2024-11-16
Migrate from AppVeyor to GitHub Actions

mus65:
This migrates from AppVeyor to GitHub Actions. This initially came up in https://github.com/sshnet/SSH.NET/pull/1480#issuecomment-2474124593 because we would have to wait for AppVeyor to support .NET 9 .

I tried to keep the functionality mostly the same. I made the following changes:

- Replaced AppVeyor.TestLogger with GitHubActionsTestLogger so the test results are displayed in GitHub.
- removed LiquidTestReports.Markdown integration. This was added with #1250 . Not sure how useful this is considering the test results are visible in GitHub directly now. I could add it back.
- removed redundant `-c` and `--no-restore` parameters from dotnet calls
- mono must be installed manually since the runner doesn't provide it out-of-the-box.
- Upgraded AotCompatibilityTestApp to net9.0 already here since net8.0 would result in the following failure. No idea why this doesn't happen in AppVeyor.

```
Run dotnet publish -r win-x64 /warnaserror .\test\Renci.SshNet.AotCompatibilityTestApp\
  Determining projects to restore...
  Restored D:\a\SSH.NET\SSH.NET\test\Renci.SshNet.AotCompatibilityTestApp\Renci.SshNet.AotCompatibilityTestApp.csproj (in 357 ms).
  Restored D:\a\SSH.NET\SSH.NET\src\Renci.SshNet\Renci.SshNet.csproj (in 365 ms).
  Renci.SshNet -> D:\a\SSH.NET\SSH.NET\src\Renci.SshNet\bin\Release\net8.0\Renci.SshNet.dll
  Renci.SshNet.AotCompatibilityTestApp -> D:\a\SSH.NET\SSH.NET\test\Renci.SshNet.AotCompatibilityTestApp\bin\Release\net8.0\win-x64\Renci.SshNet.AotCompatibilityTestApp.dll
  Generating native code
Error: /_/src/libraries/System.Net.Quic/src/System/Net/Quic/Internal/MsQuicApi.cs(90): error IL3000: System.Net.Quic.MsQuicApi..cctor(): 'System.Reflection.Assembly.Location.get' always returns an empty string for assemblies embedded in a single-file app. If the path to the app directory is needed, consider calling 'System.AppContext.BaseDirectory'. [D:\a\SSH.NET\SSH.NET\test\Renci.SshNet.AotCompatibilityTestApp\Renci.SshNet.AotCompatibilityTestApp.csproj]
  Renci.SshNet.AotCompatibilityTestApp -> D:\a\SSH.NET\SSH.NET\test\Renci.SshNet.AotCompatibilityTestApp\bin\Release\net8.0\win-x64\publish\
```

Coverlet results are still only archived but not integrated or displayed anywhere. There is #1257 open for this. There is also some kind of [GitHub Integration](https://josh-ops.com/posts/github-code-coverage/) for this. I could look into this separately.


mus65:
Test results and artifacts: https://github.com/sshnet/SSH.NET/actions/runs/11869304328

It still tries to run AppVeyor here. I assume this must be disabled in the project.

side note: I have not seen a single flaky test failure while testing this. :smile:

Rob-Hague (commented):
A lot faster too. One thing I think is missing at the moment is a CI nuget feed that we get automatically with appveyor.

Unless you want to take a look now, I suggest we just keep only the `dotnet pack` part of appveyor.yml to keep that bit going, and then look to use github packages or similar for a CI feed in a follow-up

(also because I don't think I have permissions to remove the appveyor check so easiest to keep around in the meantime)

Directory.Packages.props
```diff
@@ -4,12 +4,14 @@
...
-    <PackageVersion Include="Appveyor.TestLogger" Version="2.0.0" />
     <PackageVersion Include="BenchmarkDotNet" Version="0.14.0" />
     <PackageVersion Include="BouncyCastle.Cryptography" Version="2.4.0" />
     <PackageVersion Include="coverlet.collector" Version="6.0.2" />
     <PackageVersion Include="coverlet.msbuild" Version="6.0.2" />
-    <PackageVersion Include="LiquidTestReports.Markdown" Version="1.0.9" />
+    <PackageVersion Include="GitHubActionsTestLogger" Version="2.4.1">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageVersion>
```
Rob-Hague, line 14: are these bits needed here?
```suggestion
    <PackageVersion Include="GitHubActionsTestLogger" Version="2.4.1"/>
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,139 @@
+name: Build
+
+on:
+  - push
+  - pull_request
+  - workflow_dispatch
```
Rob-Hague, line 6: does this/can this include tags added as well?

mus65: push also triggers on tag: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#push

> Runs your workflow when you push a commit or tag, or when you create a repository from a template.

Directory.Packages.props
mus65: They may not strictly be needed, but they are set by default for this package. Imho `PrivateAssets` makes sense because the test code itself doesn't actually need this dependency, only the dotnet test host does.

mus65:
I was actually already looking into pushing to the GitHub NuGet Registry. I just finished it up and pushed it here.

NuGet result can be seen here: https://github.com/mus65/SSH.NET/pkgs/nuget/SSH.NET . This will not run for this PR because of the filter on the develop branch.

I will add a minimal appveyor.yml to avoid failure (just tested with an empty file, but that doesn't seem to work).

README.md
```diff
@@ -177,6 +177,19 @@ Private keys in OpenSSH key format can be encrypted using one of the following c
...
+If you need an unreleased bugfix or feature, you can use the Pre-Release NuGet packages from the `develop` branch which are published to the [GitHub NuGet Registry](https://github.com/sshnet/SSH.NET/pkgs/nuget/SSH.NET).
+In order to pull packages from the registry you first have to create a Personal Access Token with the `read:packages` permissions. Then add a NuGet Source for SSH.NET:
+
+Note: you may have to add `--store-password-in-clear-text` on non-Windows platforms.
+
+```
+dotnet nuget add source --name SSH.NET --username <username> --password <personalaccesstoken> https://nuget.pkg.github.com/sshnet/index.json
+```
+
+Then you can add the the package as described [here](https://github.com/sshnet/SSH.NET/pkgs/nuget/SSH.NET).
```
Rob-Hague, line 191: was this intended to be the same link as line 182?

Rob-Hague (approved):

README.md
mus65: yes, I was referring to the "dotnet add package" command that can be copied from there.

mus65:
The NuGet package was pushed successfully on the develop branch, but it's not actually visible. I think this may be related to the Package permissions settings on the organization (which I obviously don't have access to).

Rob-Hague:
Should be good now

Issue #510, opened 2019-01-18
is it possible to connect SFTP using PUTTY KEY string?


vsplcDJ25:
I am trying to connect SFTP using with PUTTY Key string through C# but could not able to connect to the server. received an exception while connecting to SFTP and the exception is **"Invalid Private key file".**
Please let me know with possible solution about using with SSH.NET.

ghost:
Convert the key to OpenSSH format using PuTTYGen. PPK files are (as far as I can tell) not supported.

mgamache:
PPK files are *not* supported. Usually PPK files are derived from PEM or other files that are supported.

Issue #434, opened 2018-05-22
Getting "invalid private key file." error


RahulVPol:
Hi,

I am new to this SFTP communication. I have received the Private_Keys.ppk file from my IT team, which is as follows,
PuTTY-User-Key-File-2: ssh-rsa
Encryption: aes256-cbc
Comment: imported-openssh-key
Public-Lines: 6
AAAAB3NzaC1yc2EAAAADAQABAAABAQCXVRQ8E0LnZ1aFepN0NaLaavip8cO+EJSp
29PBOB2pUCkua/3IXO7q3coziwdJOhPElL6vh3dHDYtuEqrXPhkEli9mruf20ZVj
vI5k4VtlQh5hzjwfgujmLpSWCA0OiH+i8f7y3J3e0PqIDD1E15ZShLWCpzB9YuXM
+Qq2eXs47QMJ9vUYlB6ZcMMftPBp7nQMDni+Z2IHFp4yq1XXo8wRu9pyoKG+j5T9
Vbb1voUCPwgLlcXETt3J3pzS/P+/lake7eTJytJXfhMCn2DovGBxnvHywILLPNaz
HE5FVMxc/ZI8IXqpbGKvY0CRqVNNlZWLRVntjL4IOuDZYMqI1cUJ
Private-Lines: 14
vXQMpVg0cnoA8nkceo5oKf86/QcWFON+Gik1gCPBkus1V60Zd2LVMAHHN0WV8836
OpZzZTUyQrCa3sM/vz9mvQwx1FyZaDNLl6goXK4acneRNHMwblobSH0AGMjCc2x/
WIPtI4gjrwZF8TTNt4eo1kxnW/4E4zilI9U6Kv3h7QTYkvlyA3G0v33y85FQwmfz
maaBi/+Deayp6eGzyEASUqedb6DNVBel7KRs6D2sazywXZZs42zAVxTd+2g8B6D6
cU7a6GXSVqmxMIicYvCT/i6O5cJ+srPLrDK3yqAsOfye5SwgAq8lK3c5q/mEStlG
MHKZfskOXXoaTN9XEZX6w3fxKdMHd4tSSzub4EcpHGfjIGzc55ns5WRQWNa9QJ4o
vb0ImTfhODgU0OHYy6d0kio/u2BTayC543S37cojdBw76eTJOTWtqjoLDr90vrfE
[...]

Issue #431, opened 2018-05-09
Putty Keyfile Regex


LordPinhead:
In Renci.SshNet.PrivateKeyFile i added for Putty an Regex to test it, but putty is missing the encryption and salt, so i could test it friday, but maybe someone could give it a look.

[0001-Testadding-of-Putty-Regex.zip](https://github.com/sshnet/SSH.NET/files/1987927/0001-Testadding-of-Putty-Regex.zip)

Regards
Lord_Pinhead

Issue #341, opened 2017-11-15
Add support for PuTTY key files


TrevorPilley:
One of the connections we want to use SSH.NET for requires us to use a PuTTY private key file, however that isn't currently supported by the SSH.NET library. Would it be possible to add support for a PuTTY key file to the ``Renci.SshNet.PrivateKeyFile`` class?

PR #1542, opened 2024-11-17
Split PrivateKeyFile into different implementations.

scott-xu:
Abstract the implementations into `IPrivateKeyParser`.
It will be helpful when add new parsers, for example, PuTTY.
Prepare for https://github.com/sshnet/SSH.NET/issues/341, https://github.com/sshnet/SSH.NET/issues/431, https://github.com/sshnet/SSH.NET/issues/434, https://github.com/sshnet/SSH.NET/issues/510,

### DIFF
--- a/src/Renci.SshNet/PrivateKeyFile.OpenSSH.cs
+++ b/src/Renci.SshNet/PrivateKeyFile.OpenSSH.cs
@@ -1,0 +1,249 @@
+﻿#nullable enable
+using System;
+using System.Globalization;
+using System.Linq;
+using System.Text;
+
+using Renci.SshNet.Common;
+using Renci.SshNet.Security;
+using Renci.SshNet.Security.Cryptography;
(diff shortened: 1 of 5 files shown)


Rob-Hague (approved):

Rob-Hague:
> It will be helpful when add new parsers, for example, PuTTY.

Are you planning on adding that? I saw some argon2-related stuff added to BouncyCastle

cc @darinkes

darinkes:
Would be neat for sure to have it as part of the main project.

I use https://github.com/kmaragon/Konscious.Security.Cryptography since I don't use the whole BouncyCastle Lib.
But should be easy to replace the few bits with BouncyCastle, I think.

scott-xu:
I also noted argon2 is added into BouncyCastle.
I'll have a try to add PuTTY support when I have free time.

PR #1545, opened 2024-11-30
Add padding when encrypt and remove padding when decrypt

scott-xu:
* Add PKCS7 padding when encrypt no matter the input is divisible or not if pkcs7padding is specified.
* Add manual padding when encrypt if input is not divisible and pkcs7padding is not specified and cipher mode is CFB or OFB.
* Remove PKCS7 padding when decrypt if pkcs7padding is specified.
* Remove manual padding when decrypt if pkcs7padding is not specified and there **was** manual padding added.

Originally I was to use BCL TripleDES/DES to replace managed implementation. Then I found the padding is wrong and was going to make a fix. During the fix, I noticed that I'm on the same path with @Rob-Hague and @zybexXL 😄 See https://github.com/sshnet/SSH.NET/pull/1238 and https://github.com/zybexXL/SSH.NET/compare/feature_AES_CSP...zybexXL:SSH.NET:fix_padding?diff=split


src/Renci.SshNet/Security/Cryptography/BlockCipher.cs
```diff
@@ -75,18 +77,29 @@ protected BlockCipher(byte[] key, byte blockSize, CipherMode mode, CipherPadding
...
+            else if (length % _blockSize > 0)
+            {
+                if (_mode is CfbCipherMode or OfbCipherMode or CtrCipherMode)
+                {
+                    paddingLength = _blockSize - (length % _blockSize);
+                    input = input.Take(offset, length);
+                    length += paddingLength;
+                    Array.Resize(ref input, length);
+                    offset = 0;
+                }
```
scott-xu, line 97: There are 2 options here if the cipher mode is CFB, OFB or CTR.
1. Manually expand the input before encrypt/decrypt and shrink after encrypt/decrypt
2. Update [Cfb|Ofb|Ctr]CipherMode to allow input length less than block size.

I choose option1 because BCL does not allow partial block as well, regardless cipher mode.
See https://github.com/dotnet/runtime/blob/e7d837da5b1aacd9325a8b8f2214cfaf4d3f0ff6/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/SymmetricPadding.cs#L20-L21

src/Renci.SshNet/Security/Cryptography/Ciphers/CipherPadding.cs
```diff
@@ -54,5 +54,12 @@ public byte[] Pad(byte[] input, int paddinglength)
         /// The padded data array.
         /// </returns>
         public abstract byte[] Pad(byte[] input, int offset, int length, int paddinglength);
+
+        /// <summary>
+        /// Gets the padd count from the specified input.
+        /// </summary>
+        /// <param name="input">The input.</param>
+        /// <returns>The padd count.</returns>
+        public abstract int PadCount(byte[] input);
```
scott-xu, line 63: We can consider replacing `CipherPadding` and `PKCS7Padding` with BouncyCastle's `Pkcs7Padding`

scott-xu:
There would be pros and cons about different ways to fix the issue, but anyway, I think we all agree that it is an issue and should be fixed. @Rob-Hague @zybexXL

Rob-Hague:
I have started reviewing this, but I need more time to refamiliarise myself in this area. I have been a bit busy lately

scott-xu:
Thanks! Reviewing is usually more difficult than DIY. No hurry, take your time.

src/Renci.SshNet/Security/Cryptography/BlockCipher.cs
```diff
@@ -154,6 +177,16 @@ public override byte[] Decrypt(byte[] input, int offset, int length)
                 throw new InvalidOperationException("Encryption error.");
             }
 
+            if (_padding is not null)
+            {
+                paddingLength = _padding.PadCount(output);
```
zybexXL, line 182: I added an Unpad() method on my original commit, maybe that's useful to have?
https://github.com/zybexXL/SSH.NET/compare/feature_AES_CSP...zybexXL:SSH.NET:fix_padding?diff=split#diff-2d0250a9e26828476447e03e6a5f970f5a3e1711b09a17b81231c33e1d5c253eR53

zybexXL (approved):
Looks good to me 👍

src/Renci.SshNet/Security/Cryptography/BlockCipher.cs
scott-xu: I borrowed `PadCount` method from BouncyCastle. It not only gets the padding length from reading the last byte, but also checks if every padding byte equals to the last byte.
I use `PadCount` instead of `UnPad` in case of replacing it with BouncyCastle in the future.

src/Renci.SshNet/Security/Cryptography/Ciphers/CipherPadding.cs
Rob-Hague: yes, I think we should take this opportunity to delete the useless `CipherPadding` and replace it with BC `IBlockCipherPadding`, and delete `PKCS7Padding` and replace it with BC `Pkcs7Padding`, if possible

test/Renci.SshNet.Tests/Classes/Security/Cryptography/Ciphers/AesCipherTest.Gen.cs.txt
```diff
@@ -21,109 +21,130 @@ foreach ((string mode, (string modeCode, CipherMode? bclMode)) in modes)
...
-        foreach (int inputLength in new int[] { 16, 32, 64 })
+        foreach (int inputLength in new int[] { 16, 35, 64 })
         {
-            byte[] input = new byte[inputLength];
-            random.NextBytes(input);
+            foreach (bool pad in new bool[] { false, true })
+            {
+                // It is not allowed to use no padding on non-block lengths
+                // It makes sense in cfb, ctr and ofb modes
+                if (!pad && inputLength % 16 != 0 && mode is not "cfb" or "ctr" or "ofb")
```
Rob-Hague: supposed to be

```suggestion
                if (!pad && inputLength % 16 != 0 && mode is not ("cfb" or "ctr" or "ofb"))
```

?

Rob-Hague (commented):
looks good

test/Renci.SshNet.Tests/Classes/Security/Cryptography/Ciphers/AesCipherTest.Gen.cs.txt
scott-xu: Yes, it is a little bit interesting. The word `or` here is for and **only for** `mode` which is not the same with `||` operator. With additional brackets, it might be easier for people who codes; without additional brackets, it might be easier for people who does not code.

src/Renci.SshNet/Security/Cryptography/Ciphers/CipherPadding.cs
scott-xu: Let me have a try

test/Renci.SshNet.Tests/Classes/Security/Cryptography/Ciphers/AesCipherTest.Gen.cs.txt
Rob-Hague: they have different behaviour due to precedence of the `not`:

```c#
int x = 3;

Console.WriteLine(x is not 1 or 2 or 3); // True
Console.WriteLine(x is not (1 or 2 or 3)); // False
```

It is not outputting e.g. `AES_CTR_128_Length35_NoPad` which I assume is not intended

src/Renci.SshNet/Security/Cryptography/Ciphers/CipherPadding.cs
scott-xu: Just had a quick look. It would be easier to remove it in my next PR https://github.com/sshnet/SSH.NET/pull/1546.

test/Renci.SshNet.Tests/Classes/Security/Cryptography/Ciphers/AesCipherTest.Gen.cs.txt
scott-xu: Ah I see. Thanks a lot!

Rob-Hague: will merge once the tests are regenerated

Rob-Hague (approved):

PR #1546, opened 2024-11-30
Use System.Security.Cryptography for TripleDesCipher

scott-xu:
This PR 
 - changes `TripleDesCipher` to use BCL's `TripleDES`.
 - drops support for `DesCipher`. It is only used to decrypt OpenSSL legacy private key format (aka PKCS1). Suggest switching to newer encryption method.
 - changes *PKCS7Padding* to use BouncyCaslte's implementation.
 - ~~renames `AesCipherMode` to `BlockCipherMode`. Some previous discussion: https://github.com/sshnet/SSH.NET/pull/865#discussion_r1407557172~~

### Benchmarks based on .NET 9.0.0 (9.0.24.52809), X64 RyuJIT AVX2

Before:

| Method      | Mean     | Error     | StdDev    | Gen0    | Gen1   | Allocated |
|------------ |---------:|----------:|----------:|--------:|-------:|----------:|
| Encrypt_CBC | 1.379 ms | 0.0268 ms | 0.0402 ms | 39.0625 | 1.9531 | 161.19 KB |
| Decrypt_CBC | 1.333 ms | 0.0202 ms | 0.0189 ms | 39.0625 | 1.9531 | 161.19 KB |
| Encrypt_CFB | 1.371 ms | 0.0263 ms | 0.0281 ms | 39.0625 | 1.9531 | 161.23 KB |
| Decrypt_CFB | 1.379 ms | 0.0164 ms | 0.0146 ms | 39.0625 | 1.9531 | 161.23 KB |

After:

| Method      | Mean       | Error    | StdDev   | Gen0   | Allocated |
|------------ |-----------:|---------:|---------:|-------:|----------:|
| Encrypt_CBC | 1,018.1 us | 20.33 us | 28.50 us | 7.8125 |  33.18 KB |
| Decrypt_CBC |   966.5 us | 19.25 us | 20.60 us | 7.8125 |  33.18 KB |
| Encrypt_CFB | 1,016.6 us | 15.57 us | 14.57 us | 7.8125 |  33.18 KB |
| Decrypt_CFB |   970.4 us | 13.38 us | 12.52 us | 7.8125 |  33.18 KB |


Rob-Hague:
We could consider dropping DesCipher

scott-xu:
Agree

scott-xu:
Might also be a good chance to drop home-made `CfbCipherMode` and `OfbCipherMode`.

 - `CfbCipherMode` is only used for PKCS1 private key encrypted by`DES-EDE3-CFB`.
 - `OfbCipherMode` is not used anywhere **inside** the library.

Further more, we can consider changing all Renci.SshNet.Security.Cryptography.* internal.

Rob-Hague:
possibly yes, but I would prefer to think about that separately. Can you restore CbcCipherMode here?

scott-xu:
CbcCipherMode is restored.

src/Renci.SshNet/Security/Cryptography/Ciphers/TripleDesCipher.BouncyCastleImpl.cs
Rob-Hague, line 1: can we not use BCL in ECB mode as in AesCipher.BlockImpl which is used for AES-CFB on lower targets?

src/Renci.SshNet/Security/Cryptography/Ciphers/TripleDesCipher.cs
```diff
@@ -1,154 +1,86 @@
...
         /// <summary>
         /// Initializes a new instance of the <see cref="TripleDesCipher"/> class.
         /// </summary>
         /// <param name="key">The key.</param>
+        /// <param name="iv">The IV.</param>
         /// <param name="mode">The mode.</param>
-        /// <param name="padding">The padding.</param>
+        /// <param name="pkcs7Padding">Enable PKCS7 padding.</param>
         /// <exception cref="ArgumentNullException"><paramref name="key"/> is <see langword="null"/>.</exception>
-        public TripleDesCipher(byte[] key, CipherMode mode, CipherPadding padding)
```
Rob-Hague, line 24: if we are not going to support CTR mode, then we might as well pass in System.Security.Cryptography.CipherMode and avoid renaming AesCipherMode

Rob-Hague (approved):

PR #1550, opened 2024-12-03
fix newline characters in comment

Varorbc:


Rob-Hague:
Thanks! are you able to modify one of the .ppk files in test/Data in order to demonstrate the problem and fix?

Varorbc:
The current test case has this problem because it is a ppk file obtained from an embedded resource, so it works fine. This problem occurs if you read directly from disk. So how do I add test cases? I hope you can give me some advice.

scott-xu:
@Varorbc  Just out of curiosity, what's the **binary difference** between embedded resource and loose file?

Varorbc:
@scott-xu Sorry about that, I got the issue mixed up earlier. Let me correct it: The PPK files in the current test cases use `\n` for new lines, but the PPK files generated by the older PuTTYgen client use `\r\n`.

Rob-Hague:
right, because the comment is included in the MAC, so we don't want to greedily match the \r. Thanks!

FYI: the changed .ppk file still has LF line endings because we normalize these files:

https://github.com/sshnet/SSH.NET/blob/3b4f2cfc1c6c0da935faf1c601b20026dceda473/.gitattributes#L21

but the problem is understood well enough

Rob-Hague (approved):

PR #1553, opened 2024-12-09
Run integration tests on alpine:latest

Rob-Hague:
edit: the integration tests started failing with Alpine Linux 3.21. This tracks setting the image back to alpine:latest after #1554


Rob-Hague:
Integration tests failing with "Killing sshd.pam service failed with exit code 1" e.g. 

```
  Failed Common_DisposeAfterLossOfNetworkConnectivity [900 ms]
  Error Message:
   Test method Renci.SshNet.IntegrationTests.ConnectivityTests.Common_DisposeAfterLossOfNetworkConnectivity threw exception: 
System.ApplicationException: Killing sshd.pam service failed with exit code 1.


  Stack Trace:
      at Renci.SshNet.IntegrationTests.SshConnectionDisruptor.PauseSshd(SshClient client) in /home/runner/work/SSH.NET/SSH.NET/test/Renci.SshNet.IntegrationTests/SshConnectionDisruptor.cs:line 36
   at Renci.SshNet.IntegrationTests.SshConnectionDisruptor.BreakConnections() in /home/runner/work/SSH.NET/SSH.NET/test/Renci.SshNet.IntegrationTests/SshConnectionDisruptor.cs:line 18
   at Renci.SshNet.IntegrationTests.ConnectivityTests.Common_DisposeAfterLossOfNetworkConnectivity() in /home/runner/work/SSH.NET/SSH.NET/test/Renci.SshNet.IntegrationTests/ConnectivityTests.cs:line 75
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

Started between 4Dec and 7Dec. Looks like it is caused by alpine:latest bumping to [3.21](https://alpinelinux.org/posts/Alpine-3.21.0-released.html). It uses OpenSSH 9.9

Rob-Hague:
works with Alpine Linux 3.20 & OpenSSH 9.7

can reproduce locally now (my "alpine:latest" was still at 3.18)

test/Renci.SshNet.IntegrationTests/SshConnectionDisruptor.cs
```diff
@@ -22,19 +22,24 @@ public SshConnectionRestorer BreakConnections()
...
+                {
+                    throw new ApplicationException(
+                        $"Blocking user sshnet failed with exit code {command.ExitStatus}.\r\n{output}\r\n{command.Error}");
+                }
             }
-            command = client.CreateCommand("sudo pkill -9 -U sshnet -f sshd.pam");
-            output = command.Execute();
-            if (command.ExitStatus != 0)
+
+            using (var command = client.CreateCommand("sudo pkill -9 -U sshnet -f sshd-session.pam"))
```
Rob-Hague, line 35: sshd.pam -> sshd-session.pam is the fix for OpenSSH 9.8+. However, that release also completely removes DSA at compile time, so now the DSA tests fail.

I think it is time to remove DSA

test/Renci.SshNet.IntegrationTests/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:3.20
+FROM alpine:latest
```
mus65, line 1: Didn't the original issue show that using latest is a bad idea? It made CI suddenly fail without an obvious indication why. Imho it would make more sense to tag it to 3.21.0 (even the patch) and use Dependabot to keep it up-to-date.

Rob-Hague: This is the first such issue in ~1.5 years of these integration tests. I don't think we necessarily need to make a process out of it, but if it's simple enough to do then sure

Rob-Hague: well it was simple enough

PR #1562, opened 2024-12-26
[PQC] Add support for sntrup761x25519-sha512 key exchange method

scott-xu:
This PR adds support for sntrup761x25519Sha512 hybrid key exchange method described at https://www.ietf.org/archive/id/draft-ietf-sshm-ntruprime-ssh-01.html

### DIFF
--- a/test/Renci.SshNet.IntegrationTests/KeyExchangeAlgorithmTests.cs
+++ b/test/Renci.SshNet.IntegrationTests/KeyExchangeAlgorithmTests.cs
@@ -23,6 +23,37 @@ namespace Renci.SshNet.IntegrationTests
         }
 
         [TestMethod]
+        [Ignore]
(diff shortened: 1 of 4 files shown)


test/Renci.SshNet.IntegrationTests/KeyExchangeAlgorithmTests.cs
```diff
@@ -22,6 +22,37 @@ public void TearDown()
             _remoteSshdConfig?.Reset();
         }
 
+        [TestMethod]
+        [Ignore]
```
scott-xu, line 26: The OpenSSH at Ubuntu is not the latest version.
My local test passed with OpenSSH 9.9p1.

Rob-Hague (approved):

PR #1563, opened 2024-12-27
[PQC] Add support for mlkem768x25519-sha256 key exchange method

scott-xu:
This PR adds support for mlkem768x25519-sha256 hybrid post-quantum key exchange method based on the FIPS 203 Module-Lattice Key Encapsulation mechanism (ML-KEM) combined with X25519 ECDH as described by https://www.ietf.org/archive/id/draft-kampanakis-curdle-ssh-pq-ke-05.html

OpenSSH added support for this key exchange method in release 9.9,  see https://www.openssh.com/txt/release-9.9.


test/Renci.SshNet.IntegrationTests/KeyExchangeAlgorithmTests.cs
```diff
@@ -22,6 +22,22 @@ public void TearDown()
             _remoteSshdConfig?.Reset();
         }
 
+        [TestMethod]
+        [Ignore]
```
scott-xu: The OpenSSH at Ubuntu is not the latest version.
My local test passed with OpenSSH 9.9p1.

Rob-Hague (approved):

Issue #1560, opened 2024-12-24
Consider dropping hand-written cipher modes


scott-xu:
We could consider dropping hand-written cipher modes in the library.
1. `CbcCipherMode` will not be used anymore once PR https://github.com/sshnet/SSH.NET/pull/1546 is merged.
2. `CfbCipherMode` is only used for PKCS1 private key encrypted byDES-EDE3-CFB. Suggest user switching to newer encryption method.
3. `OfbCipherMode` is not used anywhere inside the library.
4. `CtrCipherMode` is used for `AesCipher`. We can create a dedicated class `AesCtrCipher`, just like `AesGcmCipher`.

With all above concrete classes removed, we can remove the abstract class `CipherMode`. We can also remove `AesCipherMode` enum ~~(renamed as `BlockCipherMode` in PR https://github.com/sshnet/SSH.NET/pull/1546)~~ and use BCL's `CipherMode` enum.

Then `AesCipher.BlockImpl` can be removed, `AesCipher.CtrImpl` can be removed (moved to a dedicated class per above suggestion).
`AesCipher.BclImpl` can be removed so that `AesCipher` can have a pure BCL implementation (without nested `Impl` classes)

So does the `TripleDesCipher`.

scott-xu:
Ideally, there's only below ciphers in Ciphers folder
- AesCipher
- AesCtrCipher
- AesGcmCipher
- ChaCha20Poly1305Cipher
- TripleDesCipher

scott-xu:
Once https://github.com/dotnet/runtime/issues/89718 is implemented, we don't need BouncyCastle for AesGcmCipher

Rob-Hague:
I'd be fine with dropping the `CipherMode` classes now. `BlockCipher` could also be removed, but `BlockCipher.BlockSize` should stay (I've seen it used in the wild). It could be moved down to `Cipher` as `public virtual int BlockSize => MinimumSize`

I would prefer not to change the public API of `AesCipher`/`AesCipherMode` as in #1564 - I still don't think it is worth it

scott-xu:
> I would prefer not to change the public API of AesCipher/AesCipherMode as in https://github.com/sshnet/SSH.NET/pull/1564 - I still don't think it is worth it

`AesCipherMode` is code smell or even wrong concept. Cipher mode is not limited to AES only. I know it is public API but we changed public API before. I think it should be fine if we write down the change in release notes.

PR #1564, opened 2024-12-28
Use BCL CipherMode enum for AesCipher class

scott-xu:
Create a dedicated AesCtrCipher class just like AesGcmCipher class.

Relate to https://github.com/sshnet/SSH.NET/issues/1560


Rob-Hague:
Is there a benefit to this?

scott-xu:
FYI: some previous discussion: https://github.com/sshnet/SSH.NET/pull/865#discussion_r1399560195

scott-xu:
The main purpose is to use BCL as possible as it can.

Rob-Hague:
Thanks, I remember the previous discussion, but I don't think it is worth changing the API right now. Maybe at a later time

scott-xu:
> Is there a benefit to this?

One other benefit is the reduction of code. Less code means less maintenance.
image

src/Renci.SshNet/Security/Cryptography/Ciphers/AesCipher.cs
```diff
@@ -1,18 +1,27 @@
...
-using Renci.SshNet.Security.Cryptography.Ciphers.Modes;
+using Renci.SshNet.Common;
 
 namespace Renci.SshNet.Security.Cryptography.Ciphers
 {
     /// <summary>
     /// AES cipher implementation.
     /// </summary>
-    public sealed partial class AesCipher : BlockCipher, IDisposable
+    internal sealed class AesCipher : SymmetricCipher, IDisposable
```
scott-xu: Note: this is a breaking change of public API

src/Renci.SshNet/Security/Cryptography/Ciphers/TripleDesCipher.cs
```diff
@@ -1,24 +1,26 @@
...
-#endif
+using Renci.SshNet.Common;
 
 namespace Renci.SshNet.Security.Cryptography.Ciphers
 {
     /// <summary>
     /// Implements 3DES cipher algorithm.
     /// </summary>
-    public sealed partial class TripleDesCipher : BlockCipher, IDisposable
+    internal sealed class TripleDesCipher : SymmetricCipher, IDisposable
```
scott-xu: This is a breaking change of public API

scott-xu:
IMHO, SSH.NET should be the client of SSH but not a cryptography library. That's the reason why I _hide_ `AesCipher` and `TripleDesCipher`.

scott-xu:
Could you please have another review of this PR please? @Rob-Hague

Rob-Hague:
I am fine with dropping BlockCipher (and CipherMode), but I would rather not change AesCipher until there is a compelling reason, otherwise it is just churn. If you can put it back, I'll merge the rest

scott-xu:
The reason is
1. BCL `CipherMode` does not have `CTR` mode.
2. `OFB` and `CFB` is not used anywhere.

src/Renci.SshNet/Security/Cryptography/Ciphers/AesCipher.cs
```diff
@@ -23,67 +32,65 @@ public sealed partial class AesCipher : BlockCipher, IDisposable
...
-        public AesCipher(byte[] key, byte[] iv, AesCipherMode mode, bool pkcs7Padding = false)
-            : base(key, 16, mode: null, padding: null)
+        public AesCipher(byte[] key, byte[] iv, CipherMode mode, bool pkcs7Padding = false)
+            : base(key)
         {
-            if (mode == AesCipherMode.OFB)
-            {
-                // OFB is not supported on modern .NET
-                _impl = new BlockImpl(key, new OfbCipherMode(iv), pkcs7Padding ? new Pkcs7Padding() : null);
-            }
```
scott-xu, line 33: `OFB` is not used anywhere

src/Renci.SshNet/Security/Cryptography/Ciphers/AesCipher.cs
```diff
@@ -23,67 +32,65 @@ public sealed partial class AesCipher : BlockCipher, IDisposable
...
-#if !NET
-            else if (mode == AesCipherMode.CFB)
-            {
-                // CFB not supported on NetStandard 2.1
-                _impl = new BlockImpl(key, new CfbCipherMode(iv), pkcs7Padding ? new Pkcs7Padding() : null);
-            }
```
scott-xu, line 39: `CFB` is not used anywhere

src/Renci.SshNet/Security/Cryptography/Ciphers/AesCipher.cs
```diff
@@ -23,67 +32,65 @@ public sealed partial class AesCipher : BlockCipher, IDisposable
...
-#endif
-            else if (mode == AesCipherMode.CTR)
-            {
-                // CTR not supported by the BCL, use an optimized implementation
-                _impl = new CtrImpl(key, iv);
-            }
```
scott-xu, line 45: BCL `CipherMode` does not have `CTR` mode

scott-xu:
BTW, I would suggest renaming `Renci.SshNet.Security.Cryptography.SymmetricCipher` to `Renci.SshNet.Security.Cryptography.AsymmetricBlockCipher`. Unfortunately, it is a public class. But it should be rarely used. What do you think? @Rob-Hague

scott-xu:
> I am fine with dropping BlockCipher (and CipherMode), but I would rather not change AesCipher until there is a compelling reason, otherwise it is just churn. If you can put it back, I'll merge the rest

`BlockCipher` and `CipherMode` are dropped. `BlockSize` property is moved to `Cipher` abstract class.
Can you please share more information about not changing `AesCipher`?

scott-xu:
> BTW, I would suggest renaming `Renci.SshNet.Security.Cryptography.SymmetricCipher` to `Renci.SshNet.Security.Cryptography.AsymmetricBlockCipher`. Unfortunately, it is a public class. But it should be rarely used. What do you think? @Rob-Hague

`SymmertricCipher` is dropped.

PR #1577, opened 2025-01-19
Update bound port on ForwardedPortDynamic after connection (in case original was passed as zero)

### LINKED ISSUE #1575: Offer/expose dynamic port assignment for ForwardedPortDynamic
I understand that in some TCP stacks, if you listen on port "0", a randomly-chosen and available port will be provided for you.

I would like to be able to use `port=0` in ForwardedPortDynamic, to guarantee that an available port is chosen, and be able to use the actually bound port in later code. [`ForwardedPortDynamic` does not update `BoundHost` or `BoundPort`](https://github.com/sshnet/SSH.NET/blob/7e71bb419275d4d97599c8e356359c348457907b/src/Renci.SshNet/ForwardedPortDynamic.cs#L169) after the socket starts listening.

```csharp
using var sshClient = new SshClient(sshHost, sshPort, sshUser, sshPass);
await sshClient.ConnectAsync(token);

// I would like to do this
using var portForward = new ForwardedPortDynamic(0); 

// my current solution to choose a random (and hopefully available) port
// using var portForward = new ForwardedPortDynamic((uint)System.Random.Shared.Next(1024, 65535)); 

sshClient.AddForwardedPort(portForward);
portForward.Start();

// Use FluentFTP to connect to a remote server via the SOCKS proxy created by SSH.NET
var ftpClient = new AsyncFtpClientSocks5Proxy(new FtpProxyProfile()
{
    ProxyHost = "127.0.0.1",
    ProxyPort = (int)portForward.BoundPort,        // I need the port here, but it is still "0"
    FtpHost = ftpHost,
    FtpPort = (int)ftpPort,
    FtpCredentials = new NetworkCredential(ftpUser, ftpPass),
});

await ftpClient.Connect()

...

```

Is there a way to bind a random port that is guaranteed to be available and re-use it later?

timyhac:
Update the bound port after connection (in case original was passed as zero).
The logic was copied from ForwardedPortLocal into ForwardedPortDynamic.

### DIFF
--- a/src/Renci.SshNet/ForwardedPortDynamic.cs
+++ b/src/Renci.SshNet/ForwardedPortDynamic.cs
@@ -40,7 +40,7 @@ namespace Renci.SshNet
         /// <summary>
         /// Gets the bound port.
         /// </summary>
-        public uint BoundPort { get; }
+        public uint BoundPort { get; private set; }
 
         private Socket _listener;
         private CountdownEvent _pendingChannelCountdown;
--- a/src/Renci.SshNet/ForwardedPortDynamic.cs
+++ b/src/Renci.SshNet/ForwardedPortDynamic.cs
@@ -168,6 +168,9 @@ namespace Renci.SshNet
             _listener.Bind(ep);
             _listener.Listen(5);
 
+            // update bound port (in case original was passed as zero)
+            BoundPort = (uint)((IPEndPoint)_listener.LocalEndPoint).Port;
+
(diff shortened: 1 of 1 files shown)


Rob-Hague (approved):
Thanks! I wrote a couple of tests over the weekend (not submitted yet) so nothing further needed

timyhac:
Cheers! Thanks for the quick response!

Issue #1253, opened 2023-11-29
Unstable integration test Test_Sftp_Multiple_Async_Upload_And_Download_10Files_5MB_Each


WojciechNagorski:
```
Failed Test_Sftp_Multiple_Async_Upload_And_Download_10Files_5MB_Each [9 s]
  Error Message:
   Assert.IsTrue failed. Uploaded and downloaded bytes does not match
  Stack Trace:
     at Renci.SshNet.IntegrationTests.OldIntegrationTests.SftpClientTest.Test_Sftp_Multiple_Async_Upload_And_Download_10Files_5MB_Each() in /home/appveyor/projects/ssh-net/test/Renci.SshNet.IntegrationTests/OldIntegrationTests/SftpClientTest.Upload.cs:line 233
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
  Standard Output Messages:
 SSH Server host name: 127.0.0.1
 SSH Server port: 32769
 5iovfbli.ykm
 UploadedBytes: 5242880
 DownloadedBytes: 5242880
 UploadedHash: 1b829b119bfeb98e8446755911d7d680
 DownloadedHash: 1b829b119bfeb98e8446755911d7d680
 glzjnmjw.iiq
 UploadedBytes: 5242880
 DownloadedBytes: 5242880
 UploadedHash: 8e0c19f0a0a8d291e9ced2a5e608e322
 DownloadedHash: 8e0c19f0a0a8d291e9ced2a5e608e322
 quhkvyug.zy3
 UploadedBytes: 5242880
 DownloadedBytes: 5242880
 UploadedHash: 4c70cfc9e8f4aa12f6a793042fdba776
 DownloadedHash: 4c70cfc9e8f4aa12f6a793042fdba776
 ypghiybv.4cm
 UploadedBytes: 5242880
 DownloadedBytes: 5242880
 UploadedHash: d3c949929[381](https://ci.appveyor.com/project/drieseng/ssh-net/build/job/laf9893flglsi6rg#L381)1b03588f857761f407f5
 DownloadedHash: d3c9499293811b03588f857761f407f5
 x4db0pr0.dhb
 UploadedBytes: 5242880
 DownloadedBytes: 5242880
 UploadedHash: e658a12e0ff01f0d035b4a994c82c491
[...]

PR #1580, opened 2025-02-02
Drop net6.0 target

mus65:
.NET 6 is EOL since last November. Also simplified a lot of .NET version specific conditions in the process.


mus65:
Test failure is unrelated, see #1253

Directory.Build.props
```diff
@@ -31,7 +31,7 @@
   <!--
         Disable nullable warnings on old frameworks because of missing annotations.
   -->
-  <PropertyGroup Condition=" !$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0')) ">
+  <PropertyGroup Condition=" !$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0')) ">
```
Rob-Hague, line 34:
```suggestion
  <PropertyGroup Condition=" '$(TargetFrameworkIdentifier)' != '.NETCoreApp' ">
```

src/Renci.SshNet/Common/TaskToAsyncResult.cs
```diff
@@ -1,5 +1,5 @@
 ﻿#pragma warning disable
-#if !NET8_0_OR_GREATER
+#if !NET
 // Copied verbatim from https://github.com/dotnet/runtime/blob/78bd7debe6d8b454294c673c9cb969c6b8a14692/src/libraries/Common/src/System/Threading/Tasks/TaskToAsyncResult.cs
```
Rob-Hague:
```suggestion
// Copied verbatim from https://github.com/dotnet/runtime/blob/261611930d6b436d7c4395450356b624d903d9bf/src/libraries/Common/src/System/Threading/Tasks/TaskToAsyncResult.cs
```

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.BclImpl.cs
```diff
@@ -1,4 +1,4 @@
-#if NET6_0_OR_GREATER
+#if NET
```
Rob-Hague, line 1: the remaining ifdefs in this file can be removed

Directory.Build.props
mus65: This doesn't actually work. Seems like this is not defined yet in Directory.Build.props, it does work in Directory.Build.targets though, moved it there. See also https://github.com/dotnet/runtime/blob/4bd597ffde128555b4ff017e87c60adc2fedd178/src/libraries/Directory.Build.targets#L22 .

Rob-Hague: huh, it looked like it worked locally before I suggested it 🤷

Rob-Hague (approved):

Directory.Build.props
mus65: Did you actually try to trigger a nullable warning? Building works fine with your suggestion, but is disables the warnings for modern .NET as well.

Rob-Hague: yeah, I inverted the condition initially to test. No matter. I will check this PR is OK with the others before merging

src/Renci.SshNet/Renci.SshNet.csproj
```diff
@@ -36,11 +36,7 @@
     <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" $([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0')) ">
-    <IsTrimmable>true</IsTrimmable>
-  </PropertyGroup>
-  
-  <PropertyGroup Condition=" $([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0')) ">
+  <PropertyGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETCoreApp' ">
```
Rob-Hague: I am actually not sure now that this works. I will do some tests and if I can't convince myself then I will revert the TargetFrameworkIdentifier changes. Not sure where I got that from

Rob-Hague: The following patch successfully fails the AotCompatibilityTestApp publish on both branches, but on `develop` I also get a diagnostic in VS which does not appear on this branch. So I will revert those changes. Sorry for the noise

```diff
diff --git a/src/Renci.SshNet/SshCommand.cs b/src/Renci.SshNet/SshCommand.cs
index f900e843..9bcafd85 100644
--- a/src/Renci.SshNet/SshCommand.cs
+++ b/src/Renci.SshNet/SshCommand.cs
@@ -214,6 +214,8 @@ public string Error
         /// <exception cref="ArgumentNullException">Either <paramref name="session"/>, <paramref name="commandText"/> is <see langword="null"/>.</exception>
         internal SshCommand(ISession session, string commandText, Encoding encoding)
         {
+            _ = Activator.CreateInstance(session.GetType());
+
             ThrowHelper.ThrowIfNull(session);
             ThrowHelper.ThrowIfNull(commandText);
             ThrowHelper.ThrowIfNull(encoding);
```

PR #1581, opened 2025-02-02
Configure Dependabot to also update NuGet and GitHub Actions

mus65:
This should make it easier to keep things up-to-date .


Rob-Hague (approved):

Rob-Hague:
OK, perhaps it is worth grouping the test dependencies together...

PR #1606, opened 2025-03-01
remove Reverse extension to avoid source-breaking change with .NET 10

mus65:
The internal array `Reverse` extension is already a little weird because it can be ambigious with Linq's `Reverse(IEnumerable<T>)`, but behaves differently.  It changes the passed array instance instead of creating a new one and returns an array instead of an `IEnumerable<T>`.

.NET 10 introduces a [`Enumerable.Reverse<T>(T[])`](https://github.com/dotnet/runtime/issues/107723) which causes even more ambiguity and returns an `IEnumerable<T>` instead of an array, which causes some compilation errors with .NET 10:

```
SSH.NET/test/Renci.SshNet.Benchmarks/Common/ExtensionsBenchmarks.cs(24,26): error CS0121: The call is ambiguous between the following methods or properties: 'System.Linq.Enumerable.Reverse<TSource>(TSource[])' and 'Renci.SshNet.Common.Extensions.Reverse<T>(T[])'
SSH.NET/test/Renci.SshNet.Tests/Classes/ForwardedPortDynamicTest_Stop_PortStarted_ChannelBound.cs(211,81): error CS0121: The call is ambiguous between the following methods or properties: 'System.Linq.Enumerable.Reverse<TSource>(TSource[])' and 'Renci.SshNet.Common.Extensions.Reverse<T>(T[])'
SSH.NET/test/Renci.SshNet.Tests/Classes/ForwardedPortDynamicTest_Dispose_PortStarted_ChannelBound.cs(224,81): error CS0121: The call is ambiguous between the following methods or properties: 'System.Linq.Enumerable.Reverse<TSource>(TSource[])' and 'Renci.SshNet.Common.Extensions.Reverse<T>(T[])'
SSH.NET/test/Renci.SshNet.Tests/Classes/ForwardedPortDynamicTest_SessionErrorOccurred_ChannelBound.cs(231,81): error CS0121: The call is ambiguous between the following methods or properties: 'System.Linq.Enumerable.Reverse<TSource>(TSource[])' and 'Renci.SshNet.Common.Extensions.Reverse<T>(T[])'
```

Looking at the usage, the `Reverse`extension is only needed in the context of `BigInteger` which already has extension methods to deal with endianness (but which were not used everywhere). So I think it makes most sense to just remove the `Reverse` extension.

Also introduced a `ToBigInteger(this ReadOnlySpan<byte> data)` which is made use of in `EcdsaKey.cs`.


src/Renci.SshNet/Common/SshDataStream.cs
```diff
@@ -206,12 +206,7 @@ public void WriteBinary(byte[] buffer, int offset, int count)
         public BigInteger ReadBigInt()
         {
             var data = ReadBinary();
-
-#if NETSTANDARD2_1 || NET
-            return new BigInteger(data, isBigEndian: true);
-#else
-            return new BigInteger(data.Reverse());
-#endif
+            return data.ToBigInteger();
```
Rob-Hague: This will result in an extra array copy on lower targets - that's why the extension is not used here, because we are fine with reversing the original array

src/Renci.SshNet/Security/Cryptography/EcdsaKey.cs
```diff
@@ -147,11 +147,7 @@ public override BigInteger[] Public
                 Buffer.BlockCopy(qy, 0, q, qx.Length + 1, qy.Length);
 
                 // returns Curve-Name and x/y as ECPoint
-#if NETSTANDARD2_1 || NET
-                return new[] { curve, new BigInteger(q, isBigEndian: true) };
-#else
-                return new[] { curve, new BigInteger(q.Reverse()) };
-#endif
+                return new[] { curve, q.ToBigInteger() };
```
Rob-Hague: same

src/Renci.SshNet/Common/SshDataStream.cs
mus65: right, fixed by reverting to ifdef. I guess you could alternatively introduce an `allowInPlaceReverse` parameter to `ToBigInteger()` but that feels a little weird...

Rob-Hague (approved):

PR #1615, opened 2025-03-14
CI: run .NET Framework Integration Tests on Windows

mus65:
This runs the .NET Framework tests on Windows by setting up WSL2 in the Windows VM with [Vampire/setup-wsl](https://github.com/Vampire/setup-wsl) and starting the SSH server container with podman in the Ubuntu VM. This should catch more Windows specific issues than the current mono solution does.

To my surprise, this actually works. It's pretty slow though. A full test run takes about an hour. Also the already flaky test from #1253 has failed on every run so far. I assume the bad performance is somehow due to the nested virtualization. 

We could either disable the flaky test and live with the long CI time or only test specific tests here, like before with mono. Or not do this at all since it still feels a little experimental...

Why not with docker? Docker Desktop is not supported on Windows Server. Setting up docker in the ubuntu VM manually is a lot more complex than podman since it must be run as a daemon. I tried this as well, but couldn't get it to work. It can't connect to the Ubuntu VM for some reason.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -108,10 +108,12 @@ jobs:
           test/Renci.SshNet.Tests/
 
     - name: Setup WSL2
-      uses: vedantmgoyal9/setup-wsl2@main
+      uses: Vampire/setup-wsl@v5
+      with:
+        distribution: Ubuntu-24.04
 
     - name: Setup SSH Server
-      shell: wsl-run {0}
+      shell: wsl-bash {0}
       run: |
         apt-get update && apt-get upgrade -y
         apt-get install -y podman
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -107,6 +107,20 @@ jobs:
           -p:CoverletOutput=../../coverlet/windows_unit_test_net_4_6_2_coverage.xml `
           test/Renci.SshNet.Tests/
 
+  Windows-Integration-Tests:
+    name: Windows Integration Tests
+    runs-on: windows-2025
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+      with:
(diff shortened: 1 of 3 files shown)


Rob-Hague:
Awesome. I also tried to setup docker with WSL2 but decided I didn't know enough about it and gave up

.github/workflows/build.yml
```diff
@@ -132,6 +107,29 @@ jobs:
           -p:CoverletOutput=../../coverlet/windows_unit_test_net_4_6_2_coverage.xml `
           test/Renci.SshNet.Tests/
 
+    - name: Setup WSL2
+      uses: vedantmgoyal9/setup-wsl2@main
```
Rob-Hague: Does it also work with https://github.com/Vampire/setup-wsl? It seems to be the more popular choice

mus65: Missed this one. Also seems to work, switched to it.

Rob-Hague:
Could it be separated into a third leg, such that the publish step does not depend on its success? And then we can see how it goes with reliability

Some of the slowness might be because a lot of the tests make several connections between test initialise and cleanup. For example, the tests in `CipherTests` each make 6 connections. But it doesn't seem to be that costly locally so perhaps not worth optimising

Rob-Hague:
Have you normally only seen that particular test fail or others as well? If it's just that one, then I think we can disable it in CI (would like to keep it locally - I'm pretty sure it's a bug)

mus65:
I have only seen this particular test fail. I skip it now by checking for the CI env variable. There may be a better skip the test. NUnit has an `[Explicit]` attribute for this, but seems like MSTest doesn't have anything similar.

Rob-Hague (approved):

PR #1619, opened 2025-03-22
Use native AesGcm for .NET Framework from nuget

scott-xu:
Based on https://github.com/dotnet/runtime/issues/89718, we can use native `AesGcm` for .NET Framework from [Microsoft.Bcl.Cryptography](https://www.nuget.org/packages/Microsoft.Bcl.Cryptography/) nuget package

### DIFF
--- a/src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.cs
+++ b/src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.cs
@@ -65,15 +65,7 @@ namespace Renci.SshNet.Security.Cryptography.Ciphers
 #if !NETSTANDARD
             if (System.Security.Cryptography.AesGcm.IsSupported)
             {
-                try
-                {
-                    _impl = new BclImpl(key, _iv);
-                }
-                catch (DllNotFoundException)
-                {
-                    // Mono doesn't have BCrypt.dll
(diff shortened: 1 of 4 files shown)


Rob-Hague:
LGTM once the package is stable

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.cs
```diff
@@ -62,10 +62,18 @@ public AesGcmCipher(byte[] key, byte[] iv, int aadLength)
...
-#if NET
+#if !NETSTANDARD
             if (System.Security.Cryptography.AesGcm.IsSupported)
             {
-                _impl = new BclImpl(key, _iv);
+                try
+                {
+                    _impl = new BclImpl(key, _iv);
+                }
+                catch (DllNotFoundException)
```
Rob-Hague: have you hit this? I don't think we need to be that worried about it anymore

Rob-Hague (approved):

src/Renci.SshNet/Security/Cryptography/Ciphers/AesGcmCipher.cs
scott-xu: The CI test was ran at Ubuntu and hit this issue before.
I can unwrap "try catch" if you want.

PR #1624, opened 2025-03-29
Only enable TreatWarningsAsErrors in Release

mus65:
Having this enabled in Debug is imho annoying for multiple reasons:

- sometimes you just want to temporarily add code for debugging and don't want to deal with the errors.
- when a .NET SDK major version introduces new analyzers, you can't build anymore. I e.g. currently can't build just because I have .NET 10 Preview installed.
- In other projects I haven seen even minor SDK versions introduce new warnings because there were bugfixes in existing analyzers. So you suddenly can't build because of a .NET SDK minor update.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -13,7 +13,7 @@
     <WarningLevel>9999</WarningLevel>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(Configuration)' == 'Release'">
+  <PropertyGroup Condition="'$(Configuration)' == 'Release' Or '$(CI)' != ''">
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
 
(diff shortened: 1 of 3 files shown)


Directory.Build.props
```diff
@@ -11,6 +11,9 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <LangVersion>latest</LangVersion>
     <WarningLevel>9999</WarningLevel>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)' == 'Release'">
```
Rob-Hague: Release or CI?

Rob-Hague (commented):
agreed

Rob-Hague:
on the topic of SDKs, I *think* our global.json isn't doing anything right now. We could delete it or add `"allowPrerelease": false`

Directory.Build.props
mus65: I guess it doesn't matter much. I would prefer Release since you can easily use it locally to double-check that you didn't miss any warnings.

Rob-Hague: sorry for being lazy, I meant 

```suggestion
  <PropertyGroup Condition="'$(Configuration)' == 'Release' Or '$(CI)' != ''">
```

Rob-Hague (approved):

PR #1627, opened 2025-04-05
Add a debug helper for reading traffic with Wireshark

Rob-Hague:
Wireshark can already helpfully dissect the initial SSH handshake. When given the session keys, it can also dissect the encrypted traffic for inspection/debugging. This adds a helper in Debug mode to write out that information in the format Wireshark requires.

Usage is to set `SshNetLoggingConfiguration.WiresharkKeyLogFilePath` before connecting, and supply the same value to Wireshark in Edit -> Preferences -> Protocols -> SSH -> "Key log filename".

Description of the format is at https://wiki.wireshark.org/SSH#key-log-format and https://gitlab.com/wireshark/wireshark/-/blob/82d6f8631aad39f1f10db77a62ebae01e3be881f/epan/dissectors/packet-ssh.c#L2071

I've had this branch around for a while and had the impression that it didn't always work, but seems OK testing recently

image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,3 +41,11 @@ The tests always log to the console. See the [Logging documentation](https://ssh
 ### Wireshark
 
 Wireshark is able to dissect initial connection packets, such as key exchange, before encryption happens. Enter "ssh" as the display filter. See https://wiki.wireshark.org/SSH.md for more information.
+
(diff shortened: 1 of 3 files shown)


drieseng (commented):
I only see you logging the "shared secret" to the Wireshark log file. This does not correspond with the logging that you included in the description of this PR.

**Update**:
The logging in the description is solely produced by Wireshark, right?

Rob-Hague:
Yeah the shared secret is all that's needed, everything else comes from wireshark

WojciechNagorski (approved):

PR #1628, opened 2025-04-07
Added ExistsAsync and GetAsync to ISftpClient

Maarty:
Those methods were missing in the interface. Added so they can be easily mocked in tests.

### DIFF
--- a/src/Renci.SshNet/ISftpClient.cs
+++ b/src/Renci.SshNet/ISftpClient.cs
@@ -631,6 +631,22 @@ namespace Renci.SshNet
         bool Exists(string path);
 
         /// <summary>
+        /// Checks whether file or directory exists.
+        /// </summary>
+        /// <param name="path">The path.</param>
+        /// <param name="cancellationToken">The <see cref="CancellationToken"/> to observe.</param>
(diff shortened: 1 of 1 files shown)


drieseng:
@Rob-Hague, shouldn't we be consistent with the presence/absence of a default value for the **cancellationToken** argument? Personally, I prefer not to have a default value.

Maarty:
@drieseng, I agree, it should be united, I can edit it in all the methods in this PR. Currently, it is half/half - presence/absence.  

However, isn't cancellationToken with a default value more standard way? It is very common to have a default value in the .NET APIs, also they somehow recommend it in their blogs - [Recommended patterns for CancellationToken](https://devblogs.microsoft.com/premier-developer/recommended-patterns-for-cancellationtoken/#:~:text=It's%20a%20good%20idea%20to,you%20call%20(%232%20above).).

Rob-Hague (approved):
My preference is to minimise breaking changes i.e. keep it as it is. Adding to an interface is already a breaking change but it seems that most  people are just using it for mocking and similar has been done in previous releases, so OK

Thanks for the PR

PR #1634, opened 2025-04-24
Add UploadFileAsync and DownloadFileAsync methods

sdt-ndelarosa:
Adds TAP methods for `UploadFile` and `DownloadFile` to `ISftpClient`.

## Other Solutions

While there are other options to define both the `UploadFileAsync` and `DownloadFileAsync` methods, such as the [extension methods](https://github.com/sshnet/SSH.NET/pull/819#issuecomment-1791629586) approach and the previously proposed pull request, #1515 , this pull request has the following benefits over the other solutions:

* First class async/await support for uploading and downloading
  * Users no longer have to rely on implementing their own extension methods
* Cleaner interop with the sync versions of the `UploadFile` and `DownloadFile` methods
  * Method signatures have a closer match
* Upload and downloads now work with relative paths instead of just absolute paths
* Deeper async/await call chain, as compared to `BeginUploadFile`, which pretty much runs the sync UploadFile method, but [just in a different thread](https://github.com/sshnet/SSH.NET/blob/17afcdb0fe04c0a80f5c96018b6467af91746611/src/Renci.SshNet/SftpClient.cs#L1209)


src/Renci.SshNet/ISftpClient.cs
```diff
@@ -1104,6 +1123,37 @@ public interface ISftpClient : IBaseClient
...
+        /// <param name="cancellationToken">The <see cref="CancellationToken"/> to observe.</param>
+        /// <returns>A <see cref="Task"/> that represents the asynchronous upload operation.</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="input" /> is <see langword="null"/>.</exception>
+        /// <exception cref="ArgumentException"><paramref name="path" /> is <see langword="null"/> or contains only whitespace characters.</exception>
+        /// <exception cref="SshConnectionException">Client is not connected.</exception>
+        /// <exception cref="SftpPermissionDeniedException">Permission to upload the file was denied by the remote host. <para>-or-</para> A SSH command was denied by the server.</exception>
+        /// <exception cref="SshException">A SSH error where <see cref="Exception.Message" /> is the message from the remote host.</exception>
+        /// <exception cref="ObjectDisposedException">The method was called after the client was disposed.</exception>
+        Task UploadFileAsync(Stream input, string path, bool canOverride, CancellationToken cancellationToken = default);
```
sdt-ndelarosa: Decided to exclude the `Action<ulong> uploadCallback` because now the completion of the upload is tied to the completion of the returned `Task`

So if a user wanted to have a callback based on when the upload finishes awaiting, it would hypothetically have the following flow:

```csharp
await sftpClient.UploadFileAsync(input, path, false, cancellationToken);

// just do some logic here, since upload is done
```

sdt-ndelarosa:
actions don't seem to be triggering, gonna try closing and reopening

Rob-Hague:
Hi, I will make time to review most likely at some point next week. FYI for first time contributors the CI needs to be run manually (some kind of github security measure)

src/Renci.SshNet/ISftpClient.cs
Rob-Hague: Agreed on excluding the callback. I think we can also leave out this overload with `bool canOverride` - it is not hard to DIY and can always be added later if there is enough demand

src/Renci.SshNet/Sftp/SftpFileStream.cs
```diff
@@ -191,7 +191,7 @@ public TimeSpan Timeout
             }
         }
 
-        private SftpFileStream(ISftpSession session, string path, FileAccess access, int bufferSize, byte[] handle, long position)
```
Rob-Hague: use `SftpFileStream.OpenAsync` (line 331) instead of this constructor?

src/Renci.SshNet/SftpClient.cs
```diff
@@ -916,6 +916,30 @@ public void DownloadFile(string path, Stream output, Action<ulong>? downloadCall
...
+        /// <exception cref="ArgumentNullException"><paramref name="output" /> is <see langword="null"/>.</exception>
+        /// <exception cref="ArgumentException"><paramref name="path" /> is <see langword="null"/> or contains only whitespace characters.</exception>
+        /// <exception cref="SshConnectionException">Client is not connected.</exception>
+        /// <exception cref="SftpPermissionDeniedException">Permission to perform the operation was denied by the remote host. <para>-or-</para> A SSH command was denied by the server.</exception>
+        /// <exception cref="SftpPathNotFoundException"><paramref name="path"/> was not found on the remote host.</exception>///
+        /// <exception cref="SshException">A SSH error where <see cref="Exception.Message" /> is the message from the remote host.</exception>
+        /// <exception cref="ObjectDisposedException">The method was called after the client was disposed.</exception>
+        /// <remarks>
+        /// Method calls made by this method to <paramref name="output" />, may under certain conditions result in exceptions thrown by the stream.
+        /// </remarks>
```
Rob-Hague:
```suggestion
        /// <inheritdoc/>
```

src/Renci.SshNet/SftpClient.cs
```diff
@@ -1077,6 +1101,56 @@ public void UploadFile(Stream input, string path, bool canOverride, Action<ulong
...
+        /// <param name="input">Data input stream.</param>
+        /// <param name="path">Remote file path.</param>
+        /// <param name="cancellationToken">The <see cref="CancellationToken"/> to observe.</param>
+        /// <returns>A <see cref="Task"/> that represents the asynchronous upload operation.</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="input" /> is <see langword="null"/>.</exception>
+        /// <exception cref="ArgumentException"><paramref name="path" /> is <see langword="null"/> or contains only whitespace characters.</exception>
+        /// <exception cref="SshConnectionException">Client is not connected.</exception>
+        /// <exception cref="SftpPermissionDeniedException">Permission to upload the file was denied by the remote host. <para>-or-</para> A SSH command was denied by the server.</exception>
+        /// <exception cref="SshException">A SSH error where <see cref="Exception.Message" /> is the message from the remote host.</exception>
+        /// <exception cref="ObjectDisposedException">The method was called after the client was disposed.</exception>
```
Rob-Hague:
```suggestion
        /// <inheritdoc/>
```

test/Renci.SshNet.IntegrationTests/OldIntegrationTests/SftpClientTest.cs
```diff
@@ -31,6 +31,32 @@ protected static string CalculateMD5(string fileName)
             }
         }
 
+        protected static async Task<string> CalculateMD5Async(string fileName, CancellationToken cancellationToken)
```
Rob-Hague: I don't think we need an async version of this helper

test/Renci.SshNet.IntegrationTests/TestsFixtures/IntegrationTestBase.cs
```diff
@@ -81,5 +81,25 @@ protected void CreateTestFile(string fileName, int size)
...
             }
         }
+
+        /// <summary>
+        /// Creates the test file.
+        /// </summary>
+        /// <param name="fileName">Name of the file.</param>
+        /// <param name="size">Size in megabytes.</param>
+        /// <param name="cancellationToken">The <see cref="CancellationToken"/> to observe.</param>
+        protected async Task CreateTestFileAsync(string fileName, int size, CancellationToken cancellationToken)
```
Rob-Hague: ditto, not needed

Rob-Hague (commented):
Looks good. I think it could do with testing cancellation - i.e. that it throws `OperationCanceledException` when the token is cancelled.

It would probably not be very reliable to test cancelling during a file download (at least not in an integration test), but we could at least have a test for when a cancelled token is supplied

src/Renci.SshNet/ISftpClient.cs
Rob-Hague: by the way, the callback is not about completion but about progress of the upload. You can see that it gets invoked multiple times during the upload. Writing it from scratch one would probably use [`IProgress<T>`](https://learn.microsoft.com/dotnet/api/system.iprogress-1). Anyway I think it is fine to leave it out

sdt-ndelarosa: oooo good point on the `IProgress<T>`, I'll definitely leave that off to be potentially added later down the road.

src/Renci.SshNet/Sftp/SftpFileStream.cs
sdt-ndelarosa: i'm sticking with using the ctor directly instead of using the `OpenAsync` method mainly because calculating the optimal write length [needs the direct file handle](https://github.com/sdt-ndelarosa/SSH.NET/blob/d5b6333e0292ff581662e59da4623b9eab1f2fda/src/Renci.SshNet/SftpClient.cs#L2575), which is used for the copy buffer size

i might be misusing the intended use case for the `CalculateOptimalWriteLength` method, and if so i can just rely of the default buffer size of `81920` and swap out the call to `OpenAsync` if need be

src/Renci.SshNet/SftpClient.cs
```diff
@@ -2515,6 +2555,28 @@ private void InternalUploadFile(Stream input, string path, Flags flags, SftpUplo
...
+            cancellationToken.ThrowIfCancellationRequested();
+
+            var fullPath = await _sftpSession.GetCanonicalPathAsync(path, cancellationToken).ConfigureAwait(false);
+            var handle = await _sftpSession.RequestOpenAsync(fullPath, Flags.Write | flags, cancellationToken).ConfigureAwait(false);
+
+            using (var output = new SftpFileStream(_sftpSession, fullPath, FileAccess.Write, (int)_bufferSize, handle, 0L))
+            {
+                var bufferSize = (int)_sftpSession.CalculateOptimalWriteLength(_bufferSize, handle);
+                await input.CopyToAsync(output, bufferSize, cancellationToken).ConfigureAwait(false);
+            }
```
Rob-Hague: (moving the conversation here)

> i might be misusing the intended use case for the CalculateOptimalWriteLength method, and if so i can just rely of the default buffer size of 81920 and swap out the call to OpenAsync if need be

I think I would prefer that. I have doubts that `CalculateOptimalWriteLength` lives up to its name, so if we can reuse existing code and simplify this bit then that sounds good.

It does raise an interesting point that here you are canonicalizing the path while `SftpFileStream` currently does not. I've just spent far too long looking at it but I think it would be fine to add to `SftpFileStream`. I think it should make this (failing) test pass:

<details>

```diff
diff --git a/test/Renci.SshNet.IntegrationTests/SftpClientTests.cs b/test/Renci.SshNet.IntegrationTests/SftpClientTests.cs
index 67b3b28c..e90cf704 100644
--- a/test/Renci.SshNet.IntegrationTests/SftpClientTests.cs
+++ b/test/Renci.SshNet.IntegrationTests/SftpClientTests.cs
@@ -162,6 +162,46 @@ public async Task Create_directory_and_delete_it_using_DeleteAsync()
             Assert.IsFalse(await _sftpClient.ExistsAsync(testDirectory).ConfigureAwait(false));
         }
 
+        [TestMethod]
+        public async Task Open_PathCanonicalized()
+        {
+            var testDirectory = "/home/sshnet/sshnet-test";
+
+            // Create new directory and check if it exists
+            await _sftpClient.CreateDirectoryAsync(testDirectory).ConfigureAwait(false);
+
+            await _sftpClient.ChangeDirectoryAsync(testDirectory).ConfigureAwait(false);
+
+            Assert.AreEqual(testDirectory, _sftpClient.WorkingDirectory);
+
+            Assert.IsFalse(await _sftpClient.ExistsAsync("/home/sshnet/tempasync.txt").ConfigureAwait(false));
+            Assert.IsFalse(await _sftpClient.ExistsAsync("/home/sshnet/tempsync.txt").ConfigureAwait(false));
+
+            using (var stream = await _sftpClient.OpenAsync("../tempasync.txt", FileMode.Create, FileAccess.ReadWrite, default).ConfigureAwait(false))
+            {
+                await stream.WriteAsync(Encoding.ASCII.GetBytes("test"), 0, 4).ConfigureAwait(false);
+            }
+
+            using (var stream = _sftpClient.Open("../tempsync.txt", FileMode.Create, FileAccess.ReadWrite))
+            {
+                await stream.WriteAsync(Encoding.ASCII.GetBytes("test"), 0, 4).ConfigureAwait(false);
+            }
+
+            Assert.IsTrue(await _sftpClient.ExistsAsync("/home/sshnet/tempasync.txt").ConfigureAwait(false));
+            Assert.IsTrue(await _sftpClient.ExistsAsync("/home/sshnet/tempsync.txt").ConfigureAwait(false));
+
+            await _sftpClient.DeleteAsync("../tempasync.txt", CancellationToken.None).ConfigureAwait(false);
+            await _sftpClient.DeleteAsync("../tempsync.txt", CancellationToken.None).ConfigureAwait(false);
+
+            await _sftpClient.ChangeDirectoryAsync("../").ConfigureAwait(false);
+
+            await _sftpClient.DeleteAsync("sshnet-test", CancellationToken.None).ConfigureAwait(false);
+
+            Assert.IsFalse(await _sftpClient.ExistsAsync(testDirectory).ConfigureAwait(false));
+            Assert.IsFalse(await _sftpClient.ExistsAsync("/home/sshnet/tempasync.txt").ConfigureAwait(false));
+            Assert.IsFalse(await _sftpClient.ExistsAsync("/home/sshnet/tempsync.txt").ConfigureAwait(false));
+        }
+
         [TestMethod]
         public async Task Create_file_and_delete_using_DeleteAsync()
         {

```

</details>

Rob-Hague: canonicalizing the path in `SftpFileStream` will probably cause failures in some painfully strict unit tests, in which case it can be left for later

sdt-ndelarosa: i'm going to choose *not* to move the canonicalizing call into `SftpFileStream.OpenAsync`, mainly because the sync version of the `SftpFileStream` [also doesn't canonicalize](https://github.com/sshnet/SSH.NET/blob/7c07b10e4a9d8ba09be3ebe91ed8a7833dbc38ce/src/Renci.SshNet/Sftp/SftpFileStream.cs#L313).

And while i could add canonicalizing for both the sync and async flows to maintain consistency between the two, i think that it increases the scope of impact for this pr.  Canonicalizing definitely would be worth, but i'd say it would make sense to dedicate a separate pr to that

Rob-Hague: yeah don't worry about it, the tests for SftpFileStream are difficult to work with. I have an unfinished branch which rewrites a lot of them

Rob-Hague (approved):
Thanks!

PR #1637, opened 2025-05-03
Override more methods in PipeStream, ShellStream

Rob-Hague:
Where beneficial, add additional overrides from the base Stream class. Namely the Span variants and for PipeStream, the WriteAsync variants (see comments).

The change also adds an internal type borrowed from the runtime repo for easier buffer management, which could also be used elsewhere.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -9,6 +9,7 @@
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)Renci.SshNet.snk</AssemblyOriginatorKeyFile>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <EnablePackageValidation>true</EnablePackageValidation>
     <LangVersion>latest</LangVersion>
     <WarningLevel>9999</WarningLevel>
   </PropertyGroup>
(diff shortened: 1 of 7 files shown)


Directory.Build.props
```diff
@@ -9,6 +9,7 @@
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)Renci.SshNet.snk</AssemblyOriginatorKeyFile>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <EnablePackageValidation>true</EnablePackageValidation>
```
Rob-Hague, line 12: unrelated change, this just adds validation during `dotnet pack` that e.g. a netstandard build is compatible with netcore

https://learn.microsoft.com/dotnet/fundamentals/apicompat/package-validation/overview

src/Renci.SshNet/Abstractions/StreamExtensions.cs
```diff
@@ -8,8 +9,15 @@ internal static class StreamExtensions
...
-            return default;
+            try
+            {
+                stream.Dispose();
+                return default;
+            }
+            catch (Exception exc)
+            {
+                return new ValueTask(Task.FromException(exc));
+            }
```
Rob-Hague, line 20: unrelated, just copying the netcore impl

https://github.com/dotnet/runtime/blob/1d1bf92fcf43aa6981804dc53c5174445069c9e4/src/libraries/System.Private.CoreLib/src/System/IO/Stream.cs#L180C17-L187C10

WojciechNagorski (approved):

PR #1638, opened 2025-05-03
Read the underlying buffer in SshDataStream

Rob-Hague:
SshDataStream is a MemoryStream, so we can access the buffer directly. Also simplify some usage in PrivateKeyFile.


src/Renci.SshNet/Common/SshDataStream.cs
```diff
@@ -153,12 +143,21 @@ public void Write(string s, Encoding encoding)
...
         public byte[] ReadBinary()
+        {
+            return ReadBinarySegment().ToArray();
+        }
+
+        /// <summary>
+        /// Reads a length-prefixed byte array from the SSH data stream,
+        /// returned as a view over the underlying buffer.
+        /// </summary>
+        internal ArraySegment<byte> ReadBinarySegment()
```
Rob-Hague, line 160: later this will be used in `SshData` to remove large array allocations when reading e.g. `ChannelDataMessage` (the motivation for this change)

WojciechNagorski (approved):

PR #1643, opened 2025-05-11
Fix hang in SftpClient.UploadFile upon error

### LINKED ISSUE #957: Deadlock in InternalUploadFile
I went through SftpClient briefly and I noticed that the reset event is only set when s.StatusCode == StatusCodes.Ok

```
_sftpSession.RequestWrite(handle, offset, buffer, 0, bytesRead, null, s =>
                        {
                            if (s.StatusCode == StatusCodes.Ok)
                            {
                                Interlocked.Decrement(ref expectedResponses);
                                responseReceivedWaitHandle.Set();

                                //  Call callback to report number of bytes written
                                if (uploadCallback != null)
                                {
                                    //  Execute callback on different thread
                                    ThreadAbstraction.ExecuteThread(() => uploadCallback(writtenBytes));
                                }
                            }
                        });
```
 which means that 
once loop reaches  
```
//  Wait for expectedResponses to change 
_sftpSession.WaitOnHandle(responseReceivedWaitHandle, _operationTimeout);
```
and status code for all write requests != StatusCodes.Ok the loop will hang

This is not just observation I've seen it happen many times, this results in **thread starvation for the host application eventually**.

image

image

### LINKED ISSUE #1640: SshException check yields a wrong result during normal run, but not while stepping in debugger
SSH.NET 2025.0.0

`SshException` ("Failure") is about to happen as a result of `SendRequest(request);` call. But by the time `if (exception is not null)` check is done, the `exception` variable is still `null` (because it's set asynchronously after we get the response). So it won't be thrown by `throw exception;`. That's the case if we just run after breaking on `SendRequest(request);` line. But if we step over to the `if (exception is not null)` line, the `exception` variable gets a chance to be set, and we get that `throw exception;` executed as it was intended.

This bug makes me unable to catch `SshException` in my code while using `sftpClient.UploadFile()`.

You can see demos in the GIFs below:

Image

Image

Rob-Hague:
When the server returns an error during UploadFile, unless the response is quick enough that the exception is captured before SftpSession.RequestWrite finishes...

https://github.com/sshnet/SSH.NET/blob/d08c4aaa8a6b5cacf5c5acd0964b8ddd2f5eb2dd/src/Renci.SshNet/Sftp/SftpSession.cs#L941-L944

... then the error is not propagated because SftpClient.UploadFile does not have any handling for failure (StatusCode != Ok):

https://github.com/sshnet/SSH.NET/blob/d08c4aaa8a6b5cacf5c5acd0964b8ddd2f5eb2dd/src/Renci.SshNet/SftpClient.cs#L2475-L2480

This causes deadlock because the expectedResponses is only decremented in this path and will never hit zero.

The first commit is the fix for the deadlock (handle failure codes in UploadFile). The second commit tidies up SftpSession.RequestWrite to fix the racy exception behaviour as pointed out in #1640 (the exception handling only makes sense here when given a wait handle).

~I wrote an ugly test for this (https://github.com/Rob-Hague/SSH.NET/commit/9a7526abe6477aeb1efaeb2e3fd08716d9a50ddd) but it only fails on `develop` after cherry-picking the second commit because the test is single-threaded so we encounter the race in RequestWrite.~

closes #957
closes #1640

### DIFF
--- a/src/Renci.SshNet/Sftp/Responses/SftpStatusResponse.cs
+++ b/src/Renci.SshNet/Sftp/Responses/SftpStatusResponse.cs
@@ -12,7 +12,7 @@
         {
         }
 
-        public StatusCodes StatusCode { get; private set; }
+        public StatusCodes StatusCode { get; set; }
 
         public string ErrorMessage { get; private set; }
 
@@ -38,6 +38,13 @@
                 // expressed as US-ASCII
                 Language = ReadString(Ascii);
             }
+        }
+
+        protected override void SaveData()
+        {
+            base.SaveData();
+
+            Write((uint)StatusCode);
         }
     }
 }
--- a/src/Renci.SshNet/Sftp/SftpSession.cs
+++ b/src/Renci.SshNet/Sftp/SftpSession.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Globalization;
 using System.Text;
 using System.Threading;
@@ -914,6 +915,8 @@ namespace Renci.SshNet.Sftp
                                  AutoResetEvent wait,
                                  Action<SftpStatusResponse> writeCompleted = null)
         {
+            Debug.Assert((wait is null) != (writeCompleted is null), "Should have one parameter or the other.");
+
             SshException exception = null;
 
             var request = new SftpWriteRequest(ProtocolVersion,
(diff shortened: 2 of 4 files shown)


drieseng (approved):
Can you reproduce this using an integration test?

Rob-Hague:
Thanks. Probably not easily, but I'm adding the regression test anyway as it uncovered a race condition in the impl

PR #1646, opened 2025-05-22
Re-introduce ChangeWindow method on ShellStream

### LINKED ISSUE #40: Terminal window resizing

### LINKED ISSUE #1061: Not able to send Window Change Request from ShellStream
After opening a **_SshClient_** Connection (calling  CreateShell) there is no way to call  **_SendWindowChangeRequest_** because _channel (_IChannelSession_) is an internal field and not accessible.

The solution appears to be surfacing the SendWindowChangeRequest in the **_ShellStream_** class. I note other requests like ChangeFlow etc. would remain inaccessible though.

The alternative is to surface the IChannelSession interface and make it a read only property to get references to all the underlying methods.  But that is 2 levels of abstraction more primitive than the client - so implementing a method "WindowChangeRequest" thats calls the IChannelSession.SendWindowChangeRequest method is most obvious, but that would suggest implementing a wrapper to all the other methods: SendPseudoTerminalRequest, SendX11ForwardingRequest - ie practically the whole IChannel Interface.  

So I recommend surfacing the IChannelSession as a read only property.

### LINKED ISSUE #1252: How can I modify the cols rows in CreateShellStream according to the size of the front-end terminal?
How can I modify the cols rows in CreateShellStream according to the size of the front-end terminal?

Rasmus715:
Due to stale state of https://github.com/sshnet/SSH.NET/pull/1062, I've decided to carry the feature by myself.

All the necessary suggestions from the initial PR were applied.

closes #1061
closes #1252
closes #1062 
closes #1514
closes #40

Let's make this PR to the main branch. I desperately need this feature

### DIFF
--- a/src/Renci.SshNet/Channels/ChannelSession.cs
+++ b/src/Renci.SshNet/Channels/ChannelSession.cs
@@ -239,26 +239,18 @@ namespace Renci.SshNet.Channels
         /// <param name="rows">The rows.</param>
         /// <param name="width">The width.</param>
         /// <param name="height">The height.</param>
-        /// <returns>
-        /// <see langword="true"/> if request was successful; otherwise <see langword="false"/>.
-        /// </returns>
-        public bool SendWindowChangeRequest(uint columns, uint rows, uint width, uint height)
+        public void SendWindowChangeRequest(uint columns, uint rows, uint width, uint height)
         {
             SendMessage(new ChannelRequestMessage(RemoteChannelNumber, new WindowChangeRequestInfo(columns, rows, width, height)));
-            return true;
         }
 
         /// <summary>
         /// Sends the local flow request.
         /// </summary>
         /// <param name="clientCanDo">if set to <see langword="true"/> [client can do].</param>
-        /// <returns>
-        /// <see langword="true"/> if request was successful; otherwise <see langword="false"/>.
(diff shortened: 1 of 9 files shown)


Rasmus715:
Pipeline fails due not to my changes. I'll fix it as well.

Rob-Hague (approved):
Thanks. I have added contribution credit to @glenkleidon and @294797392 for their previous efforts.

Rasmus715:
Thank you for a quick response! 
Is there a way to put NuGet with this change as a prerelease one?

Rob-Hague:
You can use the GH feed: https://github.com/sshnet/SSH.NET#using-pre-release-nuget-packages

There probably won't be one on nuget.org for a while

PR #1647, opened 2025-05-25
Drop netstandard2.1 target

mus65:
This is imho not worth targeting. The only non-EOL frameworks that actually benefit from this are Mono 6.4 and Unity 2021.2 (see https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-1).

Even if people use SSH.NET with any of these, they will still work fine with the netstandard2.0 target and only miss very minor optimizations.

Simplified `#if NETFRAMEWORK || NETSTANDARD2_0` to `#if !NET` in the process.


Rob-Hague (approved):
Seems ok to me

PR #1658, opened 2025-06-18
scp: add flag to notify for uploading finished when uploaded an empty file

thegame4craft:
Added a new flag: `notifyOnEmptyFile` (default: `false`).

Details
    * This flag controls whether the `Uploading` event should be triggered when an uploaded file is empty (`totalLength` == 0).
    * When enabled and the uploaded file is empty, the `Uploading` event will still be raised.
    * This behavior is currently implemented only for **SCP** uploads.

Use Case
I’m building a simple .NET Framework application that uploads configuration files to an SSH server. Each upload triggers a log entry via the Uploading event. However, empty files were previously ignored, and no log was written. With this change, the log will accurately reflect all upload attempts — including empty files — when the flag is enabled.

### DIFF
--- a/src/Renci.SshNet/ScpClient.cs
+++ b/src/Renci.SshNet/ScpClient.cs
@@ -240,12 +240,13 @@ namespace Renci.SshNet
         /// </summary>
         /// <param name="source">The <see cref="Stream"/> to upload.</param>
         /// <param name="path">A relative or absolute path for the remote file.</param>
+        /// <param name="notifyOnEmptyFile">Should the <see cref="Uploading"/> event be raised when the file is empty?</param>
         /// <exception cref="ArgumentNullException"><paramref name="path" /> is <see langword="null"/>.</exception>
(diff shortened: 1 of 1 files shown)


src/Renci.SshNet/ScpClient.cs
```diff
@@ -579,6 +583,11 @@ private void UploadFileContent(IChannelSession channel, Stream input, Stream sou
                 read = source.Read(buffer, 0, buffer.Length);
             }
 
+            if (totalLength == 0 && totalRead == 0 && notifyOnEmptyFile)
+            {
+                RaiseUploadingEvent(remoteFileName, totalLength, totalRead);
+            }
```
Rob-Hague: Rather than changing the API, I would be happy just to take these 3 lines

```suggestion
            if (totalLength == 0 && totalRead == 0)
            {
                RaiseUploadingEvent(remoteFileName, totalLength, totalRead);
            }
```

thegame4craft: Updated!

Rob-Hague (approved):
Thanks 👍

PR #1659, opened 2025-06-22
Use BCL MLKem

scott-xu:
.NET 10 adds ML-Kem to BCL. For lower targets, it adds support from [Microsoft.Bcl.Cryptography](https://www.nuget.org/packages/Microsoft.Bcl.Cryptography/) nuget package.

See https://github.com/dotnet/runtime/issues/113508


scott-xu:
@Rob-Hague Any suggestions how to test BCL MLKem? Ubuntu 25.10 has OpenSSL 3.5 but it is not available from GitHub Actions https://github.com/actions/runner-images?tab=readme-ov-file#available-images

scott-xu:
> @Rob-Hague Any suggestions how to test BCL MLKem? Ubuntu 25.10 has OpenSSL 3.5 but it is not available from GitHub Actions https://github.com/actions/runner-images?tab=readme-ov-file#available-images

ML-KEM and ML-DSA are added to Windows Server 2025 and Windows 11 clients (24H2, 25H2) since build 26100.7171 and 26200.7171. See https://techcommunity.microsoft.com/blog/microsoft-security-blog/post-quantum-cryptography-apis-now-generally-available-on-microsoft-platforms/4469093

PR #1668, opened 2025-07-11
Fixes spelling errors

jacobslusser:
_See #1269 for some relevant discussion_

### List of changes
- Adds `exclusion.dic` file for spellchecker words to ignore
- Updates `.editorconfig(s)` to configure VS spellchecker for the library and unit test projects
- Fixes numerous misspelled words in comments, private members, and public members of test projects
> NOTE: There is a misspelling of the `ErrorOccured` event in the `Session` class intentionally left unchanged because it is public-facing. Where we subscribe to that event via `Session_ErrorOccurred`, those instances are all private/internal and hence have been corrected.

### Testing
All existing unit tests continue to pass.

### Impact Analysis
None. **No functional changes or public-facing APIs have been changed.**


CONTRIBUTING.md
```diff
@@ -18,7 +18,7 @@ The library has a test project for unit tests and a test project for integration
 
 Code coverage information can be generated for all test projects at once or for individual test projects. From the root of the repository or from the individual test project directory, run `dotnet test --collect:"XPlat Code Coverage"`.
 
-The coverage information can be visualised using e.g. [ReportGenerator](https://reportgenerator.io/). Install the ReportGenerator dotnet tool with `dotnet tool update -g dotnet-reportgenerator-globaltool` and then run
```
Rob-Hague: This is correct in British English. Is it easy to allow British/real English spelling?

jacobslusser: I've updated the `.editorconfig` to allow both `en-US` and `en-GB`. In my testing that stopped my IDE from flagging these words as misspelled. BTW, I also had to install the UK language pack in Windows; for anyone else who is planning on reviewing spelling, make sure you have done the same.

I've reverted this and similar words back to the King's English. 

Finally, the VS spellchecker leaves a lot to be desired. I don't want to give the impression that this PR is any way enabling it as part of our tooling, it's already part of the IDE. I just used it for the purposes of fixing these spelling errors and configured it for anyone else go forward.

Rob-Hague: Ok, thanks for the info, I didn't consider that it would need a language pack

exclusion.dic
Rob-Hague, line 1: this is showing as a binary file, does it need an entry in .gitattributes?

Rob-Hague: the file was encoded as UTF-16 LE BOM which git didn't like

src/Renci.SshNet/.editorconfig
```diff
@@ -1,4 +1,7 @@
 ﻿[*.cs]
+spelling_languages = en-us,en-gb
+spelling_exclusion_path = ../exclusion.dic
```
Rob-Hague, line 3: the file this is pointing to has only 2 words, both already included in the top-level file. Is it an artifact of earlier changes?

```suggestion
spelling_exclusion_path = ../../exclusion.dic
```

src/exclusion.dic
Rob-Hague, line 1: this is the (redundant?) file I am talking about

Rob-Hague (approved):

exclusion.dic
jacobslusser: When I was struggling to get `en-GB` to work, one of the troubleshooting suggestions said that the `exclusion.dic` file needed to be UTF-16 LE BOM encoded, hence why I made the change. In the end, I think what resolved my issue was installing the necessary language packs in Windows so I think it is safe to revert this back to UTF8 as you did.

src/exclusion.dic
jacobslusser: Another behavior (bug) in Visual Studio spellchecker. There were some times when it would honor the `exclusion.dic` file in the root and other times when it would only honor the one in the immediate same folder as the project/`.editorconfig`. From your fix, its possible I had my path wrong. Either way, I got tired of troubleshooting MS/VS issues and in the end tried to focus on the point of the PR which was to fix a few spelling errors, not necessarily build a spellchecking tool pipeline.

Rob-Hague: sure, thanks

PR #1672, opened 2025-07-16
Add .NET 10 target and make use of C#14 extension members

mus65:
This adds the new target and fixes/suppresses new analyzer warnings.

Also make use of the [C# 14 extension members](https://devblogs.microsoft.com/dotnet/csharp-exploring-extension-members/) to get rid of most `ThrowHelper` methods and some `#if`.

There may be more places to make use of it (or maybe I already overdid it :smile: ).


src/Renci.SshNet/Common/Lock.cs
```diff
@@ -5,14 +5,16 @@ namespace Renci.SshNet.Common
 {
     internal sealed class Lock
     {
+        private readonly object _lockObject = new object();
```
Rob-Hague: This doesn't look right - `lock(instance)` and `instance.TryEnter` should work with the same mutex. Suggest reverting this file

mus65: I supressed the warning.

mus65:
@Rob-Hague I created a separate PR for extension members #1699. I figured it's easier to review this separately. Or should I just add it here?

Rob-Hague:
Don't mind, if it's in separate commits then happy for it to be in 1 pr if it makes branch management easier

mus65:
Merged the extension member changes into this PR.

src/Renci.SshNet/Abstractions/SHA1Extensions.cs
Rob-Hague, line 1: is it a requirement that they are all in separate files? I mean, could these simple extensions just stay in CryptoAbstraction?

src/Renci.SshNet/Channels/ChannelDirectTcpip.cs
```diff
@@ -124,10 +124,12 @@ private void CloseSocket()
 
             lock (_socketLock)
             {
+#pragma warning disable CA1508 // Avoid dead conditional code
```
Rob-Hague: I think we can disable this one - we already have it suppressed elsewhere as well

src/Renci.SshNet/Abstractions/SHA1Extensions.cs
mus65: They must be in separate classes, because otherwise you would have multiple `CryptoAbstraction.HashData` methods with the same signature. The `extension` level doesn't matter in this case (which I can understand from a compiler and backward compatibility standpoint).

I guess you could put all classes in one file, but that's currently prevented by the "one class per file" analyzer.

Rob-Hague: Ok, thanks

src/Renci.SshNet/Channels/ChannelDirectTcpip.cs
mus65: done

src/Renci.SshNet/Abstractions/RandomNumberGeneratorExtensions.cs
```diff
@@ -0,0 +1,22 @@
+﻿#nullable enable
+namespace System.Security.Cryptography
+{
+    internal static class RandomNumberGeneratorExtensions
+    {
+#if !NET
+        private static readonly RandomNumberGenerator Randomizer = RandomNumberGenerator.Create();
```
Rob-Hague: make the one on CryptoAbstraction internal?

src/Renci.SshNet/Sftp/SftpFileStream.cs
```diff
@@ -164,7 +164,9 @@ internal static SftpFileStream Open(
             int bufferSize,
             bool isDownloadFile = false)
         {
+#pragma warning disable CA2000 // Dispose objects before losing scope
```
Rob-Hague: I think we can kill this analyzer, it seems pretty noisy

src/Renci.SshNet/SftpClient.cs
```diff
@@ -2331,13 +2335,15 @@ private async Task InternalDownloadFile(
                         break;
                     }
 
+#pragma warning disable CA1849 // Call async methods when in an async method
```
Rob-Hague: Can you move these up to line 2277 (or kill one of the analyzers since they seem like duplicates)

src/Renci.SshNet/SftpClient.cs
```diff
@@ -2434,13 +2444,15 @@ private async Task InternalUploadFile(
 
             while (true)
             {
+#pragma warning disable CA1849 // Call async methods when in an async method
```
Rob-Hague: Similar here (funnily enough, it does not fire later on in the method)

Rob-Hague (commented):
lgtm otherwise

mus65:
I just realized that 6c3c06d95a2695bf2fa582156a01431843bee46b is caused by analyzers which are still in preview. Imho we should not be dealing with these, so i disabled preview analyzers with 9f1933354b27dc3a84e79b1309e8432a5f3561e5.

Rob-Hague (approved):

Rob-Hague:
The fewer analyzers the better imo

Issue #1669, opened 2025-07-13
Cannot associate log messages with specific instance of SftpClient


shadytester:
Assume I have a few instances of SftpClient, a few open connections and want to get log messages from each of them separately. How can I do this?
I can register custom implementation of ILoggerFactory but it will receive log messages from all existing instances and these messages do not have any info that would allow to distinglish their senders.

It looks like a design bug.

mus65:
You usually only have a single ILoggerFactory in an application. This type of context-specific information is usually added with a [Logging Scope](https://learn.microsoft.com/en-us/dotnet/core/extensions/logging?tabs=command-line#log-scopes). So you could e.g. add an instance specific Guid to the logging scope while using the instance. Then this Guid is automatically added to all log messages.

If you e.g. use the client as part of an ASP .NET Core Http Request, you may not even have to do this because ASP .NET Core will automatically create a scope which already includes e.g. the Request Path and a Request Id.

I suppose SSH.NET could add some kind of instance id to the scope itself, but I have not seen other libraries do this, probably because you usually already have other Scope information from e.g. an ASP .NET Core Http Request or a scope specific to a hosted service.

shadytester:
I need to use a few instances of SftpClient in my application and without ability to distinglish log messages the overal logging system becomes useless. Any workaround will work. Maybe it's possible to add a reference to related client object or any kind of its artificial "id" to the "state" parameter (it's a collection of key-value pairs)...

desdesdes:
1+. I would really love to see this feature implemented. We use different settings to communicate with various providers, and if there are any issues, we would appreciate being able to trace a specific SftpClient session without interfering with the global logging.

You should be able to implement this in a backward-compatible manner by allowing the LoggerFactory to be set on individual objects (for example, on the ConnectionInfo), while falling back to SshNetLoggingConfiguration.LoggerFactory if not set.

This approach would also enhance our ability to conduct unit tests, as singletons are not very unit test-friendly.

If I would create a pull request for this you you consider merging it?

mus65:
> You should be able to implement this in a backward-compatible manner by allowing the LoggerFactory to be set on individual objects (for example, on the ConnectionInfo), while falling back to SshNetLoggingConfiguration.LoggerFactory if not set.

[...]

PR #1673, opened 2025-07-18
Refactor logging to allow a loggerfactory per session

desdesdes:
Refactor logging to allow a loggerfactory per session specified in the ConnectionInfo.

This commit introduces an `ILoggerFactory` to various classes, replacing the static logger factory with an instance-based approach for more flexible and session-specific logging. These changes improve the logging framework's flexibility and maintainability and allow unit testing of logging.

### DIFF
--- a/src/Renci.SshNet/BaseClient.cs
+++ b/src/Renci.SshNet/BaseClient.cs
@@ -20,6 +20,8 @@ namespace Renci.SshNet
         /// Holds value indicating whether the connection info is owned by this client.
         /// </summary>
         private readonly bool _ownsConnectionInfo;
+
+        private readonly ILogger _logger;
         private readonly IServiceFactory _serviceFactory;
         private readonly object _keepAliveLock = new object();
         private TimeSpan _keepAliveInterval;
@@ -34,14 +36,6 @@ namespace Renci.SshNet
         /// The current session.
         /// </value>
         internal ISession? Session { get; private set; }
-
-        internal ILogger? Logger
-        {
-            get
-            {
-                return Session?.SessionLoggerFactory?.CreateLogger(GetType());
-            }
-        }
 
         /// <summary>
         /// Gets the factory for creating new services.
@@ -198,6 +192,7 @@ namespace Renci.SshNet
             _connectionInfo = connectionInfo;
             _ownsConnectionInfo = ownsConnectionInfo;
             _serviceFactory = serviceFactory;
+            _logger = (connectionInfo.LoggerFactory ?? SshNetLoggingConfiguration.LoggerFactory).CreateLogger(GetType());
             _keepAliveInterval = Timeout.InfiniteTimeSpan;
         }
 
@@ -351,7 +346,7 @@ namespace Renci.SshNet
         /// <exception cref="ObjectDisposedException">The method was called after the client was disposed.</exception>
         public void Disconnect()
         {
-            Logger?.LogInformation("Disconnecting client.");
+            _logger.LogInformation("Disconnecting client.");
 
             CheckDisposed();
 
@@ -450,7 +445,7 @@ namespace Renci.SshNet
 
             if (disposing)
             {
-                Logger?.LogDebug("Disposing client.");
+                _logger.LogDebug("Disposing client.");
 
                 Disconnect();
 
@@ -513,7 +508,7 @@ namespace Renci.SshNet
                 }
                 catch (Exception ex)
                 {
-                    Logger?.LogError(ex, "Error sending keepalive message");
+                    _logger.LogError(ex, "Error sending keepalive message");
                 }
                 finally
                 {
--- a/src/Renci.SshNet/BaseClient.cs
+++ b/src/Renci.SshNet/BaseClient.cs
@@ -20,8 +20,6 @@ namespace Renci.SshNet
         /// Holds value indicating whether the connection info is owned by this client.
         /// </summary>
         private readonly bool _ownsConnectionInfo;
-
-        private readonly ILogger _logger;
         private readonly IServiceFactory _serviceFactory;
         private readonly object _keepAliveLock = new object();
         private TimeSpan _keepAliveInterval;
@@ -36,6 +34,14 @@ namespace Renci.SshNet
         /// The current session.
         /// </value>
         internal ISession? Session { get; private set; }
+
+        internal ILogger? Logger
+        {
+            get
+            {
+                return Session?.SessionLoggerFactory?.CreateLogger(GetType());
+            }
+        }
 
         /// <summary>
         /// Gets the factory for creating new services.
@@ -192,7 +198,6 @@ namespace Renci.SshNet
             _connectionInfo = connectionInfo;
             _ownsConnectionInfo = ownsConnectionInfo;
             _serviceFactory = serviceFactory;
-            _logger = SshNetLoggingConfiguration.LoggerFactory.CreateLogger(GetType());
             _keepAliveInterval = Timeout.InfiniteTimeSpan;
         }
 
@@ -346,7 +351,7 @@ namespace Renci.SshNet
         /// <exception cref="ObjectDisposedException">The method was called after the client was disposed.</exception>
         public void Disconnect()
         {
-            _logger.LogInformation("Disconnecting client.");
+            Logger?.LogInformation("Disconnecting client.");
 
             CheckDisposed();
 
@@ -445,7 +450,7 @@ namespace Renci.SshNet
 
             if (disposing)
             {
-                _logger.LogDebug("Disposing client.");
+                Logger?.LogDebug("Disposing client.");
 
                 Disconnect();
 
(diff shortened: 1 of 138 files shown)


src/Renci.SshNet/BaseClient.cs
```diff
@@ -192,7 +198,6 @@ private protected BaseClient(ConnectionInfo connectionInfo, bool ownsConnectionI
             _connectionInfo = connectionInfo;
             _ownsConnectionInfo = ownsConnectionInfo;
             _serviceFactory = serviceFactory;
-            _logger = SshNetLoggingConfiguration.LoggerFactory.CreateLogger(GetType());
```
Rob-Hague, line 195: The new Logger property is going to instantiate a new logger object on each access. Could it instead keep this line, instantiating the field from `connectionInfo.LoggerFactory`, with a `??` fallback to the static factory like you have done in Session?

src/Renci.SshNet/ForwardedPortLocal.cs
```diff
@@ -103,7 +103,7 @@ public ForwardedPortLocal(string boundHost, uint boundPort, string host, uint po
             Host = host;
             Port = port;
             _status = ForwardedPortStatus.Stopped;
-            _logger = SshNetLoggingConfiguration.LoggerFactory.CreateLogger<ForwardedPortLocal>();
+            _logger = Session.SessionLoggerFactory.CreateLogger<ForwardedPortLocal>();
```
Rob-Hague: I have kicked off a test run. I think it is going to show NullReferenceException from this line (lots of files are still not null-annotated)

src/Renci.SshNet/Security/KeyExchange.cs
```diff
@@ -69,7 +69,7 @@ public byte[] ExchangeHash
         /// </summary>
         protected KeyExchange()
         {
-            _logger = SshNetLoggingConfiguration.LoggerFactory.CreateLogger(GetType());
+            _logger = Session.SessionLoggerFactory.CreateLogger(GetType());
```
Rob-Hague: here also will probably be NRE

Rob-Hague:
ah and unfortunately a lot of the mocking tests are "strict" in that they expect every interface call to be declared. I think you can do whatever is easiest to get them working i.e. remove the strictness if it helps

src/Renci.SshNet/BaseClient.cs
```diff
@@ -37,6 +35,14 @@ public abstract class BaseClient : IBaseClient
...
         internal ISession? Session { get; private set; }
 
+        internal ILogger? Logger
+        {
+            get
+            {
+                return Session?.SessionLoggerFactory?.CreateLogger(GetType());
+            }
+        }
+
```
mus65: This could be made non-nullable, so the `Logger?.` could be avoided as well.

```suggestion
        internal ILogger Logger
        {
            get
            {
                return Session?.SessionLoggerFactory?.CreateLogger(GetType()) ?? NullLogger.Instance;
            }
        }

```

src/Renci.SshNet/BaseClient.cs
desdesdes: Thanks. Updated as suggested.

src/Renci.SshNet/ForwardedPortLocal.cs
desdesdes: This is hard to solve without a ?. For now i changed the code to create the logger only when logging is needed. Because the `Session` property could change during the lifetime of the `ForwardedPort` I chose not to cache the logger, but get an instance only when needed. Looking at the code this should only occur in exceptional cases.

src/Renci.SshNet/Security/KeyExchange.cs
desdesdes: Thanks. Updated as suggested.

src/Renci.SshNet/BaseClient.cs
desdesdes: Thanks. Solved by implementing the change after @Rob-Hague comment.

desdesdes:
> ah and unfortunately a lot of the mocking tests are "strict" in that they expect every interface call to be declared. I think you can do whatever is easiest to get them working i.e. remove the strictness if it helps

Removing the strictness and working around the null return would require a lot of ?? code in the library. So for now i chose the setup the logger everywhere. This create a lot of additions to the unit test, but I think this is the cleanest solution for now.

src/Renci.SshNet/Connection/ConnectorBase.cs
```diff
@@ -15,13 +15,15 @@ namespace Renci.SshNet.Connection
     internal abstract class ConnectorBase : IConnector
     {
         private readonly ILogger _logger;
+        private readonly ILoggerFactory _loggerFactory;
```
Rob-Hague: nit: field not needed

```suggestion
```

src/Renci.SshNet/SubsystemSession.cs
```diff
@@ -72,6 +72,14 @@ public bool IsOpen
             get { return _channel is not null && _channel.IsOpen; }
         }
 
+        public ILoggerFactory SessionLoggerFactory
+        {
+            get
+            {
+                return _session.SessionLoggerFactory;
```
Rob-Hague, line 79: _session could potentially be null here. I think it is fine to make _session readonly and remove the `_session = null;` line in Dispose

src/Renci.SshNet/Session.cs
```diff
@@ -352,6 +352,14 @@ public string ClientVersion
         /// </value>
         public ConnectionInfo ConnectionInfo { get; private set; }
 
+        /// <summary>
+        /// Gets the logger factory for this session.
+        /// </summary>
+        /// <value>
+        /// The logger factory for this session.
+        /// </value>
+        public ILoggerFactory SessionLoggerFactory { get; private set; }
```
Rob-Hague:
```suggestion
        public ILoggerFactory SessionLoggerFactory { get; }
```

src/Renci.SshNet/ConnectionInfo.cs
```diff
@@ -495,5 +497,13 @@ IList<IAuthenticationMethod> IConnectionInfoInternal.AuthenticationMethods
             get { return AuthenticationMethods.Cast<IAuthenticationMethod>().ToList(); }
 #pragma warning restore S2365 // Properties should not make collection or array copies
         }
+
+        /// <summary>
+        /// Gets or sets logger factory for this connection.
+        /// </summary>
+        /// <value>
+        /// The logger factory for this connection. If <see langword="null"/> then <see cref="SshNetLoggingConfiguration.LoggerFactory"/> should be used.
```
Rob-Hague:
```suggestion
        /// The logger factory for this connection. If <see langword="null"/> then <see cref="SshNetLoggingConfiguration.LoggerFactory"/> is used.
```

Rob-Hague (commented):
LGTM otherwise, although I have not tried it out to see if it addresses #1669 (I trust you have done that)

@mus65 any other comments?

mus65:
> LGTM otherwise, although I have not tried it out to see if it addresses #1669 (I trust you have done that)
> 
> @mus65 any other comments?

I will take a closer look on the weekend.

src/Renci.SshNet/Connection/ConnectorBase.cs
desdesdes: Thanks. Fixed according to suggestion.

src/Renci.SshNet/SubsystemSession.cs
desdesdes: Thanks. Fixed according to suggestion.

src/Renci.SshNet/Session.cs
desdesdes: Thanks. Fixed according to suggestion.

src/Renci.SshNet/ConnectionInfo.cs
desdesdes: Thanks. Fixed according to suggestion.

mus65:
LGTM. I wrote [a test case for this](https://github.com/mus65/SSH.NET/commit/27903c8690d036848d2c0cf6971567bd4171305f) in a separate branch, but I'm not sure it's worth adding since it requires yet another dependency (test-only though). I would leave this up to @Rob-Hague .

This imho makes SshNetLoggingConfiguration kind of obselete, but marking it obsolete after just introducing it doesn't seem sensible. 😄

mus65 (approved):

Rob-Hague (approved):

Rob-Hague:
> LGTM. I wrote [a test case for this](https://github.com/mus65/SSH.NET/commit/27903c8690d036848d2c0cf6971567bd4171305f) in a separate branch, but I'm not sure it's worth adding since it requires yet another dependency (test-only though). I would leave this up to @Rob-Hague .

That's a nice test to validate this change. I imagine it is less useful going forward (although perhaps moreso than some other tests...)

Issue #1215, opened 2023-10-16
No permissions to create file in Amazon S3 SFTP server using S3 file system?


kikaragyozov:
I get the following exception when using `SftpClient.Create(string)`:

```
Renci.SshNet.Common.SshException: 'Cannot open file in mode: CREATE|EXCLUSIVE|READ|WRITE'
```

I've given full access for the transfer protocol as well as the s3 file system using these 2 policies:
AmazonS3FullAccess

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:*",
                "s3-object-lambda:*"
            ],
            "Resource": "*"
        }
    ]
}
```

AWSTransferFullAccess
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": "transfer:*",
            "Resource": "*"
        },
        {
            "Effect": "Allow",
            "Action": "iam:PassRole",
            "Resource": "*",
            "Condition": {
                "StringEquals": {
                    "iam:PassedToService": "transfer.amazonaws.com"
                }
            }
        },
        {
            "Effect": "Allow",
            "Action": [
                "ec2:DescribeVpcEndpoints",
                "ec2:DescribeNetworkInterfaces",
                "ec2:DescribeAddresses"
            ],
            "Resource": "*"
        }
    ]
}
```

I can successfully create & delete a folder, but the moment I Try to do so with a file, I get the above error.

I can successfully create& delete a file using a SFTP explorer like `WinSCP`. What's going on here?

Using `SftpClient.UploadFile` or the async extension `UploadAsync` works without hiccups.

kikaragyozov:
I think after investigating, it's possible Amazon's servers do not support the `EXCLUSIVE` flag, which is weird.

mitchej7:
@kikaragyozov Did you manage to work around this issue?

kikaragyozov:
> @kikaragyozov Did you manage to work around this issue?

Hello. Yes, I'm simply not using the method that's opening the file with an `EXCLUSIVE` flag, i.e `SftpClient.Create(string)`. As I initially stated, I've got no problems using `UploadAsync` or `UploadFile` to get the job done.

mitchej7:
Having done more research on this, I don't think it's the EXCLUSIVE flag. I believe that S3 doesn't like having the READ and WRITE flag set at the same time.

Rob-Hague:
[...]

PR #1680, opened 2025-08-01
Remove unnecessary SftpFileStream unit tests and dedup Open{Async}

Rob-Hague:
SftpFileStream has decent coverage in the integration tests (namely https://github.com/sshnet/SSH.NET/blob/d40bc43ac16d0ba54aae965d4923608dfb5fa475/test/Renci.SshNet.IntegrationTests/SftpTests.cs). It means the unit tests on SftpFileStream are largely redundant, and because they are quite verbose and strongly tied to internal behaviour, serve only to hinder new development. This change deletes a lot of the unit tests; adds a few not explicitly covered by the integration tests; bolsters the integration tests a bit; and finally, deduplicates SftpFileStream.Open/OpenAsync.


src/Renci.SshNet/Sftp/SftpFileStream.cs
```diff
@@ -284,16 +280,7 @@ internal SftpFileStream(ISftpSession session, string path, FileMode mode, FileAc
...
                 case FileMode.Create:
-                    _handle = _session.RequestOpen(path, flags | Flags.Truncate, nullOnError: true);
-                    if (_handle is null)
-                    {
-                        flags |= Flags.CreateNew;
-                    }
-                    else
-                    {
-                        flags |= Flags.Truncate;
-                    }
```
Rob-Hague, line 295: It is at best ineffectual and at worst a bug that we send `SSH_FXF_TRUNC` without `SSH_FXF_CREAT` (what we have as `Flags.CreateNewOrOpen`). Per the draft SFTPv3 spec (https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-02#section-6.3), `SSH_FXF_TRUNC` must be accompanied by `SSH_FXF_CREAT`. OpenAsync had the logic correct, so we use that one. I am wondering whether it has something to do with #1215

src/Renci.SshNet/Sftp/SftpFileStream.cs
```diff
@@ -310,127 +297,44 @@ internal SftpFileStream(ISftpSession session, string path, FileMode mode, FileAc
...
-                catch
-                {
-                    try
-                    {
-                        await session.RequestCloseAsync(handle, cancellationToken).ConfigureAwait(false);
-                    }
-                    catch
-                    {
-                        // The original exception is presumably more informative, so we just ignore this one.
-                    }
```
Rob-Hague, line 427: the only other library change aside from my comment above is that we use the Open (not OpenAsync) version of the FileMode.Append handling, which doesn't try RequestClose on FStat failure. Seems sensible to call it, but given the sync method is probably used much more and there are seemingly no known issues (with this part), I forwent the RequestClose

WojciechNagorski (approved):

PR #1686, opened 2025-08-09
Truncate existing file in CreateText

### LINKED ISSUE #275: SftpClient: Modify methods that overwrite existing file to first truncate the file
The following methods in **SftpClient** currently overwrite an existing file, but do not truncate it first:

* public StreamWriter CreateText(string path)
* public StreamWriter CreateText(string path, Encoding encoding)
* public void WriteAllBytes(string path, byte[] bytes)
* public void WriteAllLines(string path, IEnumerable<string> contents)
* public void WriteAllLines(string path, string[] contents)
* public void WriteAllLines(string path, IEnumerable<string> contents, Encoding encoding)
* public void WriteAllLines(string path, string[] contents, Encoding encoding)
* public void WriteAllText(string path, string contents)
* public void WriteAllText(string path, string contents, Encoding encoding)

The API has been "inspired" by [System.IO.File](https://msdn.microsoft.com/en-us/library/system.io.file.aspx), but the corresponding methods on [System.IO.File](https://msdn.microsoft.com/en-us/library/system.io.file.aspx) actually overwrite (and truncate) an existing file.

Rob-Hague:
SftpClient.CreateText does not truncate the file if it exists, contrary to System.IO.File.CreateText which does. It is documented, but seems pretty unintuitive and more like a mistake than a concious decision (the documentation was added much later than the implementation).

I considered leaving this 14 year old behaviour as it is, but turns out other people have hit it as well (it also affects WriteAll{Bytes/Text/Lines}).

closes #275


drieseng (approved):

Rob-Hague:
Thanks

PR #1688, opened 2025-08-12
Fix SftpFileAttributes file type detection

### LINKED ISSUE #159: Get/Update Unix special permissions (setuid, setgid and Sticky Bit)
Hi,

Is there a way to get/update the unix special permissions (setuid, setgid and Sticky Bit)? 

If it's not possible, can you add this feature in the next version?

thx

### LINKED ISSUE #1000: SftpFileAttributes incorrect bit masking for file type, should always use `& S_IFMT`
Hello, it looks like the code in `SftpFileAttributes.cs` that handle the file type bits in the `permission` uint uses the wrong bitmask when comparing against the various file type flags.

For example, a symbolic link, while correctly returning true for `IsSymbolicLink`, also incorrectly returns true for `IsCharacterDevice`.

In particular, I think [these lines](https://github.com/sshnet/SSH.NET/blob/develop/src/Renci.SshNet/Sftp/SftpFileAttributes.cs#L393-L405) in the private setter should be changed from:
```
                IsSocket = ((value & S_IFSOCK) == S_IFSOCK);
                IsSymbolicLink = ((value & S_IFLNK) == S_IFLNK);
                IsRegularFile = ((value & S_IFREG) == S_IFREG);
                IsBlockDevice = ((value & S_IFBLK) == S_IFBLK);
                IsDirectory = ((value & S_IFDIR) == S_IFDIR);
                IsCharacterDevice = ((value & S_IFCHR) == S_IFCHR);
                IsNamedPipe = ((value & S_IFIFO) == S_IFIFO);
```

to

```
                IsSocket = ((value & S_IFMT) == S_IFSOCK);
                IsSymbolicLink = ((value & S_IFMT) == S_IFLNK);
                IsRegularFile = ((value & S_IFMT) == S_IFREG);
                IsBlockDevice = ((value & S_IFMT) == S_IFBLK);
                IsDirectory = ((value & S_IFMT) == S_IFDIR);
                IsCharacterDevice = ((value & S_IFMT) == S_IFCHR);
                IsNamedPipe = ((value & S_IFMT) == S_IFIFO);
```

i.e. the value should always be bitmask-AND'ed with `S_IFMT` (aka `0xF000`) before comparing to the file type.

Otherwise you get a false-positive `true` result for things like symbolic links (`0xA000`) being compared to character device (`0x2000`) because `(0xA000 & 0x2000) == 0x2000` is `true` (but incorrect), but it probably should be `(0xA000 & 0xF000) == 0x2000` which is `false` (and correct).

Also, the usage of `private bool _isBitFiledsBitSet` seems unnecessary. Perhaps this private bool field should be removed entirely (including the two statements that get and set the value). I don't think it makes sense to set `permission = permission | S_IFMT` since `S_IFMT` is just a *bitmask* intended to be used for bitwise-AND operations (`&`) to find the other file type values.

Also, somewhat unrelated, I wish the three private bools `_isUIDBitSet`, `_isGroupIDBitSet` and `_isStickyBitSet` were exposed in the public API with at least getters, so that we can show the full `-rwsr-xr-x` unix permission for `/bin/su` for example. (With the current API, we can't know that `_isUIDBitSet` is true so we'd have to show it as just `-rwxr-xr-x` instead). In fact, having a public getter for the `uint permission` directly would be very useful as it would allow us to pass a single uint to for example a web frontend, instead of dozens of `IsXyz` + `XyzCanRead` booleans.

This affects both the `develop` branch as well as the `release/2020.0` branch (the file `SftpFileAttributes.cs` is identical in both branches)

Rob-Hague:
To get the file type, S_IFMT should be used as the mask. Instead it was using each file type as the mask. It meant that e.g. a symbolic link would also show as a regular file and a character device.

Also allow setting and retrieving the setuid/setgid/sticky bits

closes #1000
closes #159


src/Renci.SshNet/Sftp/SftpFileAttributes.cs
```diff
@@ -506,7 +708,7 @@ public byte[] GetBytes()
             {
                 uint flag = 0;
 
-                if (IsSizeChanged && IsRegularFile)
```
Rob-Hague, line 509: I have removed the `IsRegularFile` condition because I don't think it needs to exist. And given the bug, this condition would still have passed for a symbolic link, so this is more like keeping the behaviour the same

WojciechNagorski (approved):

PR #1692, opened 2025-08-30
Implement gssapi-with-mic user authentication with Kerberos

### LINKED ISSUE #114: SSH NTLM Single sign on
Please add as authentication provider "gssapi-with-mic" with NTLM. It allows single sign on on windows without typing password.
Some part of gssapi is already here https://github.com/SimonWilkinson/gss-openssh/
From server point of view eg Bitwise SSH already supports this.

### LINKED ISSUE #170: GSSAPI support
Putty supports authentication using GSSAPI. This is very useful if single sign on is required.

It would be really handy if SSH.NET could also support this capability.

### LINKED ISSUE #780: GSSAPI/Kerberos Support
Request to add GSSAPI/Kerberos support to the library to support password-less login to servers.

scott-xu:
This PR adds support for gssapi-with-mic user authentication with Kerberos which is defined at https://datatracker.ietf.org/doc/html/rfc4462.

Close https://github.com/sshnet/SSH.NET/issues/780
Close https://github.com/sshnet/SSH.NET/issues/170
Close https://github.com/sshnet/SSH.NET/issues/114


scott-xu:
This PR is primarily based on https://github.com/tmds/Tmds.Ssh/pull/188. Credits to @jborean93 and @tmds

🎏 Call for help!
This PR is still at early stage. If someone can help implement `ReflectedNegotiateContext` or add integration test, that would be great!

Rob-Hague:
Some high-level questions

1. I am not excited by the prospect of reflection/UnsafeAccessor. Is it just for lower targets? If so, I would prefer to restrict it to targets where it is possible via public api (NET 9+ presumably).

2. It seems like it would be easier to just have:
```c#
public class GssApiAuthenticationMethod(NegotiateAuthenticationClientOptions options)
{ }
```
as the public api?

3. Have you thought about how to test it? Presumably something can be set up via the dockerfile

scott-xu:
> 1. I am not excited by the prospect of reflection/UnsafeAccessor. Is it just for lower targets? If so, I would prefer to restrict it to targets where it is possible via public api (NET 9+ presumably).

Reflection is to support .NET Framework.
UnsafeAccessor is to support .NET 8.0

One reason to choose SSH.NET is that it supports lower targets.


> 2. It seems like it would be easier to just have:
> 
> ```cs
> public class GssApiAuthenticationMethod(NegotiateAuthenticationClientOptions options)
> { }
> ```
> 
> as the public api?

There are too many options than needed if we use `NegotiateAuthenticationClientOptions` as public API.

> 3. Have you thought about how to test it? Presumably something can be set up via the dockerfile

That's the "call out for help" item 😄

PR #1699, opened 2025-09-11
Make use of C#14 extension members

mus65:
This makes use of the [C# 14 extension members](https://devblogs.microsoft.com/dotnet/csharp-exploring-extension-members/) to get rid of most `ThrowHelper` methods and some `#if`.

There may be more places to make use of it (or maybe I already overdid it :smile: ).

Based on .NET 10 branch #1672


mus65:
Merged with PR #1672

PR #1704, opened 2025-09-28
CI: add Windows Integration Tests for .NET

mus65:
see https://github.com/sshnet/SSH.NET/pull/1702#issuecomment-3342506642


scott-xu:
How about adding some filters `--filter "Name~EcdhSha2Nist|Name~Gcm|Name~Ecdsa|Name~Tripledes"` to speed up the integration test?

mus65:
I would be OK with that. We could also run these in parallel to the .NET Framework tests.

There is one test failure though, no idea about that one.

```
 Test method Renci.SshNet.IntegrationTests.ConnectivityTests.Common_CreateMoreChannelsThanMaxSessions threw exception: 
Renci.SshNet.Common.SshException: The pseudo-terminal request was not accepted by the server. Consult the server log for more information.
  Stack Trace:
     at Renci.SshNet.ShellStream..ctor(ISession session, String terminalName, UInt32 columns, UInt32 rows, UInt32 width, UInt32 height, IDictionary`2 terminalModeValues, Int32 bufferSize) in /_/src/Renci.SshNet/ShellStream.cs:line 93
   at Renci.SshNet.ServiceFactory.CreateShellStream(ISession session, String terminalName, UInt32 columns, UInt32 rows, UInt32 width, UInt32 height, IDictionary`2 terminalModeValues, Int32 bufferSize) in /_/src/Renci.SshNet/ServiceFactory.cs:line 180
   at Renci.SshNet.SshClient.CreateShellStream(String terminalName, UInt32 columns, UInt32 rows, UInt32 width, UInt32 height, Int32 bufferSize, IDictionary`2 terminalModeValues) in /_/src/Renci.SshNet/SshClient.cs:line 287
   at Renci.SshNet.SshClient.CreateShellStream(String terminalName, UInt32 columns, UInt32 rows, UInt32 width, UInt32 height, Int32 bufferSize) in /_/src/Renci.SshNet/SshClient.cs:line 279
   at Renci.SshNet.IntegrationTests.ConnectivityTests.Common_CreateMoreChannelsThanMaxSessions() in D:\a\SSH.NET\SSH.NET\test\Renci.SshNet.IntegrationTests\ConnectivityTests.cs:line 47
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

mus65:
I can reproduce the test failure locally on linux when running the net9.0 tests multiple times against the same SSH server container. Recreating the SSH container fixes it. So this appears to be caused by some left over state in the SSH server from the previous run.

Shouldn't we run these in in separate jobs in CI anyway? Then they would use their own SSH containers and run in parallel which would speed things up significantly.

There is obviously some kind of issue in the tests though either way. I will try to narrow it down further...

Rob-Hague:
Yeah, they should be separate legs. Thanks for digging in

mus65:
Opened #1710 to fix the test failure.

mus65:
Merged #1710 here because this maybe also causes the Ssh_CreateShellNoTerminal test failure.

I also figured out why the Windows tests are so slow, it's DNS....  e2c0ed247f960a48c56aab72424fba7e6b557a54 . Should be done in <15min now.

.github/workflows/build.yml
Rob-Hague, line 197: add the windows integration legs here as well?

Rob-Hague (commented):
Nice!

Rob-Hague (approved):

Issue #1331, opened 2024-02-20
SftpClient: slow download


goremykin:
Hi! Thank you for this project, I appreciate your work.

I have a job in js that downloads some files over sftp every night using [ssh2-sftp-client](https://www.npmjs.com/package/ssh2-sftp-client).
I decided to rewrite this job in C# using SSH.NET, but it turned out that the same files are now downloaded up to 15 times slower.
For example, a 25 MB file can be downloaded in 5 seconds using ssh2-sftp-client and in over a minute using SSH.NET.

I tested on ubuntu and arch linux in a simple .net 8 project. SSH.NET version - 2023.0.1.
```
using var sftpClient = new SftpClient(host, port, userName, password);
await using (var localFileStream = new FileStream("/someFile", FileMode.Create, FileAccess.Write))
{
    sftpClient.DownloadFile(file.FullName, localFileStream);
}
```

In case of reading from a stream it even slower:
```
await using var remoteStream = await sftpClient.OpenAsync("/someFile", FileMode.Open, FileAccess.Read, CancellationToken.None);
await remoteStream.CopyToAsync(localFileStream);
```

Do you have any ideas on what I should check or what settings I should play with?

scott-xu:
~~Not sure if it relates to compression. Do you know if your SSH service supports and **only** supports compression?~~
Well, I don't think it relates to compression.

Waynezhi:
Same situation, using 2024.0.0, download speed is around 1MB/s only on a sftp server (protocol v3), and same environment with sftp command I got 15MB/s. @WojciechNagorski

goremykin:
Can confirm the same issue with 2024.0.0

Waynezhi:
I tried with 2024.0.0 and 2023.0.1, same result @goremykin, with .net 6
https://github.com/sshnet/SSH.NET/issues/733
https://github.com/sshnet/SSH.NET/issues/1122

seems this fix does not work for me: https://github.com/sshnet/SSH.NET/pull/866

Rob-Hague:
There has been no recent investigation into SFTP performance that I know of. If you make an investigation, please do share your findings. Thanks

TheGreenAirplane:
Hello, has there been any progress on this? I'm experiencing slow download with the SCP client, this could be related.

WojciechNagorski:
@TheGreenAirplane I would be happy to review your PR that will solve this problem.

megapro17:
[...]

PR #1705, opened 2025-09-28
Build the read-ahead mechanism into SftpFileStream

### LINKED ISSUE #1618: IOException instead of SftpPathNotFoundException with SFTPgo and S3 backend
I know this sounds like a server problem at first, but all other clients I tried so far work just fine.

Scenario: GET (in SSH.NET, `SftpClient.OpenRead`) a file that doesn't exist on the server.

Usual behavior with other servers: `OpenRead` throws a `SftpPathNotFoundException` exception.

Actual behavior: `OpenRead` throws no exception, but when trying to read from the stream, a generic `IOException` is thrown:

```
System.IO.IOException: Seek operation failed.
   at Renci.SshNet.Sftp.SftpFileStream.get_Length()
   at System.IO.Stream.GetCopyBufferSize()
   at System.IO.Stream.CopyTo(Stream destination)
```

I'm using SSH.NET 2024.2.0 with SFTPGo 2.6.6 here. The described behavior only occurs when I'm using a SFTPgo user with S3 storage. When connecting with a user configured for local storage, it works as expected.

Now I know that this sounds like a SFTPGo error at first, and the behavior for the two storage backends is indeed different, but it seems like it still conform to the standard. The server logs indicate that a `SSH_FXP_STATUS` with status code 2 is sent in both cases. Also, I tried to do the same in Python using the paramiko package, and it gives me `[Errno 2] no such file` in both cases. However, with local storage, I get this exception right when opening the file, and with S3 storage, only when trying to read from the stream.

I cannot replicate the behavior with the `sftp` command. Its verbose output indicates that it always sends a `SSH_FXP_STAT` command before trying to actually get the file, so I don't run into the same situation:

```
sftp> get foobar
debug1: stat remote: No such file or directory
File "/foobar" not found.
```

I also captured the traffic, trying to find out what is actually going on on the wire, but Wireshark cannot dissect SSH packages.

If you like to reproduce the problem, the easiest way should be to run a local Docker container (image: drakkan/sftpgo:alpine-slim), and set up the a user with S3 storage: 
https://docs.sftpgo.com/2.6/docker/#supported-tags-and-respective-dockerfile-links
https://docs.sftpgo.com/2.6/s3/

If you have any idea how I could capture the raw traffic in a way that is readable for you, please let me know. If necessary, I could also set up a temporary cloud-based instance for testing.

It would be perfectly fine if the exception is thrown only when trying to read from the stream, but it should be the `SftpPathNotFoundException` exception, not some generic exception that could also be caused by e.g. a network hiccup.

### LINKED ISSUE #1633: `SftpClient.Open*` methods can conflict with SFTP servers that have "one-time download" feature
# Context
We started getting an `IOException` after migrating our application from NET472 to NET9.0 with the following error:

> "Seek operation failed."
> at long Renci.SshNet.Sftp.SftpFileStream.get_Length()
> at int System.IO.Stream.GetCopyBufferSize()
> at void System.IO.Stream.CopyTo(Stream destination)

After investigating initially, we realized this was because of a difference in behavior between the implementations of `Stream.CopyAsync(Stream)` in the BCL: in NET472, it used a constant buffer size, whereas in NET Core and up, it defaults to the same buffer size, but tries to optimize it if the source `Stream` is seekable (`CanSeek == true`). 

The optimization can be seen here:
https://github.com/dotnet/runtime/blob/1d1bf92fcf43aa6981804dc53c5174445069c9e4/src/libraries/System.Private.CoreLib/src/System/IO/Stream.cs#L132-L153

Then, we noticed that this exception was only occuring on a specific vendor's SFTP server we interfaced with, but not in many others. At this point, I was under the impression that the server was just wrong: it was returning `CanSeek` as `true`, but did not allow us to get the file's `Length` property, which is an inconsistency in the way `Stream`'s contract is supposed to behave: if `CanSeek` is `true`, then `Position` and `Length` properties should be available.

However, by looking at the code on here I realized this might not be only the SFTP server's fault. My first suspicion was that `SftpFileStream`'s `CanSeek` property appears to always be initialized to `true`, without first checking the server at all. Then, the `Length` property tries to fire a STAT request which fails, violating `Stream`'s contract. My initial reaction here was that maybe SSH.NET could actually handle this more gracefully, by conditioning the `CanSeek` property instead of hardcoding it to `true`.

# The underlying problem
It appears that some SFTP servers, including this specific one where we noticed the exception, have this capability of "one-time download", or in other words some sort of expiration counter for files.

If this capability is set to 1, the server doesn't seem to allow further "interactions" with the file after the initial request is made (I assume in the same session here, although I don't know details on SFTP protocols).

While debugging locally, I tried manually grabbing attributes for the file after obtaining it via `client.ListDirectory`, by calling `client.GetAttributes(fileName)`. This call also fails, but with a more specific exception:

> The message [/folder1/folder2/file.json] is not extractable!

This new error led me to some evidence of other libraries having the same trouble:
- https://forum.rebex.net/6025/length-property-of-the-stream-throws-an-exception

And after reading that thread, I'm now convinced SSH.NET could protect against this in an elegant manner.

# Proposal
When initially listing files using `client.ListDirectory`, the returned `ISftpFile` abstraction _already contains the file's `Length` value, whch is retrieved from the `Attributes` on that file.

We take this information and return to our consuming class, which then decides it needs to open the file for copying/transfering/etc. To to that, it calls one of the many `client.Open` overloads, passing the full file name obtained from the `ISftpFile` from earlier. This is then what finally creates the `Stream` instance (the `SftpFileStream`) that we use for reading the file's contents.

The `client.Open(string path)` method is designed to open an arbitrary file from the server, and thus takes in the name property. But because of this, it can't really tap into the _already obtained information about the file_, such as it's `Length`, stored in the `ISftpFile` interface.

My proposal then is the following:

_Create a new overload of `client.Open` that instead of taking a raw `string` as its argument, takes in the original `ISftpFile` object obtained by methods such as `ListDirectory`._

Example:
```cs
public SftpFileStream OpenRead(ISftpFile file)
```

When _this_ overload is called, the implementation could then propagate the `Length` attribute from the `ISftpFile` to the `SftpFileStream` instance returned, and this would completely avoid the need to fire a separate `RequestFStat` call inside the `Length` property:

https://github.com/sshnet/SSH.NET/blob/17afcdb0fe04c0a80f5c96018b6467af91746611/src/Renci.SshNet/Sftp/SftpFileStream.cs#L114

In turn, this would avoid the entire situation that led us into the `IOException`, as the `Length` would become accessible, and the copy would occur normally without any workarounds.

I initially thought the library could not do anything about this, but this simple addition to me looks like a nice way to go about it. It optimizes for the "list -> download" use case nicely by reusing file attributes obtained in the "list" call, in the "download" phase.

It would also not disrupt any other existing flows, since the current path-based overloads of `Open` would remain intact.

# Workaround
To workaround the issue temporarily, we decided to short-circuit the whole thing by preventing the `Length`-based optimization to even be attempted when using `Stream.CopyTo` in modern .NET: instead of using the `CopyTo(Stream)` overload and letting it try to compute a better buffer size, we instead used the explicit `CopyTo(Stream, int)` version and passed in the standard buffer size ourselves. When using this overload, the framework just honors the provided size and doesn't try to access the `Stream`'s `Length` property anymore, avoiding the entire issue.

This, of course, is far from ideal. It is a brittle solution, considering how pervasive `Stream.CopyTo` is, as it could just as well happen somewhere else in our code base and cause the same hard-to-debug situation again.

A cleaner, more elaborate workaround would be to create our own custom `Stream` wrapper, that would wrap this library's `SftpFileStream` and take the `ISftpFile` as a means to grab the `Length` value and bypass the normal `SftpFileStream`'s `Length` implementation.

We've not yet decided on doing that, but we might have to in case the authors here don't make any changes, since that would isolate the issue much more cleanly and not create a burden on the rest of our code.

### LINKED ISSUE #1644: SftpFileStream OpenAsync/CopyToAsync vs DownloadFile performance gap
I didn't find issue with a useful answer to this comparison.

I'm trying to download files from SFTP Server (the platform i'm building can access multiple SFTP in a generic way).

I'm facing an issue where downloading a 10Mo file using SftpFileStream.OpenAsync with SftpFileStream.CopyToAsync is a nightmare.

To download this kind of file, i have to wait at least 45sec-60sec in average (same behavior using the synchronous CopyTo).

Image

Seems to have an issue in the Read/ReadAsync of the SftpFileStream here.

After looking for a workaround, i found a potential solution was using the synchronous method to download a file.

Image

And, unfortunately, this solution was indeed better (10-15sec for this 10Mo file) which is still low, but i can live with that, i'm not that familar with the protocol, maybe it's part of it.

In my case, where i have a lot of SFTP traffic, using a synchronous download is a bit annoying to be honest.

I can afford to wait 10-15sec (synchronous call), but i would like to use the asynchronous call.

When i'm looking at the implementations of both, it seems that "DownloadFile" tries to download chunk concurrently (with a Semaphore limit to 100) which the ReadAsync clearly don't do. So the async read is sequential and read stream sequentially according to the bufferSize (65536 even if i try to increase it).
Is my understanding good ? If so, why this behavior change between sync and async read ?

Is the issue known ? Is that even an issue ?

PS : I'm using the latest version : 2025.0.0

PS2 : It seems that this behavior is happening on some SFTP Server, not everyones. Let me know what you need on the server specs if it can be useful (i only have what is offers by the discover made by the connection)

Rob-Hague:
This change unifies the SFTP download implementations that exist via DownloadFile and
via SftpFileStream, by rewriting SftpFileStream to perform the same "read-aheads" as
DownloadFile. This brings the performance of downloads via SftpFileStream in line with
DownloadFile, such that the latter is now effectively SftpFileStream.CopyTo. It also
brings the recently added DownloadFileAsync up to speed since that was implemented via
SftpFileStream.CopyToAsync.

The methodology is a mix of the previous one and that within OpenSSH: the first call to
SftpFileStream.Read sends one read request to the server. The second sends two and when
not interrupted by Write or similar, the number of in-flight read requests continues to
scale up in this fashion.

I have measured CopyTo to be 3-20x faster than before, depending on file size and server
round-trip time. See https://github.com/sshnet/SSH.NET/pull/1705#issuecomment-3419987228

closes #1644
contributes to #1331
fixes #1633 
fixes #1618


Rob-Hague:
Performance analysis:

Summary: ~~probable 10% regression in DownloadFile for small files on high-latency connections~~ (fixed https://github.com/sshnet/SSH.NET/pull/1705#issuecomment-3425185970). Otherwise, no statistically significant differences for DownloadFile.

Tests were run for different file sizes against a geographically close server (Europe) and one far away (US West).

<details>
<summary>1MB, Europe</summary>
The noisiest test. No statistically significant difference for DownloadFile.

1MB_Europe

</details>

<details>
<summary>1MB, West US</summary>
The DownloadFile difference is somewhat consistent 

1MB_WestUS

</details>

<details>
<summary>8MB, Europe</summary>

8MB_Europe

</details>

<details>
<summary>8MB, West US</summary>

8MB_WestUS

</details>

<details>
<summary>64MB, Europe</summary>

64MB_Europe

</details>

<details>
<summary>64MB, West US</summary>

64MB_WestUS

</details>

WojciechNagorski (approved):

Rob-Hague:
> 1MB, West US
> The DownloadFile difference is somewhat consistent

Fixed in e99f5a50c6f2b3451fe

<details>

1MB_WestUS

</details>

julealgon:
Awesome change @Rob-Hague . Any ETA for when it will be released?

Rob-Hague:
Thanks, thinking of putting out a prerelease in the next few days and a proper release in a couple of weeks (if all goes well...)

PR #1706, opened 2025-09-29
Remove calls to Socket.Poll and use SocketShutdown.Both

Rob-Hague:
The message loop currently sits in a call to Poll until the socket has data to read or it is closed. This is unnecessary - it can equally just sit in the call to Receive.

The call to Poll in Session.IsConnected is also unnecessary - we can instead just call Socket.Connected. This only returns the connection state as of the last operation, but we are always performing operations in the message loop (or else we are not connected), so it should work equally well while being cheaper.

Lastly, when shutting down the socket, shut down both sides rather than just the sending side (SocketShutdown.Both rather than SocketShutdown.Send) - at this point we do not care about reading anything else. This makes it (more) certain that we will break out of the Receive call in the message loop, as has been noted in #355 for whatever remaining issues still exist there.

### DIFF
--- a/src/Renci.SshNet/Session.cs
+++ b/src/Renci.SshNet/Session.cs
@@ -82,12 +82,6 @@ namespace Renci.SshNet
         private readonly ILogger _logger;
 
         /// <summary>
-        /// Holds an object that is used to ensure only a single thread can read from
-        /// <see cref="_socket"/> at any given time.
-        /// </summary>
-        private readonly Lock _socketReadLock = new Lock();
-
-        /// <summary>
         /// Holds an object that is used to ensure only a single thread can write to
         /// <see cref="_socket"/> at any given time.
         /// </summary>
@@ -105,7 +99,7 @@ namespace Renci.SshNet
         /// This is also used to ensure that <see cref="_socket"/> will not be disposed
         /// while performing a given operation or set of operations on <see cref="_socket"/>.
         /// </remarks>
-        private readonly SemaphoreSlim _socketDisposeLock = new SemaphoreSlim(1, 1);
+        private readonly Lock _socketDisposeLock = new Lock();
 
         /// <summary>
         /// Holds an object that is used to ensure only a single thread can connect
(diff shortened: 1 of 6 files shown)


src/Renci.SshNet/Session.cs
```diff
@@ -81,12 +81,6 @@ public sealed class Session : ISession
         private readonly ISocketFactory _socketFactory;
         private readonly ILogger _logger;
 
-        /// <summary>
-        /// Holds an object that is used to ensure only a single thread can read from
-        /// <see cref="_socket"/> at any given time.
-        /// </summary>
-        private readonly Lock _socketReadLock = new Lock();
```
Rob-Hague, line 88: no longer needed

src/Renci.SshNet/Session.cs
```diff
@@ -1988,46 +1886,37 @@ private static int TrySocketRead(Socket socket, byte[] buffer, int offset, int l
         /// </summary>
         private void SocketDisconnectAndDispose()
         {
-            if (_socket != null)
```
Rob-Hague, line 1991: not a hot path, so simplify the locking and save some indentation

src/Renci.SshNet/Session.cs
```diff
@@ -2102,25 +1972,12 @@ private void MessageListener()
...
-                // a connection exception which is raised while isDisconnecting is normal and
-                // should be ignored
-                if (connectionException != null)
-                {
-                    return;
-                }
-
-                // any timeout while disconnecting can be caused by loss of connectivity
-                // altogether and should be ignored
-                if (exp is SocketException socketException && socketException.SocketErrorCode == SocketError.TimedOut)
```
Rob-Hague, line 2120: dead code, we always wrap SocketException in SshConnection

test/Renci.SshNet.Tests/Classes/SessionTest_Connected_ConnectionReset.cs
```diff
@@ -64,8 +64,6 @@ public void ErrorOccurredIsRaisedOnce()
 
             var connectionException = (SshConnectionException)exception;
             Assert.AreEqual(DisconnectReason.ConnectionLost, connectionException.DisconnectReason);
-            Assert.IsNull(connectionException.InnerException);
-            Assert.AreEqual("An established connection was aborted by the server.", connectionException.Message);
```
Rob-Hague, line 68: not always the case - it could contain a SocketException upon RST, in which case the message is more like "An existing connection was forcibly closed by the remote host."

src/Renci.SshNet/Session.cs
Rob-Hague, line 1: whitespace off: https://github.com/sshnet/SSH.NET/pull/1706/files?w=1

WojciechNagorski (approved):

Rob-Hague:
Thanks

PR #1711, opened 2025-10-04
Override WriteAsync in ShellStream

### LINKED ISSUE #1707: WriteAsync doesn't complete with ReadAsync running in background
I have a simple program below that call `ReadAsync` in a concurrent task. On running the program, the connection succeeds and I see the initial preamble. But the program hangs on hitting `WriteAsync`, the await never completes. Am I doing something wrong? Or am I not supposed to use these async methods?

Same behavior on .NET 9 and .Net framework 4.8.1 with v2025.0.0

```
class Program
{
    private async static Task Main(string[] args)
    {
        using var cts = new CancellationTokenSource();
        using var client = new SshClient(args[0], 22, args[1], args[2]);
        await client.ConnectAsync(cts.Token);
        await using var stream = client.CreateShellStream("test", 80, 24, 80, 24, 1024);
        var task = Task.Run(() => ReadStream(stream, cts.Token));
        var command = Encoding.ASCII.GetBytes("ls\n");

        // Continue after key press
        Console.ReadKey(true)

        await stream.WriteAsync(command, cts.Token);
        await stream.FlushAsync(cts.Token);

        // Continue after key press
        Console.ReadKey(true)

        cts.Cancel();
        await task;
    }

    private static async Task ReadStream(ShellStream stream, CancellationToken token)
    {
        while (!token.IsCancellationRequested)
        {
            try
            {
                var buffer = new byte[1024];
                var readLength = await stream.ReadAsync(buffer, token);
                Console.WriteLine(Encoding.ASCII.GetString(buffer, 0, readLength));
            }
            catch (OperationCanceledException) { }
        }
    }
}

```

Rob-Hague:
ShellStream does not currently override the Read/Write async variants. They fall back to the base class implementations which run the sync variants on a thread pool thread, only allowing one call of either at a time in order to protect implementations that would break if Read/Write were called simultaneously. In ShellStream, reads and writes are independent so mutually excluding their use is unnecessary and can lead to effective deadlocks like in #1707.

We therefore override WriteAsync to get around this restriction. We do not override ReadAsync because the sync implementation does not lend itself well to async given the use of Monitor.Wait/Pulse. Note that while reading and writing simultaneously is allowed, it is not intended that ShellStream is used with multiple simultaneous reads or multiple simultaneous writes, so it is fine to keep the base one-at-a-time implementation on ReadAsync.

Another note is that the new WriteAsync will be simple (synchronous) buffer copying in most cases, with a call to FlushAsync in others. We also do not override FlushAsync, so that will go onto a thread pool thread and potentially acquire some locks. But given that the current base implementation of WriteAsync does that unconditionally, it makes the new WriteAsync slightly better and certainly no worse than the current version.

closes #1707


WojciechNagorski (approved):

Rob-Hague:
Thanks!

PR #1712, opened 2025-10-07
Avoid rounding issues when checking Timeout values (#1700)

### LINKED ISSUE #1700: ArgumentOutOfRangeException when verifying timeout range in SshCommand ctor
When constructing an `SshCommand` via `Client.CreateCommand(string)` an `ArgumentOutOfRangeException` is **sometimes** thrown with the following message:
> The timeout must represent a value between -1 and Int32.MaxValue milliseconds, inclusive

It happens very often, but as mentioned, it DOES NOT happen all the time.

Here's the call stack:

```
Renci.SshNet.dll!Renci.SshNet.Common.TimeSpanExtensions.AsTimeout(System.TimeSpan timeSpan, string paramName) Line 23
    at /_/src/Renci.SshNet/Common/TimeSpanExtensions.cs(23)
Renci.SshNet.dll!Renci.SshNet.SshCommand.SshCommand(Renci.SshNet.ISession session, string commandText, System.Text.Encoding encoding) Line 224
    at /_/src/Renci.SshNet/SshCommand.cs(224)
Renci.SshNet.dll!Renci.SshNet.SshClient.CreateCommand(string commandText, System.Text.Encoding encoding) Line 204
    at /_/src/Renci.SshNet/SshClient.cs(204)
Renci.SshNet.dll!Renci.SshNet.SshClient.CreateCommand(string commandText) Line 196
    at /_/src/Renci.SshNet/SshClient.cs(196)
```

If I'm interpreting the code correctly, the only place where this is possible is in the `CommandTimeout` setter, which calls
`value.EnsureValidTimeout(nameof(CommandTimeout));` that then calls `TimeSpanExtensions.AsTimeout` (I'm guessing `EnsureValidTimeout` along with the property setter is inlined, which is why it doesn't show up in the call stack).

To make this even weirder, it checks against `InfiniteTimeSpan.TotalMilliseconds`, which is a static readonly field and should be -1, falling within the range.

Version: 2025.0.0
.NET Framework: 4.8.9290

nikolamilekic:
AsTimeout is called from the SshCommand constructor with Timeout.InfiniteTimeSpan. In this scenario the range check should never fail, but unfortunately it does in certain scenarios, due to a runtime or compiler bug (as soon as optimizations are turned off the issue miraculously disappears).

Closes #1700

### DIFF
--- a/src/Renci.SshNet/Common/TimeSpanExtensions.cs
+++ b/src/Renci.SshNet/Common/TimeSpanExtensions.cs
@@ -19,7 +19,7 @@ namespace Renci.SshNet.Common
         /// </exception>
         public static int AsTimeout(this TimeSpan timeSpan, [CallerArgumentExpression(nameof(timeSpan))] string? paramName = null)
         {
-            var timeoutInMilliseconds = timeSpan.TotalMilliseconds;
+            var timeoutInMilliseconds = Math.Round(timeSpan.TotalMilliseconds);
             return timeoutInMilliseconds is < -1d or > int.MaxValue
                        ? throw new ArgumentOutOfRangeException(paramName, "The timeout must represent a value between -1 and Int32.MaxValue milliseconds, inclusive.")
                        : (int)timeoutInMilliseconds;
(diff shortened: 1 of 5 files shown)


nikolamilekic:
It can be argued that the range check is completely unnecessary when called from the SshCommand ctor (as it is always called with `Timeout.InifiteTimeSpan`), and that the `_commandTimeout` field should be set in the constructor directly (instead of going through the `CommandTimeout` property, which triggers the range check). I didn't want to propose this change right away, as I'm unsure if your coding guidelines allow it.

Rob-Hague:
I had half-fixed the tests that were asserting too strictly against the original commit, so I finished that and reverted to the original. Making the comparison on `long` rather than `double` feels safest for this problem

Thanks!

Rob-Hague (approved):

PR #1715, opened 2025-10-18
Automatically publish tags to NuGet with Trusted Publishing

mus65:
this automates the publishing of releases to nuget.org and makes use of trusted publishing. See

https://learn.microsoft.com/en-us/nuget/nuget-org/trusted-publishing
https://devblogs.microsoft.com/dotnet/enhanced-security-is-here-with-the-new-trust-publishing-on-nuget-org/

This is obviously untested and would require setup in GitHub and NuGet:
- a `NUGET_USER` secret must be added to the Repository settings
- Trusted publishing must be set up on nuget.org as described [here](https://devblogs.microsoft.com/dotnet/enhanced-security-is-here-with-the-new-trust-publishing-on-nuget-org/#getting-started-%F0%9F%9A%80)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -187,7 +187,8 @@ jobs:
         name: Coverlet Results Windows .NET
         path: coverlet
 
-  Publish:
+  Publish-GitHub-Package:
+    name: Publish GitHub Package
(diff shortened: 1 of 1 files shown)


Rob-Hague (approved):
Let's try it

PR #1721, opened 2025-10-25
Update to MSTest 4

mus65:
MSTest 4 has quite a few new asserts, analyzers and breaking changes, see https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-mstest-migration-v3-v4 .

Most of these were automatically fixed by the code fixer, see separate commits.

The update also triggered new sonar analyzer warnings, which seems to a bug in sonar, see https://github.com/SonarSource/sonar-dotnet/issues/9767 . I suppressed these for now.

### DIFF
--- a/test/Renci.SshNet.Tests/Classes/OrderedDictionaryTest.cs
+++ b/test/Renci.SshNet.Tests/Classes/OrderedDictionaryTest.cs
@@ -1,4 +1,4 @@
-﻿#pragma warning disable MSTEST0037 // Use proper 'Assert' methods
+﻿#pragma warning disable MSTEST0037 // Use proper 'Assert' methods: some tests are for specific interface implementations.
 using System;
 using System.Collections.Generic;
 using System.Linq;
--- a/test/Renci.SshNet.Tests/Classes/OrderedDictionaryTest.cs
+++ b/test/Renci.SshNet.Tests/Classes/OrderedDictionaryTest.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿#pragma warning disable MSTEST0037 // Use proper 'Assert' methods
+using System;
 using System.Collections.Generic;
 using System.Linq;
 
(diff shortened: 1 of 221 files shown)


WojciechNagorski:
I think it is a good time to merge this change

test/Renci.SshNet.Tests/Classes/OrderedDictionaryTest.cs
Rob-Hague, line 1: The changes in this file should be reverted - most of them are explicitly testing ICollection<T>.Contains implementations which is now lost

(I don't think MSTEST0037 really provides much benefit)

Rob-Hague (commented):
It still suffers from missing console output in the integration tests https://github.com/sshnet/SSH.NET/pull/1693#discussion_r2370042881:

This branch:

Image

develop:

Image


I found https://github.com/microsoft/testfx/issues/6457 but apparently it is working as intended...

mus65:
> It still suffers from missing console output in the integration tests [#1693 (comment)](https://github.com/sshnet/SSH.NET/pull/1693#discussion_r2370042881):
> 
> This branch:
> 
> Image
> develop:
> 
> Image
> I found [microsoft/testfx#6457](https://github.com/microsoft/testfx/issues/6457) but apparently it is working as intended...

To be fair, I don't see how they could solve this properly (and apparently it was [broken in other ways before](https://github.com/microsoft/testfx/issues/6457#issuecomment-3322695367)). How are you going to attach the logs to a specific test if the writing happens on a separate thread?

As a workaround, we could use a custom logger that writes to the console synchronously, I pushed this in a separate branch here: https://github.com/mus65/SSH.NET/commit/043f18a393cd439c367c3aced8ec6cdda4273ff2 . It works, but I'm not sure whether it's the right way to go.

Rob-Hague:
> To be fair, I don't see how they could solve this properly (and apparently it was [broken in other ways before](https://github.com/microsoft/testfx/issues/6457#issuecomment-3322695367)). How are you going to attach the logs to a specific test if the writing happens on a separate thread?

It does currently in fact write everything to one test, but normally if I want to see the stdout logs it's because I'm debugging a single test, so that's fine (I only just noticed it now)

> As a workaround, we could use a custom logger that writes to the console synchronously, I pushed this in a separate branch here: [mus65@043f18a](https://github.com/mus65/SSH.NET/commit/043f18a393cd439c367c3aced8ec6cdda4273ff2) . It works, but I'm not sure whether it's the right way to go.

That appears to do the trick. I had thought about something that writes to the test context (https://learn.microsoft.com/dotnet/api/microsoft.visualstudio.testtools.unittesting.testcontext.writeline) but I can't see it being any better. So I'd be fine with taking that.

It would be nice if MSTest had some kind of M.E.L integration

Rob-Hague (approved):

Issue #1691, opened 2025-08-29
SftpClient.OpenAsync should return Stream for ease of mocking


doxxx:
`SftpClient.OpenAsync` currently returns an `SftpFileStream` which cannot be mocked by NSubstitute or even faked because it has private/internal constructors.

It would be nice if `OpenAsync` was instead defined as returning `Stream`, even though it actually returns a `SftpFileStream`. The only non-`Stream` public member that I can see is the `Timeout` property. If it's absolutely vital that this property is exposed, maybe define an `ISftpFileStream` interface instead? Or the timeout could be passed to `OpenAsync`.

waldemarsson:
Any update on this? Trying to write unit tests for my integration and am stuck because I can't mock `OpenAsync`

waldemarsson:
As a workaround I just created a proxy service that wraps ISftpClient.

```
internal class SftpClientProxy : IDisposable
{
    private readonly ISftpClient _sftpClient;

    public SftpClientProxy(ISftpClient sftpClient) => _sftpClient = sftpClient;

    public virtual async Task ConnectAsync(CancellationToken cancellationToken) =>
        await _sftpClient.ConnectAsync(cancellationToken);

    public virtual async Task<Stream> OpenAsync(string path, CancellationToken cancellationToken) =>
        await _sftpClient.OpenAsync(path, FileMode.Open, FileAccess.Read, cancellationToken);

    public void Dispose() => _sftpClient.Dispose();
}
```

fkubicek:
I also ran into this when trying to mock. @doxxx describes it perfectly. I would consider the option of adding `ISftpFileStream` because, IMO, it would be nicer to mock than returning `Stream`. But anything is better than the current state.

doxxx:
I took a look at this. It is not simple.

There's a lot of assumptions in SSH.NET that the type returned by `OpenAsync` (and all the other similar methods) extends `Stream`. Not to mention all of the existing user code that could assume that these methods return a type that extends `Stream`. It would be a significantly breaking change to use an interface type instead.

Changing the return type to `Stream` however is much easier. Still requires a bunch of changes here and there, mostly in test code, which specifies the expected return type exactly, but easily fixed. The only drawback to this option is that you lose access to the `Timeout` property. That said, none of the test code actually accesses this property...

doxxx:
As @Rob-Hague said in #1725, the wrapper solution by @waldemarsson is the best option.

PR #1725, opened 2025-10-31
Replace SftpFileStream with Stream in ISftpClient method signatures.

doxxx:
I'm opening this PR to initiate discussion around solutions for #1691. The solution implemented here solves #1691 because it is relatively easy to mock a method that returns `Stream`. However, this solution prevents access to the `Timeout` property of `SftpFileStream`. It is also not binary-backwards-compatible but somewhat source-backwards-compatible.

### DIFF
--- a/src/Renci.SshNet/ISftpClient.cs
+++ b/src/Renci.SshNet/ISftpClient.cs
@@ -389,7 +389,7 @@ namespace Renci.SshNet
         /// </summary>
         /// <param name="path">The path and name of the file to create.</param>
         /// <returns>
-        /// A <see cref="SftpFileStream"/> that provides read/write access to the file specified in path.
+        /// A <see cref="Stream"/> that provides read/write access to the file specified in path.
         /// </returns>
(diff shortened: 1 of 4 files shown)


Rob-Hague:
Thanks for the PR. My thoughts are basically the same as in #1508 - that changing popular public API for the sake of mocking is just not worth it at this point.

The impact of this one could potentially be lessened by changing only the interface and explicitly implementing the methods on the concrete client, but I'd still not really be convinced

doxxx:
Do you have any suggestions for how to solve the mocking issue? Right now the only viable solution is https://github.com/sshnet/SSH.NET/issues/1691#issuecomment-3365766959, which is not terrible but still a little sad that it's necessary.

Rob-Hague:
IMO the wrapper is totally valid since it provides the most control and flexibility. I think it would be the same approach as that in which one wanted to mock a component that called methods on `System.IO.File`, or `System.Net.Http.HttpClient` etc.

PR #1733, opened 2025-11-07
Add a socket receive buffer

Rob-Hague:
Currently an array is allocated to read each packet from the socket, followed by decryption which allocates another array for the plaintext payload. We can save one of these two allocations by adding a persistent buffer for socket receives, and allowing the cipher implementations to decrypt into the given payload array.

We can save the other allocation similarly, but in a separate change.

### DIFF
--- a/src/Renci.SshNet/Session.cs
+++ b/src/Renci.SshNet/Session.cs
@@ -3,6 +3,7 @@ using System.Buffers.Binary;
 using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
+using System.Net;
 using System.Net.Sockets;
 using System.Security.Cryptography;
 using System.Threading;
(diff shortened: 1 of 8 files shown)


src/Renci.SshNet/Session.cs
```diff
@@ -1927,6 +1990,11 @@ private void MessageListener()
         {
             try
             {
+                if (_socket is { } s)
+                {
+                    s.ReceiveTimeout = 0;
+                }
```
Rob-Hague, line 1969: this is just moved out from the previous implementation of TrySocketRead

WojciechNagorski:
@Rob-Hague How does this affect performance?

WojciechNagorski (approved):

Rob-Hague:
I am not expecting any noticeable speed improvements (will check). I will demonstrate the memory improvements in the next PR 🙂

PR #1750, opened 2025-12-22
Add UploadFileAsync override with canOverride

### LINKED ISSUE #1734: Missing overload for SftpClientUploadFileAsync that takes in a bool canOverride
There is an overload `public void UploadFile(Stream input, string path, bool canOverride, Action<ulong>? uploadCallback = null)`.

But in the async variant, there is no overload taking in the canOverride bool.

Please add an overload `public Task UploadFileAsync(Stream input, string path, bool canOverride, CancellationToken cancellationToken = default)`

mus65:
fixes #1734


mus65:
Test failure seems unrelated.

Rob-Hague (approved):

PR #1751, opened 2025-12-22
ScpClient: allow disabling the -d flag

### LINKED ISSUE #1746: ScpClient: Make -d flag optional or avoid it for better compatibility with Cisco SCP servers
Hi, and thanks for SSH.NET 👋

I’m using SSH.NET 2025.1.0 (via NuGet) and ran into the same Cisco SCP issue as described in #829 (“Scp on cisco device not working”).

On Cisco devices, the -d option used by ScpClient for uploads is not supported by their SCP implementation. In #829 this shows up as errors like:

_-p -d options not supported_

When I look at the 2025.1.0 source for ScpClient, I still see scp -t -d ... being used for uploads.

Example usage:

```
using var scp = new ScpClient(host, username, password);
scp.RemotePathTransformation = RemotePathTransformation.None;
scp.Connect();

using var stream = new MemoryStream(Encoding.UTF8.GetBytes(content));
scp.Upload(stream, "startup-config");
```

This works fine with OpenSSH servers but fails against Cisco devices due to the -d flag.

Request / proposal

For better interoperability, it would be helpful if ScpClient:

either removed -d for file uploads and just used

scp -t <path>

or exposed a property to disable -d, e.g.:

`public bool UseDirectoryFlag { get; set; } = true;`

and then:

```
var target = _remotePathTransformation.Transform(UseDirectoryFlag ? posixPath.Directory : posixPath.File);
var exec = UseDirectoryFlag
    ? $"scp -t -d {target}"
    : $"scp -t {target}";
if (!channel.SendExecRequest(exec))
{
    throw SecureExecutionRequestRejectedException();
}
```

That would keep current behavior for OpenSSH, but allow users to work with Cisco and other limited SCP implementations without maintaining a custom fork.

Happy to help test any change against Cisco devices if needed.

mus65:
Allows disabling the "-d" flag with `scpClient.UseDirectoryFlag = false` to workaround scp implementations that do not support this.

Fixes #1746

### DIFF
--- a/src/Renci.SshNet/ScpClient.cs
+++ b/src/Renci.SshNet/ScpClient.cs
@@ -122,6 +122,23 @@ namespace Renci.SshNet
         }
 
         /// <summary>
+        /// Gets or sets a value indicating whether the "-d" flag should be passed to the scp process on the server
+        /// when uploading files.
+        /// The "-d" flag is an undocumented flag that ensures that the target is actually a directory. However,
+        /// some scp implementations (like Cisco) do not support this flag and will fail.
+        /// You can set this to <see langword="false"/> to work around this.
+        /// </summary>
+        public bool UseDirectoryFlag { get; set; } = true;
+
+        private string EnsureIsDirectoryArg
+        {
+            get
+            {
+                return UseDirectoryFlag ? "-d" : string.Empty;
+            }
+        }
+
+        /// <summary>
         /// Occurs when downloading file.
         /// </summary>
         public event EventHandler<ScpDownloadEventArgs>? Downloading;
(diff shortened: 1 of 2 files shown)


src/Renci.SshNet/ScpClient.cs
```diff
@@ -121,6 +121,23 @@ public IRemotePathTransformation RemotePathTransformation
             }
         }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether the "-d" flag should be passed to the scp process on the server
+        /// when uploading files.
+        /// The "-d" flag is an undocumented flag that ensures that the target is actually a directory. However,
+        /// some scp implementations (like Cisco) do not support this flag and will fail.
+        /// You can set this to <see langword="false"/> to work around this.
+        /// </summary>
```
Rob-Hague:
```suggestion
        /// <summary>
        /// Gets or sets a value indicating whether the "-d" flag should be passed to the scp process on the server
        /// when uploading files. Defaults to <see langword="true"/>.
        /// </summary>
        /// <remarks>
        /// The "-d" flag is an undocumented flag that ensures that the target is actually a directory. However,
        /// some scp implementations (like Cisco) do not support this flag and will fail.
        /// You can set this to <see langword="false"/> to work around this.
        /// </remarks>
```

Rob-Hague (approved):

PR #1771, opened 2026-03-14
SftpClient: add IProgress to DownloadFileAsync and UploadFileAsync

### LINKED ISSUE #1765: Cancalletion token and download callback in the SftpClient DownloadFileAsync / BeginDownloadFile?
Both methods call `InternalDownloadFile`, which accepts both a download callback and a cancellation token.

However:
 - `DownloadFileAsync` does not support a download callback.
 - `BeginDownloadFile` does not support a cancellation token.

Is there a specific reason for this design?

I'm struggling to implement a sequential download of multiple files with a global timeout (for example, a timeout that applies to the entire download process across all files) along with progress reporting. With the current API, this seems unnecessarily challenging.

Please note that I'm not a very advanced user of async in c#. If the above does not make sense, I would appreciate advice on how to do it better.

mus65:
fixes #1765

### DIFF
--- a/src/Renci.SshNet/DownloadFileProgressReport.cs
+++ b/src/Renci.SshNet/DownloadFileProgressReport.cs
@@ -1,0 +1,13 @@
+namespace Renci.SshNet
+{
+    /// <summary>
+    /// Provides the progress for a file download.
+    /// </summary>
+    public struct DownloadFileProgressReport
+    {
+        /// <summary>
+        /// Gets the total number of bytes downloaded.
+        /// </summary>
+        public ulong TotalBytesDownloaded { get; internal set; }
+    }
+}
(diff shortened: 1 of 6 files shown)


mus65:
Build failure unrelated, opened #1772 .

Rob-Hague:
I'd be inclined to go with [`IProgress<T>`](https://learn.microsoft.com/dotnet/api/system.iprogress-1) instead of Action for new/async APIs - `Progress<T>` provides some handling of the SynchronizationContext. And perhaps make T := `struct DownloadFileProgressReport { long BytesDownloaded }` or similar so that e.g. another property could be added later without much friction. WDYT?

Also do you want to do the same for UploadFileAsync as well? I guess it could be added to

`Task UploadFileAsync(Stream input, string path, bool canOverride, CancellationToken cancellationToken = default);`

since it hasn't made a release yet. All parameters after the first two could be made optional.

mus65:
> I'd be inclined to go with [IProgress<T>](https://learn.microsoft.com/dotnet/api/system.iprogress-1) instead of Action for new/async APIs

Done. This required another overload for `UploadFileAsync` though. We can't make `canOverride` optional since that would make `UploadAsync(stream, file)` ambigous.

Rob-Hague (approved):

luijgcbvnnbbgc5 (commented):
Listo

PR #1772, opened 2026-03-14
fix Build with newer .NET 10 SDKs

mus65:
The IDE0370 are a mess since they only affect certain target frameworks. Maybe we should disable this one completely instead?

Also set a fixed SDK Version in CI so this doesn't randomly break again.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,6 +16,8 @@ jobs:
 
     - name: Setup .NET
       uses: actions/setup-dotnet@v5
+      with:
+        dotnet-version: '10.0.201'
 
     - name: Build Unit Tests .NET
       run: dotnet build -f net10.0 test/Renci.SshNet.Tests/
@@ -63,6 +65,8 @@ jobs:
 
     - name: Setup .NET
       uses: actions/setup-dotnet@v5
+      with:
+        dotnet-version: '10.0.201'
 
     - name: Build Solution
       run: dotnet build Renci.SshNet.slnx
@@ -114,6 +118,8 @@ jobs:
 
     - name: Setup .NET
       uses: actions/setup-dotnet@v5
+      with:
+        dotnet-version: '10.0.201'
 
     - name: Setup WSL2
       uses: Vampire/setup-wsl@6a8db447be7ed35f2f499c02c6e60ff77ef11278 # v6.0.0
@@ -156,6 +162,8 @@ jobs:
 
     - name: Setup .NET
       uses: actions/setup-dotnet@v5
+      with:
+        dotnet-version: '10.0.201'
 
     - name: Setup WSL2
       uses: Vampire/setup-wsl@6a8db447be7ed35f2f499c02c6e60ff77ef11278 # v6.0.0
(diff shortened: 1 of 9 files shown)


.github/workflows/build.yml
```diff
@@ -16,6 +16,8 @@ jobs:
 
     - name: Setup .NET
       uses: actions/setup-dotnet@v5
+      with:
+        dotnet-version: '10.0.201'
```
Rob-Hague: does it work to bump global.json instead?

src/Renci.SshNet/Security/Cryptography/EcdsaKey.BclImpl.cs
```diff
@@ -65,8 +65,10 @@ public override bool Verify(byte[] input, byte[] signature)
             public override void Export(out byte[] qx, out byte[] qy)
             {
                 var parameter = Ecdsa.ExportParameters(includePrivateParameters: false);
+#pragma warning disable IDE0370 // Remove unnecessary suppression
```
Rob-Hague: yeah I think we can disable this

.github/workflows/build.yml
mus65: The problem with global.json is that is has `"rollForward": "latestFeature"`, so if we bump the version there, 10.0.300 could silently break CI again...

On the other hand, disabling `rollForward` is annoying for local development because you force every developer to install a specific .NET SDK version.

Imho setting a fixed version for CI but allowing different patch versions for local development is the best thing to do.

src/Renci.SshNet/Security/Cryptography/EcdsaKey.BclImpl.cs
mus65: done

.github/workflows/build.yml
Rob-Hague: I would normally agree but given the current activity level in the library, the fewer places to maintain these numbers the better. The right long term fix is probably to stop treating every analyzer as a warning

OTOH, perhaps passing global.json to the action would do it. Hard to tell whether or not the action reads the file by default or not https://github.com/actions/setup-dotnet#using-the-global-json-file-input

mus65: The action does read global.json by default, but it also [specificially handles rollForward=latestFeature and ignores the patch version](https://github.com/actions/setup-dotnet/blob/main/src/setup-dotnet.ts#L180). 

Funnily enough, that means that it would actually use the specificed patch version if we specify latestMajor or latestMinor instead.

Maybe just set latestMajor in global.json? Since we don't treat warnings as errors in Debug anymore, this should'nt cause any issues?

Rob-Hague: If latestMinor does it then that sounds good - building older commits in release might be done for e.g. benchmarking or getting an API diff, so slightly less friction there. Otherwise latestMajor sounds good

global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
     "version": "10.0.100",
```
Rob-Hague: ?
```suggestion
    "version": "10.0.201",
```

Rob-Hague (approved):

PR #1777, opened 2026-03-25
Add ability to send posix/ansi signals in SshCommand

Levi--G:
I noticed SshCommand does not have a method for sending signals although it is supported by the channel and underlying system.

This code:
- Parses the Signal
- Sends the Signal using the preexisting method

All signals listed in https://datatracker.ietf.org/doc/html/rfc4254#section-6.9 are supported and use their libc compatible values in case someone needs conversion from libc.

Still missing are tests, but i don't understand the test project structure, if someone could help me out there that would be great.

### DIFF
--- a/src/Renci.SshNet/CommandSignal.cs
+++ b/src/Renci.SshNet/CommandSignal.cs
@@ -1,0 +1,74 @@
+﻿namespace Renci.SshNet
+{
+    /// <summary>
+    /// The ssh compatible POSIX/ANSI signals with their libc compatible values.
+    /// </summary>
+#pragma warning disable CA1720 // Identifier contains type name
+    public enum CommandSignal
+    {
+        /// <summary>
+        /// Hangup (POSIX).
+        /// </summary>
+        HUP = 1,
+
+        /// <summary>
+        /// Interrupt (ANSI).
+        /// </summary>
+        INT = 2,
+
+        /// <summary>
+        /// Quit (POSIX).
+        /// </summary>
+        QUIT = 3,
+
+        /// <summary>
+        /// Illegal instruction (ANSI).
+        /// </summary>
+        ILL = 4,
+
+        /// <summary>
+        /// Abort (ANSI).
+        /// </summary>
+        ABRT = 6,
+
+        /// <summary>
+        /// Floating-point exception (ANSI).
+        /// </summary>
+        FPE = 8,
+
+        /// <summary>
+        /// Kill, unblockable (POSIX).
+        /// </summary>
+        KILL = 9,
+
+        /// <summary>
+        /// User-defined signal 1 (POSIX).
+        /// </summary>
+        USR1 = 10,
+
+        /// <summary>
+        /// Segmentation violation (ANSI).
+        /// </summary>
+        SEGV = 11,
+
+        /// <summary>
+        /// User-defined signal 2 (POSIX).
+        /// </summary>
+        USR2 = 12,
+
+        /// <summary>
+        /// Broken pipe (POSIX).
+        /// </summary>
+        PIPE = 13,
+
+        /// <summary>
+        /// Alarm clock (POSIX).
+        /// </summary>
+        ALRM = 14,
+
+        /// <summary>
+        /// Termination (ANSI).
+        /// </summary>
+        TERM = 15,
+    }
+}
--- a/src/Renci.SshNet/SshCommand.cs
+++ b/src/Renci.SshNet/SshCommand.cs
@@ -492,7 +492,7 @@ namespace Renci.SshNet
         /// <summary>
         /// Tries to send a POSIX/ANSI signal to the remote process executing the command, such as SIGINT or SIGTERM.
         /// </summary>
-        /// <param name="signal">The signal to send</param>
+        /// <param name="signal">The signal to send.</param>
         /// <returns>If the signal was sent.</returns>
         public bool TrySendSignal(CommandSignal signal)
         {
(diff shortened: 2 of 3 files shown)


src/Renci.SshNet/CommandSignal.cs
```diff
@@ -0,0 +1,74 @@
+﻿namespace Renci.SshNet
+{
+    /// <summary>
+    /// The ssh compatible POSIX/ANSI signals with their libc compatible values.
+    /// </summary>
+#pragma warning disable CA1720 // Identifier contains type name
+    public enum CommandSignal
```
Rob-Hague: Is this list exhaustive? I am thinking not based on the penultimate paragraph of 4254 section 6.10:

> Additional 'signal name' values MAY be sent in the format
   "sig-name@xyz", where "sig-name" and "xyz" may be anything a
   particular implementer wants (except the "@" sign).

I think it would be more future-proof to have `SendSignal(string)` rather than `SendSignal(enum)`. This type could still exist as a static class for discoverability e.g.

```c#
public static class CommandSignals
{
    /// <summary>
    /// Hangup (POSIX).
    /// </summary>
    public const string HUP = nameof(HUP);
}
```

Similar to https://learn.microsoft.com/dotnet/api/system.net.mime.mediatypenames

src/Renci.SshNet/SshCommand.cs
```diff
@@ -478,6 +478,73 @@ public void CancelAsync(bool forceKill = false, int millisecondsTimeout = 500)
...
+        }
+
+        /// <summary>
+        /// Tries to send a POSIX/ANSI signal to the remote process executing the command, such as SIGINT or SIGTERM.
+        /// </summary>
+        /// <param name="signal">The signal to send</param>
+        /// <exception cref="ArgumentException">Signal was not a valid CommandSignal.</exception>
+        /// <exception cref="SshConnectionException">The client is not connected.</exception>
+        /// <exception cref="SshOperationTimeoutException">The operation timed out.</exception>
+        /// <exception cref="InvalidOperationException">The size of the packet exceeds the maximum size defined by the protocol.</exception>
```
Rob-Hague, line 519:
```suggestion
```

src/Renci.SshNet/SshCommand.cs
```diff
@@ -478,6 +478,73 @@ public void CancelAsync(bool forceKill = false, int millisecondsTimeout = 500)
...
+            return Enum.GetName(typeof(CommandSignal), signal);
+#endif
+        }
+
+        /// <summary>
+        /// Tries to send a POSIX/ANSI signal to the remote process executing the command, such as SIGINT or SIGTERM.
+        /// </summary>
+        /// <param name="signal">The signal to send</param>
+        /// <returns>If the signal was sent.</returns>
+        public bool TrySendSignal(CommandSignal signal)
```
Rob-Hague: let's drop this and stick with just `SendSignal`

Rob-Hague (commented):
Looks reasonable at first glance. Hopefully CONTRIBUTING.md at the repo root will answer any questions about tests (if not then maybe it can be improved). test\Renci.SshNet.IntegrationTests\SshClientTests.cs (or test\Renci.SshNet.IntegrationTests\OldIntegrationTests\SshCommandTest.cs) would be suitable places for basic integration test. Don't think this change needs extensive testing

src/Renci.SshNet/CommandSignal.cs
Levi--G: You are right, that would be better, I will adjust them. I propose we prepend the SIG then so its:
`public const string SIGHUP = "HUP";`
Would that be ok?

src/Renci.SshNet/SshCommand.cs
Levi--G: Why exactly would these be removed? They are exceptions from the underlying ssh calls.

src/Renci.SshNet/SshCommand.cs
Levi--G: I think there is some benefit of having a TrySendSignal though, it can be used as a last resort to send to every command started upon something happening like a server going down or similar without having to catch all the exceptions or doing all the checks yourself (some of which you don't even have access to).
channel.IsOpen for instance will cause an exception in SendSignal, but is impossible to check outside the class, leaving no option but to catch exceptions if that would be the case. Exceptions should be exceptional and methods that are prone to throwing should have Try variants according to ms documentation and best practices: https://learn.microsoft.com/en-us/dotnet/standard/exceptions/best-practices-for-exceptions

src/Renci.SshNet/CommandSignal.cs
Rob-Hague: Sounds good

src/Renci.SshNet/SshCommand.cs
Rob-Hague: 1. I am not seeing where SshOperationTimeoutException would be thrown in this path
2. You can leave it I guess, and maybe the library is inconsistent in its documentation, but I don't think this is documented (publicly) on any other path where it could happen, and it doesn't seem to have been a problem

src/Renci.SshNet/SshCommand.cs
Rob-Hague: I get it but if the best we can do is a catch-all then I would prefer to leave that up to the caller until there is a better solution (which I don't think would happen soon since the library throws a lot of exceptions around, for better or worse)

Levi--G: Its not just a catch all, it completely avoids the exceptions for the channel not being open or command having completed, which is the biggest issue atm since its not possible to check for. The internal catch would otherwise need to have been done on the caller site either way so i don't really see the issue with that? It is an option and not mandatory, so people who want exceptions can still get them, but having a method that doesn't throw when executing in a dispose scenario is really a must. Exceptions take time to catch and unwind, which is not feasible at all in high usage scenarios. Especially for a call that is likely to be used in cleanup scenarios.

src/Renci.SshNet/SshCommand.cs
Levi--G: I just followed the call stack:

- SShCommand.SendSignal()
- ChannelSession.SendSignalRequest()
- ChannelSession.SendMessage() => missing documentation of exceptions
- Session.SendMessage() => documented as throwing them
- Session.WaitOnHandle()
- Session.WaitOnHandle() => throws SshOperationTimeoutException

I think its better to have a few calls that have all of them rather than all of them are missing ;)

PR #1787, opened 2026-04-26
Encrypt packets in-place

Rob-Hague:
Support in-place encryption in the cipher types, then use it on the plaintext packets instead of allocating a new array each time. Removes 2 of 4 bytes allocated for each byte uploaded over SFTP.

For AES-CTR, supporting in-place encryption in this case means adding a persistent buffer for the keystream and encrypting in chunks. The performance difference is ~1-2% i.e. marginal versus one-shotting it. The variance is similar also for different choices of buffer size (here 4096 is used).


WojciechNagorski (approved):

Rob-Hague:
Thanks

PR #1792, opened 2026-05-04
Serialise packets into a buffer

Rob-Hague:
A byte array is allocated to hold each plaintext packet. This removes that by adding
a buffer for that purpose.

This is the last per-byte-sent allocation between the SSH channel layer and the socket transport. There remains one more in the SFTP layer.

Based on #1788, this change is the third commit

### DIFF
--- a/src/Renci.SshNet/Session.cs
+++ b/src/Renci.SshNet/Session.cs
@@ -212,6 +212,7 @@ namespace Renci.SshNet
         private Socket _socket;
 
         private ArrayBuffer _receiveBuffer = new(4 * 1024);
+        private byte[] _sendBuffer = new byte[4 * 1024];
 
         /// <summary>
         /// Gets the session semaphore that controls session channels.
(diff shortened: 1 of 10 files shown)


WojciechNagorski (approved):

src/Renci.SshNet/Session.cs
```diff
@@ -1073,29 +1074,29 @@ internal void SendMessage(Message message)
...
-            if (packetData.Length > MaximumSshPacketSize)
-            {
-                throw new InvalidOperationException(string.Format(CultureInfo.CurrentCulture, "Packet is too big. Maximum packet size is {0} bytes.", MaximumSshPacketSize));
-            }
-
             // take a write lock to ensure the outbound packet sequence number is incremented
             // atomically, and only after the packet has actually been sent
             lock (_socketWriteLock)
             {
+                var activeBufferLength = message.GetPacket(
```
Rob-Hague, line 1081: for posterity: I measured and did not see any increased lock contention from moving GetPacket inside the lock. Most parallel scenarios are already made serial by the lock in Channel.SendData

PR #1810, opened 2026-07-06
docs: add missing SftpPathNotFoundException exception documentation for ListDirectory, ListDirectoryAsync

cmgchess:
While testing, I came across `SftpPathNotFoundException` thrown by `ListDirectory`. However the exception was not listed in the XML documentation.

also came across [this](https://github.com/sshnet/SSH.NET/blob/dc8d2db12285d29c2534717c20748f450e9de4f8/test/Renci.SshNet.IntegrationTests/OldIntegrationTests/SftpClientTest.ListDirectory.cs#L24-L34) which tests this scenario


Rob-Hague (approved):
Thanks 👍

PR #1812, opened 2026-07-28
Generate ten synchronous SftpClient methods from their async versions

virzak:
## What this is

Ten of the synchronous methods on `SftpClient` are mechanical translations of the asynchronous ones beside them - the same argument checks, the same session calls with `Async` dropped, the same order. Each is maintained by hand next to its twin, so every fix has to be made twice.

This generates the synchronous version from the asynchronous source at compile time, into the same partial class. The public API is unchanged.

`DeleteFile` is representative:

```cs
public async Task DeleteFileAsync(string path, CancellationToken cancellationToken)
{
    CheckDisposed();
    ArgumentException.ThrowIfNullOrWhiteSpace(path);

    if (_sftpSession is null)
    {
        throw new SshConnectionException("Client not connected.");
    }

    cancellationToken.ThrowIfCancellationRequested();

    var fullPath = await _sftpSession.GetCanonicalPathAsync(path, cancellationToken).ConfigureAwait(false);
    await _sftpSession.RequestRemoveAsync(fullPath, cancellationToken).ConfigureAwait(false);
}
```

`await` is unwrapped, `Task` becomes `void`, the token parameter and the statements which only exist to observe it are dropped, `GetCanonicalPathAsync` resolves to `GetCanonicalPath`, and what comes out is exactly the `DeleteFile` which is in the file today.

## Verification

I compared each generated method against the hand-written one it replaces, reducing the fully qualified names the generator emits. **All ten are identical** - not equivalent, identical. The methods are `ChangeDirectory`, `CreateDirectory`, `DeleteDirectory`, `DeleteFile`, `Get`, `Exists`, `GetStatus`, `Delete`, `Open(string, FileMode, FileAccess)` and `GetAttributes`.

- `dotnet build Renci.SshNet.slnx -c Release` - clean, no new warnings, across all five target frameworks
- `Renci.SshNet.Tests` on net10.0 - 2312 passed, 0 failed, 12 skipped
- Integration tests not run here; they need Docker, which I do not have set up for this repository

To read what is generated:

```
dotnet build src/Renci.SshNet/Renci.SshNet.csproj -p:EmitCompilerGeneratedFiles=true
```

## Documentation

The converted async methods now carry `<inheritdoc />` where they previously repeated what `ISftpClient` already declares. That is the pattern `DeleteFileAsync` and `DeleteDirectoryAsync` already use, and the generated synchronous methods inherit the same documentation from the interface. Published documentation is unchanged - worth stating plainly, since the docs site is built from these.

## Deliberately left alone

- **`RenameFile`** - the synchronous version routes through an `isPosix` overload which the async version has no equivalent of, so the two are not twins.
- **`DownloadFile`, `UploadFile`, `ListDirectory`** - these genuinely differ between the two versions rather than being translations.

`SftpSession` has another 16 pairs in the same shape. I have not touched them; if this lands well, that is the obvious follow-up.

## About the dependency

[Zomp.SyncMethodGenerator](https://github.com/zompinc/sync-method-generator) is a Roslyn source generator, referenced with `PrivateAssets="all"`, so it adds nothing to the shipped package or the runtime graph. [SharpCompress adopted it](https://github.com/adamhathcock/sharpcompress/pull/1381) earlier this week for the same reason.

Disclosure: I maintain the generator. Two fixes went into it while preparing this PR - crefs in generated documentation not resolving, which would otherwise have been a build failure here, and argument list spacing. I am proposing this because the duplication in this file is exactly the case the tool handles, but if you would rather not take a build-time dependency, no hard feelings and I will close it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -23,5 +23,6 @@
     <!-- Should stay on LTS .NET releases. -->
     <PackageVersion Include="System.Formats.Asn1" Version="10.0.9" />
     <PackageVersion Include="Testcontainers" Version="4.13.0" />
+    <PackageVersion Include="Zomp.SyncMethodGenerator" Version="2.0.42" />
   </ItemGroup>
 </Project>
--- a/src/Renci.SshNet/Renci.SshNet.csproj
+++ b/src/Renci.SshNet/Renci.SshNet.csproj
@@ -47,6 +47,7 @@
     <PackageReference Include="PolySharp" PrivateAssets="all">
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="Zomp.SyncMethodGenerator" PrivateAssets="all" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'net462' ">
--- a/src/Renci.SshNet/SftpClient.cs
+++ b/src/Renci.SshNet/SftpClient.cs
@@ -23,7 +23,7 @@ namespace Renci.SshNet
     /// <summary>
     /// Implementation of the SSH File Transfer Protocol (SFTP) over SSH.
     /// </summary>
-    public class SftpClient : BaseClient, ISftpClient
+    public partial class SftpClient : BaseClient, ISftpClient
     {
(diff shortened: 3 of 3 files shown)


Rob-Hague:
Interesting, but no thanks

PR #1823, opened 2026-08-12
Fix race between SendMessage and server-initiated key re-exchange

### LINKED ISSUE #1764: [SFTP] Race Condition during Rekeying: "Message type 93 is not valid" or Connection Drop
**Description**
We are encountering critical connection failures during SFTP uploads under high network throughput. The issue appears to be a race condition during the SSH Rekeying phase.

When the server triggers a Key Rotation (due to RekeyLimit), the client fails to pause the data stream immediately. Consequently, the client sends a SSH_MSG_CHANNEL_DATA packet while the server has entered the Key Exchange state and expects SSH_MSG_KEXINIT, leading to a protocol violation.

**Exceptions & Stack Traces**
We observe two different exceptions depending on the server software/configuration:

**Variant 1: Protocol Violation (Server sends Disconnect message)**

```
Renci.SshNet.Common.SshConnectionException: Message type 93 is not valid.
   at Renci.SshNet.Session.WaitOnHandle(WaitHandle waitHandle, TimeSpan timeout)
   at Renci.SshNet.Channels.Channel.SendData(Byte[] data)
   at Renci.SshNet.Channels.Channel.SendMessage(Message message)
   at Renci.SshNet.Channels.Channel.SendData(Byte[] buffer, Int32 offset, Int32 size)
   at Renci.SshNet.Sftp.SftpSession.RequestWrite(Byte[] handle, UInt64 serverOffset, Byte[] data, Int32 offset, Int32 length, Action`1 writeStatusCallback)
```
**Variant 2: Connection Drop (Server resets TCP connection)**

```
Renci.SshNet.Common.SshConnectionException: Client not connected.
   at Renci.SshNet.Common.AsyncResult.EndInvoke()
   at Renci.SshNet.SftpClient.EndUploadFile(IAsyncResult asyncResult)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncCoreLogic(IAsyncResult iar, Func`2 endFunction, Action`1 endAction, Task`1 promise, Boolean requiresSynchronization)
   --- End of stack trace from previous location ---
   at FtpUtils.FtpUtils.SftpTransferFileAsync(...)
```

**Steps to Reproduce**
Configure an SFTP server with a low RekeyLimit (e.g., RekeyLimit 500M 0).

Establish a high-speed connection (e.g., > 500 Mbps).

Upload a file significantly larger than the limit.

Result: Connection crashes exactly when the limit is reached.

**Analysis**
The library seems unable to handle the state transition to KeyExchange fast enough when the outgoing socket buffer is saturated with file data. Confirmed Workaround: The issue is resolved only by setting RekeyLimit none on the server side, which confirms the issue is isolated to the rekeying phase.

**Additional Info**

- Reproducible on the latest revision 2025.1.0 as well as older versions.
- Reproducible with both synchronous UploadFile and asynchronous UploadFileAsync / BeginUploadFile implementations.

**Question**
 Is there any client-side configuration or internal buffer setting that could mitigate this race condition without changing the server config? Or any other trick. Thanks!

radekvermirovsky:
Fixes #1764

## The race

`Session.SendMessage` checks `_keyExchangeCompletedWaitHandle` *before* acquiring `_socketWriteLock`. When the server initiates a key re-exchange in that window, the message loop thread resets the wait handle and sends the client's `SSH_MSG_KEXINIT` (under the same write lock). A data message that already passed the check then acquires the lock and goes out on the wire *after* our KEXINIT, violating RFC 4253 section 7.1 ("Once a party has sent a SSH_MSG_KEXINIT message ... it MUST NOT send any messages other than: Transport layer generic messages ...; Algorithm negotiation messages ...; Specific key exchange method messages").

Strict servers such as ProFTPD mod_sftp then fail the exchange or drop the connection, producing the two failure variants reported in #1764 (`Message type 93 is not valid` / connection drop with `Key exchange failed`).

## The fix

`SendMessage` now re-checks the wait handle *while holding the write lock* and goes back to waiting when a re-exchange has started in the meantime. Re-checking under the lock is required: making the reset + KEXINIT send atomic is not sufficient, because a sender can already be blocked on the lock with a stale check result.

The packet is now also built entirely under the write lock, so a completing re-exchange can no longer swap `_clientCipher`/`_clientMac`/compression in the middle of building a packet.

The diff looks bigger than it is because the packet-building body moved unchanged into `SendMessageWithinWriteLock` — reviewing with whitespace changes hidden shows the actual change is the retry loop.

## Why #1774 could not reproduce it

The repro attempt in #1774 ran against the OpenSSH test server. OpenSSH queues non-key-exchange output while a re-exchange is in progress and tolerates client data that slips in, so the race stays invisible there. ProFTPD mod_sftp does not tolerate it, so this PR adds an integration test running concurrent SFTP uploads against a ProFTPD container (`alpine:3.24`, ProFTPD 1.3.9c) configured to re-key every 1 MB (`SFTPRekey required 3600 1`), giving the race many trials per upload.

## Evidence

Same environment (Docker Desktop/WSL2, net10.0), 4 concurrent uploads of 128 MB files, 3 attempts per run:

| | Result |
|---|---|
| Without the fix | 7 of 8 runs failed with `SshConnectionException: Key exchange failed` |
| With the fix | 4 of 4 runs passed |


radekvermirovsky:
A note on CI coverage: the new ProFTPD rekey test currently runs on the Linux job only. On the Windows jobs it reports inconclusive, because Docker on the Windows runners is in Windows containers mode and cannot run the Linux ProFTPD image.

If you'd like it to run there too, I'm happy to extend the "Setup SSH Server" step of the Windows Integration Tests .NET job to also build and start the ProFTPD container with Podman in WSL2 on a second fixed port (mirroring how the OpenSSH server is provided), and to switch the test to the same Windows-CI detection `InfrastructureFixture` uses instead of skipping. Just let me know if that's wanted, or if the Linux coverage is sufficient.

radekvermirovsky:
@Rob-Hague  @WojciechNagorski  @drieseng  — gentle ping on this one.

The branch is up to date with `develop` and CI is green on the Linux integration job. I'm happy to keep maintaining it (rebases, review feedback, extending the ProFTPD rekey test to the Windows runners as offered above), but I don't want it sitting in the queue if it isn't a direction you want to take.

Could one of you let me know whether this is worth keeping open? Specifically:

1. Is the approach (re-checking `_keyExchangeCompletedWaitHandle` under `_socketWriteLock` and building the packet inside the lock) acceptable in principle, or would you prefer the fix somewhere else in the send path?
2. Is the Linux-only ProFTPD integration test enough, or should I wire it up for the Windows CI jobs too?

If there's no appetite for it, just say so and I'll close it — no hard feelings. Otherwise I'll keep it rebased and ready.

PR #1824, opened 2026-08-12
Remove some dead internal code

mus65:
Remove some methods and classes that are internal, but never actually used anywhere.

Note there are some additional events in `Session.cs` that could technically be removed because there are internal, but nothing ever adds a handler to them (e.g. `IgnoreReceived`). I kept these for now for consistency with the rest of the code, but I could remove these (and the callbacks the are used in) as well.


drieseng:
@Rob-Hague, I'm ok with this but I leave it up to you.

FYI: I recently upgraded to **2026.0.0** at work, and performance is great. Thx!!!!!

Rob-Hague (approved):
> but I could remove these (and the callbacks the are used in) as well.

Anything dead seems good to delete (I have briefly thought about improving symlink stuff, but it can always be brought back)

@drieseng thanks 🙂

PR #1826, opened 2026-08-15
Add ConnectionHandler for customizing connection establishment

cloud-hai-vo:
## What this does

Adds a public, composable `ConnectionHandler` abstraction for customizing how the underlying socket connection is established — mirroring `System.Net.Http.DelegatingHandler`/`HttpClient`, as discussed in #1821.

- `ConnectionHandler` — abstract base (`Connect`/`ConnectAsync`/`Dispose`)
- `DelegatingConnectionHandler` — wraps an inner handler, forwards by default; override only what you need
- `DefaultConnectionHandler` — public singleton reproducing this library's built-in connect behavior (including proxy support), usable as the innermost handler in a custom chain
- `ConnectionInfo.ConnectionHandler` — new opt-in property, default `null`

`Session`'s own default path is unchanged — it still goes through the existing connector machinery directly, and is only bypassed when a caller explicitly sets `ConnectionInfo.ConnectionHandler`.

## Testing

New tests for the handler delegation logic, `DefaultConnectionHandler`, and that `Session` uses a custom handler when one is set. Full existing suite passes, no regressions.

## Docs

Added an example to `docfx/examples.md` showing how to compose a custom handler on top of `DefaultConnectionHandler`.

cc @Rob-Hague @drieseng @WojciechNagorski

### DIFF
--- a/docfx/examples.md
+++ b/docfx/examples.md
@@ -126,6 +126,54 @@ using (var client = new SshClient("sftp.foo.com", "user", "password"))
 }
 ```
 
+### Customize the underlying socket connection
+
+Implement `ConnectionHandler` (or `DelegatingConnectionHandler`, if you only want to customize
+part of the default behavior) to configure the underlying socket beyond what `ConnectionInfo`
+exposes directly - for example, tuning the socket's buffer sizes on a high-latency connection:
+
+```cs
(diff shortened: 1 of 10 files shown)


Rob-Hague:
Thanks, but it's not what I have in mind (allow general Stream not Socket). I prefer to do it myself

